Issue #21, opened 2023-07-19
Fix logerrors  type in pg_exttable


my-ship-it:
logerrors in pg_exttable is char. Spike whether we need to change it bool?

```
postgres=# select version();
----------------------------------------------------------------------------------------
 PostgreSQL 14.4 (Cloudberry Database 1.3.0 build commit:bb9d07938bab13f6117bd565d5c71ca878bc4dbe) on x86_64-pc-linux-gnu, compiled by gcc (GCC) 10.2.1 20210130 (Red Hat 10.2.1-11), 64-bit compiled on May 24 2023 16:39:21
(1 row)


postgres=# \d pg_exttable;
               View "pg_catalog.pg_exttable"
     Column      |  Type   | Collation | Nullable | Default 
-----------------+---------+-----------+----------+---------
 reloid          | oid     |           |          | 
 urilocation     | text[]  |           |          | 
 execlocation    | text[]  |           |          | 
 fmttype         | "char"  |           |          | 
 fmtopts         | text    |           |          | 
 options         | text[]  |           |          | 
 command         | text    |           |          | 
 rejectlimit     | integer |           |          | 
 rejectlimittype | "char"  |           |          | 
 logerrors       | "char"  |           |          | 
 encoding        | integer |           |          | 
 writable        | boolean |           |          | 


db1=# select version();

---------------------------------
 PostgreSQL 12.12 (Greenplum Database 7.0.0-beta.4 build commit:f3782c6dd89c9b6565a84518174ed4dd448f3293 Open Source) on x86_64-pc-linux-gnu, compiled by gcc (GCC) 8.5.0 20210514 (Red Hat 8.5.0-18), 64-bit compiled on 
Jun 15 2023 01:11:10 Bhuvnesh C.
(1 row)


db_gp7=# \d pg_exttable
               View "pg_catalog.pg_exttable"
     Column      |  Type   | Collation | Nullable | Default 
-----------------+---------+-----------+----------+---------
 reloid          | oid     |           |          | 
 urilocation     | text[]  |           |          | 
 execlocation    | text[]  |           |          | 
 fmttype         | "char"  |           |          | 
 fmtopts         | text    |           |          | 
 options         | text[]  |           |          | 
 command         | text    |           |          | 
 rejectlimit     | integer |           |          | 
 rejectlimittype | "char"  |           |          | 
 logerrors       | boolean |           |          | 
 encoding        | integer |           |          | 
 writable        | boolean |           |          |

```

[...]

PR #116, opened 2023-08-03
Fix incompatible logerrors type in pg_exttable

Ray-Eldath:
closes: #21 

---

### Change logs

This commit is directly cherry-picked from https://github.com/greenplum-db/gpdb/pull/14123.

`logerrors` type in `pg_foreign_table` has been changed from bool to char *by design* in GPDB7 to support log error persistence for external tables, and originally the `logerrors` of extension `gp_exttable_fdw` was changed accordingly. This PR changes the type back to preserve backward compatability.

Because the factual data are stored in `pg_foreign_table` and this PR only changes the display of the data, it should have no compatability implication.

No.


Ray-Eldath:
I reset the author of this PR from the original author of https://github.com/greenplum-db/gpdb/pull/14123 to me. Don't know if this will cause problems.

tuhaihe:
> I reset the author of this PR from the original author of [greenplum-db/gpdb#14123](https://github.com/greenplum-db/gpdb/pull/14123) to just me. Don't know if this will cause problems. If so I can change it back.

Hi @Ray-Eldath, thanks for you to point this out. If we just cherry-pick the commits for the upstream without any changes, we should keep the original author and copyright information.

Ray-Eldath:
> > I reset the author of this PR from the original author of [greenplum-db/gpdb#14123](https://github.com/greenplum-db/gpdb/pull/14123) to just me. Don't know if this will cause problems. If so I can change it back.
> 
> Hi @Ray-Eldath, thanks for you to point this out. If we just cherry-pick the commits for the upstream without any changes, we should keep the original author and copyright information.

but then there's a CLA issue.

tuhaihe:
> but then there's a CLA issue.

@Ray-Eldath: Sorry to confuse you. The CLA is optional, and not required, especially for the upstream contributors. Their code is already under Apache License v2.0, for us, we have got the grant to redistribute them, so It's ok to ignore the CLA. Let's keep the original copyright information, thanks!

avamingli (approved):

my-ship-it (approved):

Issue #93, opened 2023-07-25
CHECK index unique throw exception


my-ship-it:
**Root Cause Analysis**

DirtySnapshot is created on stack, but will be recorded in Resowner.
If transaction abort, there is no chance to remove snapshot from resource owner anyway, which caused invalid memory access.

The stack trace:
```
_bt_check_unique(DirtySnapshot is created here)
   -> table_index_fetch_tuple_check
      -> ...
         -> AppendOnlyVisimapStore_Init
            -> RegisterSnapshot(snapshot)
```

UnregisterSnapshot is called in AppendOnlyVisimapStore_Finish. If transaction aborts, there is no chance to call AppendOnlyVisimapStore_Finish.

**Reproduce steps**

kd97:
```
set default_table_access_method=ao_column;
create table test_range_excl(
  id int4,
  room int4range,
  speaker int4range,
  during tsrange,
  exclude using gist (room with =, during with &&),
  exclude using gist (speaker with =, during with &&)
) DISTRIBUTED REPLICATED;
insert into test_range_excl values(1, int4range(123, 123, '[]'), int4range(3, 3, '[]'), '[2010-01-02 10:00, 2010-01-02 11:00)');
// conflicting key value violates exclusion constraint
// expected catch error,but will crash occasionally in aarch64 centos platform
insert into test_range_excl values(1, int4range(123, 123, '[]'), int4range(3, 3, '[]'), '[2010-01-02 10:00, 2010-01-02 11:00)');
```

PR #117, opened 2023-08-03
FIX: constraint check exception with snapshot memory leak in AO/AOCS

lss602726449:
closes: #93

---

### Change logs
This bug comes from AppendOnlyVisimapStore_Init calling RegisterSnapshot in appendonly_fetch_init, which creates a deep copy of the current snapshot. Therefore, when a constraint is violated(see function check_exclusion_or_unique_constraint in detail), EROOR will be performed directly, but UnregisterSnapshot in AppendOnlyVisimap_Finish is not called. Therefore, an Asset error will occur when the resource is finally recycled.

In order to solve this problem: directly delete the RegisterSnapshot in AppendOnlyVisimapStore_Init which affects the deletion logic of the AO table. So we did not change AppendOnlyVisimap_Init, but passed in an InvalidSnapshot during initialization. And at the same time, in order to ensure that Visimap can get the correct snapshot. Before the execution of appendonly_fetch, the snapshot of visimap is assigned and deleted afterwards. Note that the free of the snapshot is still done by who malloc.

See the issue for detail 

NO

Still wait for adding

### Contributor's Checklist

- [x] Make sure your Pull Request has a clear title and commit message. You can take [git-commit](https://github.com/cloudberrydb/cloudberrydb/blob/main/.gitmessage) template as a reference.
- [x] Sign the Contributor License Agreement as prompted for your first-time contribution.
- [x] List your communication in the [GitHub Issues](https://github.com/cloudberrydb/cloudberrydb/issues) or [Discussions](https://github.com/orgs/cloudberrydb/discussions) (if has or needed).
- [x] Document changes.
- [x] Feel free to @cloudberrydb/dev team for review and approval when your PR is ready🥳

### DIFF
--- a/src/backend/access/appendonly/appendonlyam_handler.c
+++ b/src/backend/access/appendonly/appendonlyam_handler.c
@@ -1978,6 +1978,8 @@ appendonly_scan_bitmap_next_tuple(TableScanDesc scan,
 
 	}
 
+	aoscan->aofetch->visibilityMap.visimapStore.snapshot = aoscan->appendOnlyMetaDataSnapshot;
+
 	ExecClearTuple(slot);
 
 	/* ntuples == -1 indicates a lossy page */
@@ -2011,11 +2013,12 @@ appendonly_scan_bitmap_next_tuple(TableScanDesc scan,
 		{
 			/* OK to return this tuple */
 			pgstat_count_heap_fetch(aoscan->aos_rd);
+			aoscan->aofetch->visibilityMap.visimapStore.snapshot = InvalidSnapshot;
 
 			return true;
 		}
 	}
-
+	aoscan->aofetch->visibilityMap.visimapStore.snapshot = InvalidSnapshot;
 	/* Done with this block */
 	return false;
 }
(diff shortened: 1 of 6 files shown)


src/backend/access/appendonly/appendonlyam_handler.c
```diff
@@ -545,8 +545,12 @@ appendonly_index_fetch_tuple(struct IndexFetchTableData *scan,
 	 */
 	Assert(aoscan->aofetch->snapshot == snapshot);
 
+	aoscan->aofetch->visibilityMap.visimapStore.snapshot = snapshot;
+
 	appendonly_fetch(aoscan->aofetch, (AOTupleId *) tid, slot);
 
+	aoscan->aofetch->visibilityMap.visimapStore.snapshot = InvalidSnapshot;
+
 	return !TupIsNull(slot);
```
avamingli, line 554: Hmm.. set/reset at fetching each index tuple, we usually have million/billion of rows, and the set/rest cost will be double.
That's not great.

lss602726449: Thanks for your suggestion, I think it is not great, too. But it works....
1. we have already try another easy way to solve this problem, by removing register snapshot. But failed in CI test and trying to fix it. https://github.com/cloudberrydb/cloudberrydb/pull/136
2. we see the same way to slove the problem in unique index check by set/reset. And I think AO may not efficent in index fetch, so millions tuple index scan may not efficient for index fetch even without set and reset 
3. beg your better way to solve this problem, thanks.

PR #136, opened 2023-08-10
FIX: constraint check exception with snapshot memory leak in AO/AOCS

lss602726449:
closes: #93

---
### Change logs

This bug comes from AppendOnlyVisimapStore_Init calling RegisterSnapshot in appendonly_fetch_init, which creates a deep copy of the current snapshot. Therefore, when a constraint is violated(see function check_exclusion_or_unique_constraint in detail), EROOR will be performed directly, but UnregisterSnapshot in AppendOnlyVisimap_Finish is not called. Therefore, an Asset error will occur when the resource is finally recycled.

In order to solve this problem, we directly delete the RegisterSnapshot in AppendOnlyVisimapStore_Init which is useless and at the same time delete the UnregisterSnapshot.

FIX: constraint check exception with snapshot memory leak in AO/AOCS

No

No need

### Contributor's Checklist

- [x] Make sure your Pull Request has a clear title and commit message. You can take [git-commit](https://github.com/cloudberrydb/cloudberrydb/blob/main/.gitmessage) template as a reference.
- [x] Sign the Contributor License Agreement as prompted for your first-time contribution.
- [x] List your communication in the [GitHub Issues](https://github.com/cloudberrydb/cloudberrydb/issues) or [Discussions](https://github.com/orgs/cloudberrydb/discussions) (if has or needed).
- [x] Document changes.
- [x] Feel free to @cloudberrydb/dev team for review and approval when your PR is ready🥳

### DIFF
--- a/src/backend/access/appendonly/appendonly_visimap_store.c
+++ b/src/backend/access/appendonly/appendonly_visimap_store.c
@@ -43,7 +43,6 @@ AppendOnlyVisimapStore_Finish(AppendOnlyVisimapStore *visiMapStore,
 		visiMapStore->scanKeys = NULL;
 	}
 
-	UnregisterSnapshot(visiMapStore->snapshot);
 	index_close(visiMapStore->visimapIndex, lockmode);
 	table_close(visiMapStore->visimapRelation, lockmode);
 }
@@ -71,7 +70,7 @@ AppendOnlyVisimapStore_Init(AppendOnlyVisimapStore *visiMapStore,
 	Assert(OidIsValid(visimapRelid));
 	Assert(OidIsValid(visimapIdxid));
 
-	visiMapStore->snapshot = RegisterSnapshot(snapshot);
+	visiMapStore->snapshot = snapshot;
 	visiMapStore->memoryContext = memoryContext;
 
 	visiMapStore->visimapRelation = table_open(visimapRelid, lockmode);


HuSen8891 (approved):

src/backend/access/appendonly/appendonly_visimap_store.c
```diff
@@ -71,7 +70,7 @@ AppendOnlyVisimapStore_Init(AppendOnlyVisimapStore *visiMapStore,
 	Assert(OidIsValid(visimapRelid));
 	Assert(OidIsValid(visimapIdxid));
 
-	visiMapStore->snapshot = RegisterSnapshot(snapshot);
```
avamingli, line 74: If we don't register snapshot, when and where the snapshot is freed?

lss602726449: I'm sorry to answer your question for so long. For unique check, the snapshot is allocate and destory in the stack in the fuction _bt_check_unique, see nbtinsert.c:431(InitDirtySnapshot(SnapshotDirty);) for detail

avamingli:
Need a rebase.

avamingli (approved):

liming01:
It seems that this fix has 2 problems:

- AppendOnlyVisimapStore_Init() & AppendOnlyVisimapStore_Finish() can be called by other scenarios. i.e. _bt_check_unique() is not the only function calls these 2 functions internally. So we can not just remove calling UnregisterSnapshot().
- The root cause of crash in this scenario I guess is:  the stack variable DirtySnapshot initial value is random ( ref to [link](https://stackoverflow.com/questions/3845353/what-will-be-the-default-value-of-an-uninitialized-boolean-value-in-c) ), when the snapshot->copied is true, RegisterSnapshotOnOwner() will directly register this snapshot instead of copying one.  So we just need to memset DirtySnapshot to '\0' before using it.

avamingli:
> @lss602726449 , @avamingli , @my-ship-it 
> 
> 
> 
> It seems that this fix has 2 problems:
> 
> 
> 
> - AppendOnlyVisimapStore_Init() & AppendOnlyVisimapStore_Finish() can be called by other scenarios. i.e. _bt_check_unique() is not the only function calls these 2 functions internally. So we can not just remove calling UnregisterSnapshot().
> 
> - The root cause of crash in this scenario I guess is:  the stack variable DirtySnapshot initial value is random ( ref to [link](https://stackoverflow.com/questions/3845353/what-will-be-the-default-value-of-an-uninitialized-boolean-value-in-c) ), when the snapshot->copied is true, RegisterSnapshotOnOwner() will directly register this snapshot instead of copying one.  So we just need to memset DirtySnapshot to '\0' before using it.
> 
> 

Hi,thanks! Reasonable, +1.

lss602726449:
Thanks, it is my mistake. I'll fix this problem and test as you suggest

lss602726449:
@avamingli, @liming01 
I'm sorry to answer this too late. I have try to fix this problem by setting snapshot->copied. Howerver it doesn't work. And I find that DirtySnapshot will not cause deep copy in the CopySnapshot. I'm thinking of other solutions.
```
	if (!IsMVCCSnapshot(snapshot))
		return snapshot;
```

PR #142, opened 2023-08-11
Add final Motion earlier in grouping_planner for partial paths

avamingli:
This is same as pathlist, if we know where the result will be needed, create a Motion to move it there.
standard_planner() will tack a Motion on top of the cheapest partial path anyway, if we don't do it here. 
But doing the Motion here allows the cost of the Motion to be taken into account when deciding which partial path is the cheapest.

closes: #ISSUE_Number

---

### Change logs

### Contributor's Checklist

- [ ] Sign the Contributor License Agreement as prompted for your first-time contribution.
- [ ] Document changes.
- [ ] Feel free to @cloudberrydb/dev team for review and approval when your PR is ready🥳

### DIFF
--- a/src/backend/optimizer/plan/planner.c
+++ b/src/backend/optimizer/plan/planner.c
@@ -2579,6 +2579,26 @@ grouping_planner(PlannerInfo *root, double tuple_fraction)
 												  offset_est, count_est);
 			}
 
+			/*
+			 * Like pathlist:
+			 * Take Motion cost into accout before standard_planner().
+			 * Don't foo by one-phase LIMIT with partial_path here.
+			 */
+			if ((CdbPathLocus_IsHashed(root->final_locus) ||
+				CdbPathLocus_IsSingleQE(root->final_locus) ||
+				CdbPathLocus_IsEntry(root->final_locus) ||
+				CdbPathLocus_IsReplicated(root->final_locus)) &&
+				!root->glob->is_parallel_cursor)
+			{
+				Path	   *orig_path = partial_path;
+
+				partial_path = cdbpath_create_motion_path(root, orig_path,
+												  root->sort_pathkeys,
+												  false,
+												  root->final_locus);
+				if (!partial_path)
+					partial_path = orig_path;
+			}
 			add_partial_path(final_rel, partial_path);
 		}
 	}


src/backend/optimizer/plan/planner.c
```diff
@@ -2561,6 +2561,26 @@ grouping_planner(PlannerInfo *root, double tuple_fraction)
 												  offset_est, count_est);
 			}
 
+			/*
+			 * Like pathlist:
+			 * Take Motion cost into accout before standard_planner().
+			 * Don't foo by one-pahse LIMIT with partial_path here.
```
my-ship-it, line 2567: s/one-pahse/one-phase ?

avamingli: Done.

src/backend/optimizer/plan/planner.c
```diff
@@ -2579,6 +2579,26 @@ grouping_planner(PlannerInfo *root, double tuple_fraction)
 												  offset_est, count_est);
 			}
 
+			/*
+			 * Like pathlist:
+			 * Take Motion cost into accout before standard_planner().
+			 * Don't foo by one-phase LIMIT with partial_path here.
```
foreyes, line 2585: Are these logics duplicated with the above logic? Should we consider moving the above logic directly here instead of adding once more?

avamingli: > Are these logics duplicated with the above logic?

Yeah, the codes are same.

> Should we consider moving the above logic directly here instead of adding once more?

No, the codes are same, but for different purpose and condition. 
The previous one are used for collect results for Limit  on QD, that must be ahead of add_limit_node.
And this new one is used for Motion cost before we do it finally at standard_planner, and we must put this one after Limit.
We couldn't combine them easily, it will introduce complex logic with `if conditions` for just saving a couple of lines.
Maybe I should tell more behind the comments: `Don't foo by one-phase LIMIT with partial_path here.`

foreyes: Oh, that make sense.

foreyes (approved):
LGTM

my-ship-it (approved):

PR #149, opened 2023-08-15
Implement Parallel-aware Hash Left Anti Semi (Not-In) Join

avamingli:
For parallel-aware hash join, we need to sync between parallel
    workers to tell the right results when there are NULL values.

 If we are LASJ and found NULL value by ourself or sibling processes
    had found NULL values, quit and tell siblings to quit if possible.
    It's safe to fetch and set phs_lasj_has_null without lock here and at
    other places. As it's a boolean and we don't need to have the most
    recent value from CPU or Mem cache. And we should avoid more locks in
    HashJion Impl.
 If we miss it here and some others set it at the same time, just
    bypass and we may get it at the next Hash batch.
    If we missed it across all batches, we will know it when
    PHJ_BUILD_HASHING_INNER ends with the help of build_barrier.
    If we never participated in building hash table, check it when hash
    table creation job is finished.

```sql
gpadmin=# explain(costs off) select c1 from ao1 where c1 not in(select c2 from ao2);
                                QUERY PLAN
---------------------------------------------------------------------------
 Gather Motion 12:1  (slice1; segments: 12)
   ->  Parallel Hash Left Anti Semi (Not-In) Join
         Hash Cond: (ao1.c1 = ao2.c2)
         ->  Parallel Seq Scan on ao1
         ->  Parallel Hash
               ->  Parallel Broadcast Motion 12:12  (slice2; segments: 12)
                     ->  Parallel Seq Scan on ao2
(8 rows)
gpadmin=# set enable_parallel=off;
SET
Time: 1.020 ms
gpadmin=# explain(costs off) select c1 from ao1 where c1 not in(select c2 from ao2);
                          QUERY PLAN
---------------------------------------------------------------
   ->  Hash Left Anti Semi (Not-In) Join
         Hash Cond: (ao1.c1 = ao2.c2)
         ->  Seq Scan on ao1
         ->  Hash
               ->  Broadcast Motion 3:3  (slice2; segments: 3)
                     ->  Seq Scan on ao2
(8 rows)
```

### A special case NOT IN subslect has null value:

Table ao2 has 1 billion rows in seg file 0-3 and with a NULL value in seg file 4, launch a 4-workers plan.

```sql
gpadmin=# select count(*) from ao2;
   count
------------
 1000000003
(1 row)
```

```sql
gpadmin=# select c1 from ao1 where c1 not in(select c2 from ao2);
 c1
----
(0 rows)

Time: 309224.911 ms (05:09.225)
set enable_parallel = on;
gpadmin=# select c1 from ao1 where c1 not in(select c2 from ao2);
 c1
----
(0 rows)

Time: 192.844 ms
```
Time: non-parallel plan **309224.911 ms** to parallel-aware plan **192.844 ms,** **1600x faster**.

### NOT IN subselect has no null values.

```sql
select count(*) from t2 where c1 not in (select c2 from t1);
```

parallel workers | avg duration(s) | 1st | 2nd | 3rd
-- | -- | -- | -- | --
0|41.504|41.792|41.446|41.275
2|27.757|28.637|27.099|27.534
4|24.990|25.130|24.482|25.360
6|24.056|24.489|23.721|23.958

DDL & DML
```sql
create table t1(c1 int, c2 int);
create table t2(c1 int, c2 int);
insert into t1 select i, i+1 from generate_series(1, 40000000) i;
insert into t2 select i, i+1 from generate_series(1, 40000000) i;
```

closes: #ISSUE_Number

---

### Change logs

### Contributor's Checklist

- [ ] Sign the Contributor License Agreement as prompted for your first-time contribution.
- [ ] Document changes.
- [ ] Feel free to @cloudberrydb/dev team for review and approval when your PR is ready🥳

### DIFF
--- a/src/backend/executor/nodeHash.c
+++ b/src/backend/executor/nodeHash.c
@@ -324,6 +324,14 @@ MultiExecParallelHash(HashState *node)
 			{
 				bool		hashkeys_null = false;
 
+				/* CBDB_PARALLEL: Siblings must have found null value. */
+				if (pstate->phs_lasj_has_null)
+				{
+					node->hs_hashkeys_null = true;
+					ExecSquelchNode(outerNode);
+					break;
+				}
+
 				slot = ExecProcNode(outerNode);
 				if (TupIsNull(slot))
 					break;
@@ -333,14 +341,40 @@ MultiExecParallelHash(HashState *node)
 										 &hashvalue, &hashkeys_null))
 					ExecParallelHashTableInsert(hashtable, slot, hashvalue);
 				hashtable->partialTuples++;
+
+				if (node->hs_quit_if_hashkeys_null && hashkeys_null)
+				{
+					/* CBDB_PARALLEL:
+					 * If we are LASJ and found NULL value by ourself or sibling processes had
+					 * found NULL values, quit and tell siblings to quit if possible.
+					 *
+					 * It's safe to fetch and set phs_lasj_has_null without lock here and at
+					 * other places. As it's a atomic boolean value. And we should avoid more locks in HashJion Impl.
+					 * If other processes miss it here and some others set it at the same time, just bypass
+					 * and we may get it at the next Hash batch.
+					 * If we missed it across all batches, we will know it when PHJ_BUILD_HASHING_INNER
+					 * ends with the help of build_barrier.
+					 * If we never participated in building hash table, check it when hash table
+					 * creation job is finished.
+					 */
+					pstate->phs_lasj_has_null = true;
+					pg_write_barrier();
+					node->hs_hashkeys_null = true;
+					ExecSquelchNode(outerNode);
+					break;
+				}
 			}
 
+			/* CBDB_PARALLEL: No need to flush tuples if phs_lasj_has_null. */
 			/*
 			 * Make sure that any tuples we wrote to disk are visible to
 			 * others before anyone tries to load them.
 			 */
-			for (i = 0; i < hashtable->nbatch; ++i)
-				sts_end_write(hashtable->batches[i].inner_tuples);
+			if (!pstate->phs_lasj_has_null)
+			{
+				for (i = 0; i < hashtable->nbatch; ++i)
+					sts_end_write(hashtable->batches[i].inner_tuples);
+			}
 
 			/*
 			 * Update shared counters.  We need an accurate total tuple count
@@ -366,7 +400,21 @@ MultiExecParallelHash(HashState *node)
 				 * skew).
 				 */
 				pstate->growth = PHJ_GROWTH_DISABLED;
+				/* In case we didn't find null values ourself. */
+				if (pstate->phs_lasj_has_null)
+				{
+					node->hs_hashkeys_null = true;
+					return;
+				}
 			}
+	}
+
+	/* In case we didn't participate in PHJ_BUILD_HASHING_INNER */
+	pg_memory_barrier();
+	if (pstate->phs_lasj_has_null)
+	{
+		node->hs_hashkeys_null = true;
+		return;
 	}
 
 	/*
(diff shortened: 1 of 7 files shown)


src/backend/executor/nodeHash.c
```diff
@@ -3769,7 +3816,12 @@ ExecHashTableDetachBatch(HashJoinTable hashtable)
 		sts_end_parallel_scan(hashtable->batches[curbatch].outer_tuples);
 
 		/* Detach from the batch we were last working on. */
-		if (BarrierArriveAndDetach(&batch->batch_barrier))
+		/*
+		 * CBDB_PARALLEL: Parallel Hash Left Anti Semi (Not-In) Join(parallel-aware)
+		 * If phs_lasj_has_null is true, that means we have found null when building hash table,
+		 * there were no batches to detach.
+		 */
+		if (!hashtable->parallel_state->phs_lasj_has_null && BarrierArriveAndDetach(&batch->batch_barrier))
```
yjhjstz, line 3824: No need to Detach in some case ? Consider `Detach` always ?

avamingli: > No need to Detach in some case ?

Yeah, as comments: when  phs_lasj_has_null is true

> Consider Detach always ?

Can't do that, detach without attach will hang forever.

yjhjstz: since phs_lasj_has_null is not atomic , `BarrierArriveAndDetach ` maybe called when another process set phs_lasj_has_null to true, but `phs_lasj_has_null` in this process is false.

avamingli: > BarrierArriveAndDetach  maybe called when another process set phs_lasj_has_null to true

Impossible, if someone is setting phs_lasj_has_null to true, all processes are in building hash inner phase, can't arrive here.

avamingli: > since phs_lasj_has_null is not atomic , BarrierArriveAndDetach  maybe called when another process set phs_lasj_has_null to true, but phs_lasj_has_null in this process is false.

After a think: phs_lasj_has_null is atomic, but it's possible for such a extreme case:  a process doesn't participate in building hash inner and doesn't get the real-time phs_lasj_has_null though others have set it to true in shared memory( CPU L1 cache coherency).

A `pg_memory_barrier` in  MultiExecParallelHash will prevent that, I will add it, thanks.

avamingli: > A pg_memory_barrier in MultiExecParallelHash will prevent that, I will add it, thanks.

Done.

src/backend/executor/nodeHash.c
```diff
@@ -366,9 +400,23 @@ MultiExecParallelHash(HashState *node)
...
+				if (pstate->phs_lasj_has_null)
+				{
+					node->hs_hashkeys_null = true;
+					return;
+				}
 			}
 	}
 
+	/* In case we didn't participate in PHJ_BUILD_HASHING_INNER */
+	pg_memory_barrier();
```
yjhjstz, line 413: use pg_write_barrier(); ?

avamingli: pg_memory_barrier is more strict, if we write variable by ourself  pg_write_barrier() is better.

src/backend/executor/nodeHash.c
```diff
@@ -333,14 +341,40 @@ MultiExecParallelHash(HashState *node)
...
+					 * other places. As it's a boolean and we don't need to have the most recent
+					 * value from CPU or Mem cache. And we should avoid more locks in HashJion Impl.
+					 * If we miss it here and some others set it at the same time, just bypass
+					 * and we may get it at the next Hash batch.
+					 * If we missed it across all batches, we will know it when PHJ_BUILD_HASHING_INNER
+					 * ends with the help of build_barrier.
+					 * If we never participated in building hash table, check it when hash table
+					 * creation job is finished.
+					 */
+					pstate->phs_lasj_has_null = true;
```
my-ship-it, line 361: Do we need to add barrier after updating phs_lasj_has_null?

my-ship-it (approved):

src/backend/executor/nodeHash.c
avamingli: Done.

src/include/executor/hashjoin.h
```diff
@@ -286,6 +286,7 @@ typedef struct ParallelHashJoinState
 	Barrier		grow_buckets_barrier;
 	Barrier		sync_barrier;
 	Barrier		batch0_barrier;
+	volatile 	bool	phs_lasj_has_null; 	/* LASJ has found null value, identify early quit */
```
yjhjstz, line 289: should we use `pg_atomic_flag` to avoid barrier usage ?

avamingli: Sounds good, I will take a look at it.

avamingli: > should we use pg_atomic_flag to avoid barrier usage ?

@yjhjstz 
I dig into for a minute, `pg_atomic_flag` is used for atomic value for more efficiency on non-Intel platforms(is used for x86[_64]).
And it has nothing to do with barrier, we still need barrier to sync between CPU/cache if use pg_atomic_flag.
As phs_lasj_has_null is a simple bool, i think we don't need to introduce a struct with more complexity.

yjhjstz (approved):

my-ship-it (approved):
LGTM

Issue #26, opened 2023-07-20
[Bug]  Regression reindextable_while_reindex_idx_ao_part_btree failed occasionally


avamingli:
### Cloudberry Database version

commit 29374ecac2cee5b09d404dc72a6a19a4384213dd

### What happened

Flaky tests:

https://github.com/cloudberrydb/cloudberrydb/actions/runs/5597543818/jobs/10236557444

```diffs
/code/gpdb_src/src/test/isolation2/results/reindex/reindextable_while_reindex_idx_ao_part_btree.out	2023-07-19 11:24:41.248927590 +0000
@@ -40,9 +40,9 @@
 3: select distinct count(distinct relfilenode), relname from old_relfilenodes group by dbid, relname;
  count | relname                                           
 -------+---------------------------------------------------
- 1     | reindex_crtab_part_ao_btree_1_prt_p_one_id_idx    
  1     | reindex_crtab_part_ao_btree_id_idx                
  2     | reindex_crtab_part_ao_btree_1_prt_de_fault_id_idx 
+ 2     | reindex_crtab_part_ao_btree_1_prt_p_one_id_idx    
 (3 rows)
 3: COMMIT;
 COMMIT
@@ -57,8 +57,8 @@
  count | relname                                           
 -------+---------------------------------------------------
  1     | reindex_crtab_part_ao_btree_id_idx                
+ 2     | reindex_crtab_part_ao_btree_1_prt_de_fault_id_idx 
  2     | reindex_crtab_part_ao_btree_1_prt_p_one_id_idx    
- 3     | reindex_crtab_part_ao_btree_1_prt_de_fault_id_idx 
 (3 rows)
```

### How to reproduce

[icw-orca-test]

ORCA planner run regression reindextable_while_reindex_idx_ao_part_btree

### Operating System

CI [icw-orca-test]'s OS

### Anything else

The regression is under ORCA, it seems has nothing to do with parallel, I guess.
Until this commit, GUC gp_appendonly_insert_files is not enabled with enable_parallel is false. So, it should has nothing with that too.
Any reindex is a writeable operation, can't be parallel in principle.
I saw this regression failed several times at Gitlab last month.

### Are you willing to submit PR?

- [ ] Yes, I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://cloudberrydb.org/community/coc).

Ray-Eldath:
initial findings:
[...]

PR #156, opened 2023-08-18
Fix unstable tests reindextable_while_reindex_idx

Ray-Eldath:
closes: #26 

---

### Change logs

Three unstable tests matching reindex/reindextable_while_reindex_idx_*_part_btree are fixed. They failed in the same way thus could be fixed in the same manner.

These tests actually assumes when txn1 commits, its lock is acquired by txn3, and txn2 is blocked by it. Normally this is the case, but when the system load is high (e.g., when parallel is enabled, or other processes stress the system), it could be the case that the scheduler pass the access exclusive lock acquired by txn1 to txn2 (not txn3) in L22-L23, invalidate the assumption, causing the test to fail. Here I workaround this by instead of relying on the only one expected file which cannot checks two behaviour, SQL statement is used to directly check the correctness of one column.

This problem is also found in the upstream, mainline GPDB.

---

Note that this is a initial fix of one out of three total problematic cases. I'll fixed others once this passed review.


Ray-Eldath:
This PR **only fixes one out of the three** unstable tests. I'll fix all of them after this initial fix passed the review.

avamingli:
LGTM, please fix other similar cases.

Ray-Eldath:
All three cases have now been fixed/

avamingli (approved):

my-ship-it (approved):

Issue #131, opened 2023-08-09
create AO table with(parallel_workers) doesn't take effect.


avamingli:
### What happened

Alter AO's parallel_workers is ok, but set it when creating AO table doesn't take effect.

### How to reproduce

Alter AO's parallel_workers is ok, but set it when creating AO table doesn't take effect..

```sql
gpadmin=# create table ao1(c1 int) using ao_row distributed by(c1);
CREATE TABLE
gpadmin=# alter table ao1 set(parallel_workers=3);
ALTER TABLE
gpadmin=# \d+ ao1;
                                           Table "public.ao1"
 Column |  Type   | Collation | Nullable | Default | Storage | Compression | Stats target | Description
--------+---------+-----------+----------+---------+---------+-------------+--------------+-------------
 c1     | integer |           |          |         | plain   |             |              |
Compression Type: None
Compression Level: 0
Block Size: 32768
Checksum: t
Distributed by: (c1)
Access method: ao_row
Options: parallel_workers=3
```
```sql
gpadmin=# create table ao2(c1 int) using ao_row with(parallel_workers=3) distributed by(c1);
CREATE TABLE
gpadmin=# \d+ ao2;
                                           Table "public.ao2"
 Column |  Type   | Collation | Nullable | Default | Storage | Compression | Stats target | Description
--------+---------+-----------+----------+---------+---------+-------------+--------------+-------------
 c1     | integer |           |          |         | plain   |             |              |
Compression Type: None
Compression Level: 0
Block Size: 32768
Checksum: t
Distributed by: (c1)
Access method: ao_row
```

No Options: parallel_workers info of ao2.

### Operating System

Ubuntu

### Are you willing to submit PR?

- [ ] Yes, I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://cloudberrydb.org/community/coc).

PR #168, opened 2023-08-29
Fix create ao table set parallel_workers doesn't take effect

Ray-Eldath:
closes: #131 

---

### Change logs

`create table ao(c int) using ao_row with(parallel_workers=3)` doesn't take effect because only storage options are taken care of in `reloptions_gp.c:transformAOStdRdOptions`. This PR fixes this by add code processing `parallel_workers`.

The default value of `parallel_workers` (`-1`) is taken from `reloptions.c:intRelOpts`. There're no `#define`-ed constant for `parallel_workers`'s name and default value so I hard-coded them.

`transformAOStdRdOptions` actually reads `WITH` clauses **before** reads `StdRdOptions`. In the comment they say this is to preserve the order of options because some tests rely on it. I take the value from `StdRdOptions` w/o examining `WITH` clause and I see no necessity in doing so. Don't know if this could go wrong :-/

Yes. It fixes a user-perceptible bug.

### DIFF
--- a/src/test/regress/sql/gp_parallel.sql
+++ b/src/test/regress/sql/gp_parallel.sql
@@ -39,6 +39,25 @@ set search_path to test_parallel;
 -- set this to default in case regress change it by gpstop.
 set gp_appendonly_insert_files = 4;
 
+-- CBDB(#131): test parallel_workers during create AO/AOCO table take effect
+begin;
+set local enable_parallel = on;
+create table test_131_ao1(x int, y int) using ao_row with(parallel_workers=2);
+create table test_131_ao2(x int, y int) using ao_row with(parallel_workers=2);
+create table test_131_ao3(x int, y int) using ao_row with(parallel_workers=0);
+create table test_131_ao4(x int, y int) using ao_row with(parallel_workers=0);
+create table test_131_aoco1(x int, y int) using ao_column with(parallel_workers=2);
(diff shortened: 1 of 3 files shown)


Ray-Eldath:
A test case has been added. I've verified that before this fix the test would fail.

src/test/regress/sql/gp_parallel.sql
```diff
@@ -39,6 +39,19 @@ set search_path to test_parallel;
 -- set this to default in case regress change it by gpstop.
 set gp_appendonly_insert_files = 4;
 
+-- CBDB(#131): test parallel_workers during create AO/AOCO table take effect
+begin;
+set local enable_parallel = on;
+create table ao1(x int, y int) using ao_row with(parallel_workers=2);
+create table ao2(x int, y int) using ao_row with(parallel_workers=2);
+create table aoco1(x int, y int) using ao_column with(parallel_workers=2);
+create table aoco2(x int, y int) using ao_column with(parallel_workers=2);
```
yjhjstz, line 48: Add a test about parallel_workers=0 ?

Ray-Eldath: done. noted that default value for `parallel_workers` is -1. so it's actually useful to check whether everything fine under `parallel_workers=0`.

yjhjstz (approved):

avamingli (approved):

PR #169, opened 2023-08-30
Feature: add user-defined index access method

hw118118:
The purpose of the commit is in support of user-defined index access method. Currently, there are many restrictions for index access methods and new index access methods are incompatible with internal framework. At the same time, it's impossible for user to change default index access method. So I build some hooks that are for changing default behavior and being compatible for those strange restrictions.
   
For express the meaning of the modification, I show a example as below.

For example ,  I want to add 7 kinds of new index access method that based on diffrent storage engine (maybe unionstore storage) and they are correspond to internal index access methods as below:

- usbtree ---btree
- usgin --- gin
- usgist --- gist
- usspgist --- spgist
- ushash --- hash
- usbitmap --- bitmap

The main diffrence between them is index data can separate from other data and support Compute and Storage Separation.

Suppose we want to use these new functions, we can create a new extension and we can run sql （for example by `create extension unionstore;`) command and you can find new access methods : 
```
regression=# select * from pg_am;
  oid   |    amname     |         amhandler         | amtype
--------+---------------+---------------------------+--------
      2 | heap          | heap_tableam_handler      | t
    403 | btree         | bthandler                 | i
    405 | hash          | hashhandler               | i
    783 | gist          | gisthandler               | i
   2742 | gin           | ginhandler                | i
   4000 | spgist        | spghandler                | i
   3580 | brin          | brinhandler               | i
   7024 | ao_row        | ao_row_tableam_handler    | t
   7166 | ao_column     | ao_column_tableam_handler | t
   7013 | bitmap        | bmhandler                 | i
  16394 | union_store   | heap_tableam_handler      | t
  16402 | ushash        | ushashhandler             | i
  16403 | usbtree       | usbthandler               | i
  16404 | usgist        | usgisthandler             | i
  16405 | usgin         | usginhandler              | i
  16406 | usspgist      | usspghandler              | i
  16407 | usbrin        | usbrinhandler             | i
  16408 | usbitmap      | usbmhandler               | i
  22820 | heap2         | heap_tableam_handler      | t
 111831 | ao_row_testam | ao_row_tableam_handler    | t
 111847 | ao_col_testam | ao_column_tableam_handler | t
 111850 | heap_testam   | heap_tableam_handler      | t
```
So you can use these new index access methods as internal index access method. At the same time, I add new guc variable and you can correct default index type by `set default_index_type = usbtree` or show relevant info by `show default_index_type`;

In conclusion, it's more flexible and it decoupe from other part.

### DIFF
--- a/src/backend/access/brin/brin.c
+++ b/src/backend/access/brin/brin.c
@@ -1204,7 +1204,7 @@ brin_summarize_range_internal(PG_FUNCTION_ARGS)
 
 	/* Must be a BRIN index */
 	if (indexRel->rd_rel->relkind != RELKIND_INDEX ||
-		!isIndexAccessMethod(indexRel->rd_rel->relam, BRIN_AM_OID))
+		!IsIndexAccessMethod(indexRel->rd_rel->relam, BRIN_AM_OID))
 		ereport(ERROR,
 				(errcode(ERRCODE_WRONG_OBJECT_TYPE),
 				 errmsg("\"%s\" is not a BRIN index",
@@ -1290,7 +1290,7 @@ brin_desummarize_range(PG_FUNCTION_ARGS)
 
 	/* Must be a BRIN index */
 	if (indexRel->rd_rel->relkind != RELKIND_INDEX ||
-		!isIndexAccessMethod(indexRel->rd_rel->relam, BRIN_AM_OID))
+		!IsIndexAccessMethod(indexRel->rd_rel->relam, BRIN_AM_OID))
 		ereport(ERROR,
 				(errcode(ERRCODE_WRONG_OBJECT_TYPE),
 				 errmsg("\"%s\" is not a BRIN index",
--- a/src/backend/access/brin/brin.c
+++ b/src/backend/access/brin/brin.c
@@ -50,6 +50,7 @@
 #include "storage/procarray.h"
 #include "utils/snapshot.h"
 
+is_likeam_hook_type is_likebrin_hook = NULL;
 /*
  * We use a BrinBuildState during initial construction of a BRIN index.
  * The running state is kept in a BrinMemTuple.
--- a/src/backend/access/brin/brin.c
+++ b/src/backend/access/brin/brin.c
@@ -1205,7 +1205,8 @@ brin_summarize_range_internal(PG_FUNCTION_ARGS)
 
 	/* Must be a BRIN index */
 	if (indexRel->rd_rel->relkind != RELKIND_INDEX ||
-		indexRel->rd_rel->relam != BRIN_AM_OID)
+		((is_likebrin_hook && !(*is_likebrin_hook)(indexRel->rd_rel->relam)) ||
+		 (!is_likebrin_hook && indexRel->rd_rel->relam != BRIN_AM_OID)))
 		ereport(ERROR,
 				(errcode(ERRCODE_WRONG_OBJECT_TYPE),
 				 errmsg("\"%s\" is not a BRIN index",
@@ -1291,7 +1292,8 @@ brin_desummarize_range(PG_FUNCTION_ARGS)
 
 	/* Must be a BRIN index */
 	if (indexRel->rd_rel->relkind != RELKIND_INDEX ||
-		indexRel->rd_rel->relam != BRIN_AM_OID)
+		((is_likebrin_hook && !(*is_likebrin_hook)(indexRel->rd_rel->relam)) ||
+		 (!is_likebrin_hook && indexRel->rd_rel->relam != BRIN_AM_OID)))
 		ereport(ERROR,
 				(errcode(ERRCODE_WRONG_OBJECT_TYPE),
 				 errmsg("\"%s\" is not a BRIN index",
@@ -1299,7 +1301,7 @@ brin_desummarize_range(PG_FUNCTION_ARGS)
 
 	/* User must own the index (comparable to privileges needed for VACUUM) */
 	if (!pg_class_ownercheck(indexoid, GetUserId()))
-		aclcheck_error(ACLCHECK_NOT_OWNER, OBJECT_INDEX,
+		aclcheck_error(ACLCHECK_NOT_OWNER, OBJECT_INDE
 					   RelationGetRelationName(indexRel));
 
 	/*
--- a/src/backend/access/brin/brin.c
+++ b/src/backend/access/brin/brin.c
@@ -50,7 +50,6 @@
 #include "storage/procarray.h"
 #include "utils/snapshot.h"
 
-is_likeam_hook_type is_likebrin_hook = NULL;
 /*
  * We use a BrinBuildState during initial construction of a BRIN index.
  * The running state is kept in a BrinMemTuple.
@@ -1205,8 +1204,7 @@ brin_summarize_range_internal(PG_FUNCTION_ARGS)
 
 	/* Must be a BRIN index */
 	if (indexRel->rd_rel->relkind != RELKIND_INDEX ||
-		((is_likebrin_hook && !(*is_likebrin_hook)(indexRel->rd_rel->relam)) ||
-		 (!is_likebrin_hook && indexRel->rd_rel->relam != BRIN_AM_OID)))
+		!isIndexAccessMethod(indexRel->rd_rel->relam, BRIN_AM_OID))
 		ereport(ERROR,
 				(errcode(ERRCODE_WRONG_OBJECT_TYPE),
 				 errmsg("\"%s\" is not a BRIN index",
@@ -1292,8 +1290,7 @@ brin_desummarize_range(PG_FUNCTION_ARGS)
 
 	/* Must be a BRIN index */
 	if (indexRel->rd_rel->relkind != RELKIND_INDEX ||
-		((is_likebrin_hook && !(*is_likebrin_hook)(indexRel->rd_rel->relam)) ||
-		 (!is_likebrin_hook && indexRel->rd_rel->relam != BRIN_AM_OID)))
+		!isIndexAccessMethod(indexRel->rd_rel->relam, BRIN_AM_OID))
 		ereport(ERROR,
 				(errcode(ERRCODE_WRONG_OBJECT_TYPE),
 				 errmsg("\"%s\" is not a BRIN index",
--- a/src/backend/access/brin/brin.c
+++ b/src/backend/access/brin/brin.c
@@ -1301,7 +1301,7 @@ brin_desummarize_range(PG_FUNCTION_ARGS)
 
 	/* User must own the index (comparable to privileges needed for VACUUM) */
 	if (!pg_class_ownercheck(indexoid, GetUserId()))
-		aclcheck_error(ACLCHECK_NOT_OWNER, OBJECT_INDE
+		aclcheck_error(ACLCHECK_NOT_OWNER, OBJECT_INDEX,
 					   RelationGetRelationName(indexRel));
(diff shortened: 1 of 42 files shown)


src/backend/access/brin/brin.c
```diff
@@ -1204,7 +1205,8 @@ brin_summarize_range_internal(PG_FUNCTION_ARGS)
 
 	/* Must be a BRIN index */
 	if (indexRel->rd_rel->relkind != RELKIND_INDEX ||
-		indexRel->rd_rel->relam != BRIN_AM_OID)
```
HuSen8891, line 1207: use this function IsIndexAccessMethod(Oid relam, Oid IndexAccessMethod) instead ?
indexRel->rd_rel->relam != BRIN_AM_OID --> IsIndexAccessMethod(indexRel->rd_rel->relam, BRIN_AM_OID)

avamingli: +1

src/include/utils/sync_guc_name.h
```diff
@@ -17,6 +17,7 @@
 		"DateStyle",
 		"debug_discard_caches",
 		"default_table_access_method",
+        "default_index_type",
```
avamingli, line 20: indent

src/backend/utils/misc/guc.c
```diff
@@ -4078,6 +4079,17 @@ static struct config_string ConfigureNamesString[] =
 		check_default_table_access_method, NULL, NULL
 	},
 
+    {
+            {"default_index_type", PGC_USERSET, CLIENT_CONN_STATEMENT,
+                    gettext_noop("Sets the default index type for primary key."),
```
avamingli, line 4084: Is it limited to ` primary key.`?

src/backend/utils/cache/lsyscache.c
```diff
@@ -866,8 +866,10 @@ equality_ops_are_compatible(Oid opno1, Oid opno2)
 		Form_pg_amop op_form = (Form_pg_amop) GETSTRUCT(op_tuple);
 
 		/* must be btree or hash */
-		if (op_form->amopmethod == BTREE_AM_OID ||
-			op_form->amopmethod == HASH_AM_OID)
+		if (((is_likebtree_hook && (*is_likebtree_hook)(op_form->amopmethod)) ||
+             (!is_likebtree_hook && op_form->amopmethod == BTREE_AM_OID)) ||
+            ((is_likehash_hook && (*is_likehash_hook)(op_form->amopmethod)) ||
+             (!is_likehash_hook && op_form->amopmethod == HASH_AM_OID)))
```
avamingli, line 872: I prefer to use a function too.

src/backend/mock.mk
```diff
@@ -38,9 +38,6 @@ EXCL_OBJS=\
 # of the test programs. Feel free to link them back (i.e. remove them from
 # this exclusion list) as needed.
 EXCL_OBJS+=\
-	src/backend/access/hash/hash.o \
-	src/backend/access/hash/hashsearch.o \
-	\
```
avamingli, line 43: Why we remove these

avamingli (commented):
Shall we add cases to test new hooks?

src/backend/mock.mk
hw118118: Because hash index access method has changed some, so for run test succssfully, it has to remove from exclude test files.

src/backend/access/brin/brin.c
hw118118: Thank for @HuSen8891 @avamingli. I had made a change, you can review then again.

src/backend/access/index/amapi.c
```diff
@@ -20,7 +20,18 @@
...
+bool
+isIndexAccessMethod(Oid relam, Oid indexAccessMethod)
+{
+	if ((is_same_index_am_hook && (*is_same_index_am_hook)(indexAccessMethod, relam)) ||
+		(!is_same_index_am_hook && relam == indexAccessMethod))
+	{
+		return true;
+	}
+	return false;
+}
```
avamingli, line 34: Could this simply to:
```c
if (relam == indexAccessMethod)
  return true;
return (is_same_index_am_hook && (*is_same_index_am_hook)(indexAccessMethod, relam);
```
The key difference is: 
relam is  indexAccessMethod and is_same_index_am_hook is not null, could it happen? And which one  has the priority?

src/include/catalog/pg_index.h
```diff
@@ -73,6 +73,8 @@ FOREIGN_KEY(indrelid REFERENCES pg_class(oid));
  * ----------------
  */
 typedef FormData_pg_index *Form_pg_index;
+typedef bool (*is_likeam_hook_type)(Oid expected_oid, Oid id);
```
HuSen8891, line 76: typedef bool (* IsIndexAccessMethod_hook_type) (Oid relam, Oid indexAccessMethod);
extern PGDLLIMPORT IsIndexAccessMethod_hook_type IsIndexAccessMethod_hook;

src/backend/access/index/amapi.c
```diff
@@ -20,7 +20,18 @@
 #include "utils/builtins.h"
 #include "utils/syscache.h"
 
+is_likeam_hook_type is_same_index_am_hook = NULL;
 
+bool
+isIndexAccessMethod(Oid relam, Oid indexAccessMethod)
```
HuSen8891, line 26: isIndexAccessMethod --> IsIndexAccessMethod

src/backend/access/index/amapi.c
hw118118: Theoretically，when is_same_index_am_hook is not null, it should include relam == indexAccessMethod, but it's hard to make decision for user, so I think is_same_index_am_hook has higher priority.

src/include/catalog/pg_index.h
hw118118: ‘is’ is a verb, so I think it is more proper with lower-case and refer other hook in cbdb core kernel .

src/backend/access/gin/ginutil.c
```diff
@@ -29,7 +29,6 @@
 #include "utils/index_selfuncs.h"
 #include "utils/typcache.h"
 
-
```
avamingli, line 32: Some times our IDE will delete extra blank like this one. And we usually choose not to change it. 
The Postgres will run pgindent at fixed periods to resolve code style, we will get unnecessary conflicts when Merge to next version. That's painful.

avamingli:
Hi,  I test the branch with the added GUC, there are sth we need to fix.

1. set `default_index_type`, and create a new index, the GUC doesn't take effect.
We set it to brin, but new created index is also btree.
```sql
gpadmin=# set default_index_type = brin;
SET
gpadmin=# show default_index_type;
 default_index_type
--------------------
 brin
(1 row)

gpadmin=# create index on t2(c1);
CREATE INDEX
gpadmin=# \d+ t2;
                                           Table "public.t2"
 Column |  Type   | Collation | Nullable | Default | Storage | Compression | Stats target | Description
--------+---------+-----------+----------+---------+---------+-------------+--------------+-------------
 c1     | integer |           |          |         | plain   |             |              |
 c2     | integer |           |          |         | plain   |             |              |
Indexes:
    "t2_c1_idx" btree (c1)
Distributed by: (c1)
Access method: heap
```
2. We lack of test when setting this GUC to check the invalidation.
User could set a not-exists value, it's risky.
```sql
gpadmin=# select * from pg_am;
 oid  |  amname   |         amhandler         | amtype
------+-----------+---------------------------+--------
    2 | heap      | heap_tableam_handler      | t
  403 | btree     | bthandler                 | i
  405 | hash      | hashhandler               | i
  783 | gist      | gisthandler               | i
 2742 | gin       | ginhandler                | i
 4000 | spgist    | spghandler                | i
 3580 | brin      | brinhandler               | i
 7024 | ao_row    | ao_row_tableam_handler    | t
 7166 | ao_column | ao_column_tableam_handler | t
 7013 | bitmap    | bmhandler                 | i
(10 rows)

gpadmin=# set default_index_type = not_exist;
SET
gpadmin=# show default_index_type;
 default_index_type
--------------------
 not_exist
(1 row)

gpadmin=# create table t3(c1 int unique, c2 int);
ERROR:  access method "not_exist" does not exist
```
3. We changed the codes when creating a table in parse stage.
But doesn't have a case to check, like
```sql
create unique index on t3(c1);
```

And as the GUC is added by CBDB, we should put it in guc_gp.c, not guc.c
I see the GUC is only used at parse stage,  does it need to sync to segments(src/include/utils/sync_guc_name.h)?

src/backend/parser/parse_utilcmd.c
```diff
@@ -3413,7 +3413,7 @@ transformIndexConstraint(Constraint *constraint, CreateStmtContext *cxt)
 		 * else dump and reload will produce a different index (breaking
 		 * pg_upgrade in particular).
 		 */
-		if (index_rel->rd_rel->relam != get_index_am_oid(DEFAULT_INDEX_TYPE, false))
+		if (index_rel->rd_rel->relam != get_index_am_oid(default_index_access_method, false))
```
HuSen8891, line 3416: Suppose this case:

create table test(c1 int);
create unique index on test using btree(c1);
set default_index_access_method = 'brin';
alter table test add CONSTRAINT abc2 unique using index test_c1_idx;

The 'add constraint' statement fails here, and reports the ERROR 'index "test_c1_idx" is not a btree'，but index 'test_c1_idx' is indeed a btree.

src/include/catalog/pg_index.h
HuSen8891: Refer to 'planner_hook'、'ProcessUtility_hook'、'ExecutorRun_hook' and other hooks, we should use 'IsIndexAccessMethod_hook_type' instead of 'is_likeam_hook_type', and 'IsIndexAccessMethod_hook' instead of 'is_same_index_am_hook'.

src/backend/parser/parse_utilcmd.c
hw118118: Its a bug and change to !IsIndexAccessMethod(BTREE_AM_OID, index_rel->rd_rel->relam)

src/include/catalog/pg_index.h
HuSen8891: The hook's parameters should be the same as the function by type and order, this could avoid logical error.

avamingli:
Add some test cases?

hw118118:
> Add some test cases?

There are many test cases but place into extension UnionStore, because it mainly comply with UnionStore extension.

src/backend/optimizer/path/indxpath.c
```diff
@@ -3529,7 +3529,7 @@ ec_member_matches_indexcol(PlannerInfo *root, RelOptInfo *rel,
 	 * generate_implied_equalities_for_column; see
 	 * match_eclass_clauses_to_index.
 	 */
-	if (index->relam == BTREE_AM_OID &&
+	if (IsIndexAccessMethod(index->relam, BTREE_AM_OID) &&
```
avamingli, line 3532: This can't be simply replaced by  IsIndexAccessMethod, see the comments above.

Not everywhere could be replaced by `IsIndexAccessMethod` too,  could you check other places too?

hw118118: yes, before I made a research in all places in kernel and checked them multi times and tested them correctly.

HuSen8891 (approved):

avamingli (approved):

PR #77, opened 2023-07-24
Feat/support singlenode mode

wfnuser:
---

Close #69 

### Change logs

This pr introduces a new deploy mode: singlenode mode. In singlenode mode, cbdb is deployed as a single node service and we will set the gp_role to 'singlenode'.

We introduce a bunch of new tests and scripts for singlenode in this pr.
For the core part, we branching the logic carefully by using the flag of gp_role. These changes are scattered in the whole project.

We also cherry-pick two bug-fix commits from upstream. Since our test for cbdb singlenode is based on a bit newer PG14 than cbdb v1.0 was rebased from.

### Why are the changes needed?

CBDB users need singlenode deploy mode.

### Does this PR introduce any user-facing change?

Yes. Now our users can deploy cbdb in single node. You can create a demo service by using the following command:
```bash
make create-demo-singlenode
```

More details can be referenced in commit messages.

### How was this patch tested?

Pass ICW test in cluster mode.
Pass 'make installcheck-singlenode' in single node mode. It runs singlenode mode tests we introduced in this pr.

### Contributor's Checklist
Here are some reminders before you submit the pull request:
* Document changes
* Communicate in the [GitHub Issues](https://github.com/cloudberrydb/cloudberrydb/issues) or [Discussions](https://github.com/orgs/cloudberrydb/discussions) (list them if needed)
* Add tests for the change
* Pass `make installcheck`
* Pass `make -C src/test installcheck-cbdb-parallel`

PR #206, opened 2023-09-19
SingleNode deployment

### LINKED ISSUE #53: Deploy CBDB as a single node service

### LINKED ISSUE #69: CBDB as a single node service

### LINKED ISSUE #100: support gpstart in singlenode mode without parameter "singlenodeMode"
When we using tools like gpstart\gpstop\gpconfig, we rely the cluster alive to fetch cluster info from it to decide if we're in singlenode mode or cluster mode.
However, if we kill the cluster by gpstop, it's a little hard to get info from the dead service. Thus, when we use gpstart to restart the stopped singlenode service, we need to add "--singlenodeMode" to control the behavior of gpstart.

An idea is to write mode info in postgres conf file? Not sure. Support it in the future.

### LINKED ISSUE #123: Disable utility mode in singlenode deployment
Utilty mode is not necessary in singlenode deployment.
But if we disable directly, many of the isolaiton tests will be broken. "1U:" grammar require connection to be set in utility mode.

Can be fixed by this.
```c
void
assign_gp_role(const char *newval, void *extra)
{
	/*
	 * Since postgres don't have utility mode, singlenode mode should not support
	 * utility mode either.
	 */
	if (Gp_role == GP_ROLE_SINGLENODE && string_to_role(newval) == GP_ROLE_UTILITY)
	{
		ereport(ERROR,
				(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
					errmsg("utility mode is not supported in ")));
	}

	Gp_role = string_to_role(newval);

	if (IS_UTILITY_OR_SINGLENODE(Gp_role) && MyProc != NULL)
		MyProc->mppIsWriter = false;
}
```

Singlenode role should not be used to connect to any of the cluster node either. Can be taken into consideration along with this issue.

Ray-Eldath:
closes #69, closes #53, closes #100, closes #123

---

### Change logs

This PR implements singlenode deployment. All code, scripts and tests are incorporated.

**Please see the full design proposal: https://github.com/orgs/cloudberrydb/discussions/188#discussion-5593615** which has already been discussed with and approved by reviewers.

~~Commit messages're still in a mess. I'll squash them into a nice write-up after we finish reviewing.~~

This PR is heavily based on @wfnuser's previous work (#77).

Please refer to the *Motivation* section of the design proposal.


avamingli:
> Commit messages're still in a mess. I'll squash them into a nice write-up after we finish reviewing.

Hi, thanks for your contribution.
Commit messages  are also part of review process, please squash commits as CBDB codes change/All tests added/PG commits cherry-picked. 
Review by commits is helpful for reviewers  as this pr have so many test files.

src/test/regress/pg_regress.c
```diff
@@ -940,7 +940,7 @@ convert_sourcefiles_in(const char *source_subdir, const char *dest_dir, const ch
 static void
 convert_sourcefiles(void)
 {
-	content_zero_hostname = get_host_name(-1, 'p');
+	content_master_hostname = get_host_name(-1, 'p');
```
avamingli, line 943: The upstream codes are:
> content_zero_hostname = get_host_name(0, 'p');  

Not sure why it doesn't show the 0 -> -1 differences.

Better not to change them `content_zero_hostname`,  if we are singlenode, get_host_name(-1, 'p'), else doesn't need to change.

src/include/executor/executor.h
```diff
@@ -113,7 +113,7 @@ extern bool ExecMaterializesOutput(NodeTag plantype);
 /*
  * prototypes from functions in execCurrent.c
  */
```
avamingli, line 115: Add comments the meaning of return true  or false.

src/backend/utils/misc/guc_gp.c
```diff
@@ -2983,6 +2983,17 @@ struct config_bool ConfigureNamesBool_gp[] =
 		NULL, NULL, NULL
 	},
 
+	{
+		{"gp_internal_is_singlenode", PGC_POSTMASTER, UNGROUPED,
+			 gettext_noop("is in SingleNode mode or not. WARNING: user SHOULD NOT set this by any means."),
```
avamingli, line 2988: `is in SingleNode mode or not.` -> `Is in SingleNode mode(no segments).`

src/backend/postmaster/postmaster.c
```diff
@@ -1119,7 +1119,7 @@ PostmasterMain(int argc, char *argv[])
 		ereport(ERROR,
 				(errmsg("WAL streaming (max_wal_senders > 0) requires wal_level \"replica\" or \"logical\"")));
 
-    if ( GpIdentity.dbid == -1 && Gp_role == GP_ROLE_UTILITY)
+	if ( GpIdentity.dbid == -1 && Gp_role == GP_ROLE_UTILITY)
```
avamingli, line 1122: Rollback

src/backend/executor/execCurrent.c
```diff
@@ -217,14 +223,34 @@ getCurrentOf(CurrentOfExpr *cexpr,
...
+				}
+			}
+		}
+
 		/*
 		 * The target relation must directly match the cursor's relation. This throws out
 		 * the simple case in which a cursor is declared against table X and the update is
-		 * issued against Y. Moreover, this disallows some subtler inheritance cases where
-		 * Y inherits from X. While such cases could be implemented, it seems wiser to
-		 * simply error out cleanly.
```
avamingli, line 225: According to the comments of this function :
>  GPDB calls it before dispatching to make QEs get the same current position
>  of the cursor.

Do we need to change here? we won't dispatch in singlenode. Enable this in utility is another topic and not sure for its benefit or risk.

avamingli (commented):
Most codes seems good.

src/include/executor/executor.h
Ray-Eldath: I add one comment at the definition site.

src/test/regress/pg_regress.c
Ray-Eldath: you means we should probe whether the server is in singlenode mode and then `get_host_name(-1, 'p')` and keep the rest of code the same?

avamingli: > you means we should probe whether the server is in singlenode mode and then get_host_name(-1, 'p') and keep the rest of code the same?

yeah

Ray-Eldath: `initialize_environment` which calls `convert_sourcefiles` happens very early in pg_ctl's initialization, during which the database even havn't started yet. I have to take `convert_sourcefiles` and call it just before running the actual tests. dunno if it's okay.

see 27757d781fff1f324f59d7237b0f3e60f529374e

src/backend/commands/vacuum.c
```diff
@@ -1636,7 +1636,7 @@ vac_update_relstats(Relation relation,
 	 * Because there's a chance that we overwrite perfectly good stats with zeros
 	 */
 
-	bool ifUpdate = ! (IS_QUERY_DISPATCHER() && Gp_role == GP_ROLE_UTILITY);
+	bool ifUpdate = ! (IS_QUERY_DISPATCHER() && Gp_role == GP_ROLE_UTILITY) || IS_SINGLENODE();
```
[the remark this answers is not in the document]
Ray-Eldath: i'm not sure... this is the root cause of the _plan diff_ because statistics aren't updated on QD (now an active segment) during `analyze`.

src/backend/commands/tablecmds.c
```diff
@@ -18774,6 +18785,12 @@ make_distributedby_for_rel(Relation rel)
 
 	dist = makeNode(DistributedBy);
 
+	if (Gp_role == GP_ROLE_UTILITY)
+	{
+		Assert(policy->ptype == POLICYTYPE_ENTRY);
+		return NULL;
+	}
+
```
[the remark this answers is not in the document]
Ray-Eldath: this code is only called when additional info is needed to decide query's distribution policy (like, when it's inherited from patrent table). it doesn't sits in the normal code path, because normally the `DISTRIBUTED BY` would already be parsed by the parser and stored in `CreateStmt->distributedBy`. I guess this is for https://github.com/cloudberrydb/cloudberrydb/pull/206#discussion_r1335544239 because utility supports inheritance now and this may be called in the case.

src/backend/access/table/table.c
```diff
@@ -209,8 +209,19 @@ CdbTryOpenTable(Oid relid, LOCKMODE reqmode, bool *lockUpgraded)
...
+	 * However, it will cause behavior difference in `parallel_update_readcommitted`.
+	 * The second process in that case want to update the table is blocked but there is
+	 * `duplicate key value violates unique constraint "pg_aovisimap_95435_index"`
+	 * error after the first process commit.
+	 * Need to figure it out later. Currently, let's just use the stronger
+	 * lock as cbdb does.
+	 */
 	if (reqmode == RowExclusiveLock &&
-		Gp_role == GP_ROLE_DISPATCH &&
+		(Gp_role == GP_ROLE_DISPATCH || IS_SINGLENODE()) &&
```
Ray-Eldath, line 224: this line is buggy in currrent implementation or if we disabled utility login in "singlenode as a new role" scheme. `isolation` test suite use utility mode to login. in normal cases, `lockmode` are passed in as-is. but because of the failure in `uao/parallel_update_readcommitted`, locks are elevated from `RowExclusiveLock` to `ExclusiveLock` under singlenode mode, casing unnecessary blocking in `isolation` test suites. but if we remove this elevation behaviour, `uao/parallel_update_readcommitted` in `singlenode_isolation2` will failed. we could either disable tests in `isolation`, or `uao/parallel_update_readcommitted` in `singlenode_isolation2`, or trying to figure out why the latter failed. what do you think? @avamingli

avamingli: It sounds more like a bug, we'd better figure out.

Ray-Eldath: fixed in https://github.com/cloudberrydb/cloudberrydb/pull/206/commits/5e3d229d478a4ba4bddcab3d70baa8b8c614630a.

src/backend/commands/tablecmds.c
```diff
@@ -5314,6 +5319,12 @@ ATPrepCmd(List **wqueue, Relation rel, AlterTableCmd *cmd,
 			pass = AT_PASS_MISC;
 			break;
 		case AT_SetDistributedBy:	/* SET DISTRIBUTED BY */
+			/* Setting distributed by is meaningless in utility mode. */
+			if (Gp_role == GP_ROLE_UTILITY)
+			{
+				pass = AT_PASS_MISC;
+				break;
+			}
```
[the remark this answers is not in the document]
Ray-Eldath: i think guard against utility and singlenode mode is fair here. it prevents future misuse of this function in utility mode as well. i suppose the owner of this code just forget to guard against utility mode in the very beginning, and we're here closing the loophole.

src/backend/commands/copy.c
```diff
@@ -778,6 +778,11 @@ ProcessCopyOptions(ParseState *pstate,
 		}
 		else if (strcmp(defel->defname, "on_segment") == 0)
 		{
+			/* copy on segment is meaningless in utility mode or singlenode mode. */
```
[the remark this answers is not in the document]
Ray-Eldath: save as https://github.com/cloudberrydb/cloudberrydb/pull/206#discussion_r1351288405. I can revert the changes if you think my explanation doesn't sounds convincing.

src/backend/executor/execCurrent.c
Ray-Eldath: test `portal` in `regress` suite tests inheritance cases at https://github.com/cloudberrydb/cloudberrydb/blob/10538adb70ac5e81abbe0cb16a9e1cf66ad808ec/src/test/regress/sql/portals.sql#L392. we need this to pass the test. or we have to disable this part of it.

src/backend/executor/execCurrent.c
```diff
@@ -93,8 +93,14 @@ execCurrentOf(CurrentOfExpr *cexpr,
...
 	else
 	{
-		getCurrentOf(cexpr, econtext, table_oid, current_tid,
+		bool res = getCurrentOf(cexpr, econtext, table_oid, current_tid,
 					 &current_gp_segment_id, &current_table_oid, NULL);
+		/*
+		 * In utility mode, we don't bother to check segment match.
+		 * And if we have partitioned table, we will do the table oid check inside getCurrentOf.
+		 */
+		if (Gp_role == GP_ROLE_UTILITY)
```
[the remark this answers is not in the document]
Ray-Eldath: this change is to support https://github.com/cloudberrydb/cloudberrydb/pull/206#discussion_r1351353991. do you need we should narrow the inheritance support to only singlenode mode not utility mode? if so two `if` here and there need to be changed together.

Ray-Eldath:
really can't replicate the failure of https://github.com/cloudberrydb/cloudberrydb/actions/runs/6467257609/job/17557480521?pr=206 on my machine even i've already enabled xml support with `--with-libxml`

Ray-Eldath:
finally all ci pipeline (including the newly-added `ic-singlenode-test` are passed. we may move on to the second stage of reviewing? @avamingli

avamingli:
> finally all ci pipeline (including the newly-added `ic-singlenode-test` are passed, and i've processed review comments and squashed commits. we may move on to the second stage of reviewing? @avamingli

I think this pr is very close to be merged, thanks for your work.

Few refinements of commit messages  I found:

```shell
Since GPDB isn't fully compatible with Postgres, many features and
grammars have been dependent upon. In a scenario where only one node is
needed but GPDB is already relied by business logic, a singlenode mode
can be used. Utility mode is different from singlenode because it still
requires segments be created, it just connects to one of them in a
special mode designated by PGOPTIONS. On the contary, singlenode mode is
fully supported by all deploying scripts, user can create a running node
that have no segment and directly connect to it as if it were a normal
cluster. Plus, there're some functionalities that aren't enabled in
utility mode but should be in singlenode.
```
`On the contary` 

```shell
This new mode is supported by adding a new GUC gp_internal_is_singlenode.
We've settled to leverage the already-written code to support
GP_ROLE_UTILITY by running singlenode deployment in GP_ROLE_UTILITY.
This way seems leads to the minimum code change. There're several places
where special care are needed: code that should run only under
normal utility mode but not in singlenode are taken care of by
IS_UTILITY_BUT_NOT_SINGLENODE() macro, and, conversely, code that should
only run in singlenode are handled by
Gp_role == GP_ROLE_DISPATCH || IS_SINGLENODE(). For the remaining code
changes, they're all caused by several global replacements.

Scripts has been changed to support singlenode deployment. When creating
demo cluster, use NUM_PRIMARY_MIRROR_PAIRS=0 to create singlenode
deployment. It'll write the GUC in postgresql.conf and only start QD and
one standby. gpstart and gpstop behaves normally because singlenode mode
is controlled by a GUC that only initialized in gpinitsystem. We
delibrately pick a long name because we don't want user to accidentally
change this after cluster initialization.
```
replace all `GPDB` -> `CBDB`

```shell
See: Discussion#188 <https://github.com/orgs/cloudberrydb/discussions/188>

add CI for singlenode mode

add CI for singlenode mode
```
duplicated messages.

src/backend/commands/tablecmds.c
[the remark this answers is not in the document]
Ray-Eldath: places where `make_distributedby_for_rel` are called:
- `alter table set distributed by`: not supported in utility
- `alter table expand`: ditto
- `create table distributed by`: ditto
- `create table partition by`: when creating partitioned table in utility mode (yes, singlenode already supports partitioned table), because `distributed by` is meaningless, there's no need to deduce distribution policy from the clause, we can always return a NULL. and because creating table **under utility mode** is the only case when this function may be called, we can safely assert that `policy->ptype` is `ENTRY`. as described above, there won't be a case where table get created in cluster mode but altered in singlenode or utility mode.

src/backend/commands/copy.c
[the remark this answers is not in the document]
Ray-Eldath: reverted per discussed.

avamingli (dismissed):
LGTM.

Ray-Eldath:
@gfphoenix78 IIRC this Monday you propose that we should rename `singlenode` to `utility_singlenode` and `utility` to `utility_maintenance`. I think this is a great idea. I'm planning to rename all `singlenode` -> `utility_singlenode` **in code** *and* rename both `singlenode` and `utility` **in documentation**. what's your idea? do you come up with a better name? (so far I think the name `utility_singlenode` and `utility_maintenance` are both good and acceptable.)

avamingli:
The CI failed @Ray-Eldath , need a rebase?

avamingli (approved):

yjhjstz:
how to install single node with rpm package ?

gpMgmt/bin/gppylib/gparray.py
```diff
@@ -959,6 +959,9 @@ def initFromCatalog(dbURL, utility=False):
             if not version.isVersionCurrentRelease():
                 raise Exception("Cannot connect to GPDB version %s from installed version %s"%(version.getVersionRelease(), MAIN_VERSION[0]))
 
+            # Get gp_internal_is_singlenode GUC
+            is_singlenode = dbconn.querySingleton(conn, "SHOW gp_internal_is_singlenode") == "on"
```
yjhjstz, line 963: what about config to true ?

src/test/isolation/specs/udf-insert-deadlock.spec
```diff
@@ -10,8 +10,8 @@
 
 setup
 {
-	CREATE TABLE udf_dl_one (a int, b int) DISTRIBUTED BY (a);
```
yjhjstz, line 13: why remove `DISTRIBUTED BY (a)`

gpMgmt/bin/gppylib/gparray.py
Ray-Eldath: the variable `is_singlenode` is used in `gpstart`. when set to true, after `_startCoordinator()` and `_start_standby()` in the Phase 1 (start everything in utility mode) the script exited immediately:

image

more on phase 1 & 2 of gpstart: https://github.com/orgs/cloudberrydb/discussions/188

yjhjstz: but true is equal to on .

src/test/isolation/specs/udf-insert-deadlock.spec
Ray-Eldath: you're right. this is unnecessary. reverted.

gpMgmt/bin/gppylib/gparray.py
Ray-Eldath: you mean `SHOW gp_internal_is_singlenode` could returns `true` instead of ` on`?

Ray-Eldath:
> how to install single node with rpm package ? think about install like pg style: `initdb`, then `pg_ctl start ` @Ray-Eldath

I've tried with proper `postmaster.conf` using `env GPSESSID=0000000000 GPERA=f6a2f666f34494b3_230907184019 $GPHOME/bin/pg_ctl -D /home/gpadmin/cbdb/gpAux/gpdemo/datadirs/standby -l /home/gpadmin/cbdb/gpAux/gpdemo/datadirs/standby/log/startup.log -w -t 600 -o " -p 7001" start` and `env GPSESSID=0000000000 GPERA=f6a2f666f34494b3_230907184019 $GPHOME/bin/pg_ctl -D /home/gpadmin/cbdb/gpAux/gpdemo/datadirs/qddir/demoDataDir-1 -l /home/gpadmin/cbdb/gpAux/gpdemo/datadirs/qddir/demoDataDir-1/log/startup.log -w -t 600 -o " -p 7000 -c gp_role=utility" start` can successfully create a singlenode deployment. but the recommended approach is still to use `gpinitsystem --singlenodeMode` to start one coordinator and one coordinator standby.

yjhjstz:
> > how to install single node with rpm package ? think about install like pg style: `initdb`, then `pg_ctl start ` @Ray-Eldath
> 
> I've tried with proper `postmaster.conf` using `env GPSESSID=0000000000 GPERA=f6a2f666f34494b3_230907184019 $GPHOME/bin/pg_ctl -D /home/gpadmin/cbdb/gpAux/gpdemo/datadirs/standby -l /home/gpadmin/cbdb/gpAux/gpdemo/datadirs/standby/log/startup.log -w -t 600 -o " -p 7001" start` and `env GPSESSID=0000000000 GPERA=f6a2f666f34494b3_230907184019 $GPHOME/bin/pg_ctl -D /home/gpadmin/cbdb/gpAux/gpdemo/datadirs/qddir/demoDataDir-1 -l /home/gpadmin/cbdb/gpAux/gpdemo/datadirs/qddir/demoDataDir-1/log/startup.log -w -t 600 -o " -p 7000 -c gp_role=utility" start` can successfully create a singlenode deployment. but the recommended approach is still to use `gpinitsystem --singlenodeMode` to start one coordinator and one coordinator standby.

can we change default `datadirs/qddir` to make singlenode dir ?

Ray-Eldath:
> > > how to install single node with rpm package ? think about install like pg style: `initdb`, then `pg_ctl start ` @Ray-Eldath
> > 
> > 
> > I've tried with proper `postmaster.conf` using `env GPSESSID=0000000000 GPERA=f6a2f666f34494b3_230907184019 $GPHOME/bin/pg_ctl -D /home/gpadmin/cbdb/gpAux/gpdemo/datadirs/standby -l /home/gpadmin/cbdb/gpAux/gpdemo/datadirs/standby/log/startup.log -w -t 600 -o " -p 7001" start` and `env GPSESSID=0000000000 GPERA=f6a2f666f34494b3_230907184019 $GPHOME/bin/pg_ctl -D /home/gpadmin/cbdb/gpAux/gpdemo/datadirs/qddir/demoDataDir-1 -l /home/gpadmin/cbdb/gpAux/gpdemo/datadirs/qddir/demoDataDir-1/log/startup.log -w -t 600 -o " -p 7000 -c gp_role=utility" start` can successfully create a singlenode deployment. but the recommended approach is still to use `gpinitsystem --singlenodeMode` to start one coordinator and one coordinator standby.
> 
> can we change default `datadirs/qddir` to make singlenode dir ?

we can arbitrarily change dir of the coordinator and the coordinator standby. when using `gpinitsystem`, env `COORDINATOR_DATADIR` can be set, and when using `pg_ctl`, just change `-D` argument.

yjhjstz:
> > > > how to install single node with rpm package ? think about install like pg style: `initdb`, then `pg_ctl start ` @Ray-Eldath
> > > 
> > > 
> > > I've tried with proper `postmaster.conf` using `env GPSESSID=0000000000 GPERA=f6a2f666f34494b3_230907184019 $GPHOME/bin/pg_ctl -D /home/gpadmin/cbdb/gpAux/gpdemo/datadirs/standby -l /home/gpadmin/cbdb/gpAux/gpdemo/datadirs/standby/log/startup.log -w -t 600 -o " -p 7001" start` and `env GPSESSID=0000000000 GPERA=f6a2f666f34494b3_230907184019 $GPHOME/bin/pg_ctl -D /home/gpadmin/cbdb/gpAux/gpdemo/datadirs/qddir/demoDataDir-1 -l /home/gpadmin/cbdb/gpAux/gpdemo/datadirs/qddir/demoDataDir-1/log/startup.log -w -t 600 -o " -p 7000 -c gp_role=utility" start` can successfully create a singlenode deployment. but the recommended approach is still to use `gpinitsystem --singlenodeMode` to start one coordinator and one coordinator standby.
> > 
> > 
> > can we change default `datadirs/qddir` to make singlenode dir ?
> 
> we can arbitrarily change dir of the coordinator and the coordinator standby. when using `gpinitsystem`, env `COORDINATOR_DATADIR` can be set, and when using `pg_ctl`, just change `-D` argument.

`COORDINATOR_DATADIR` can only change `datadirs`, but I want to rename `qddir` to singlenode dir.

Ray-Eldath:
> > > > > how to install single node with rpm package ? think about install like pg style: `initdb`, then `pg_ctl start ` @Ray-Eldath
> > > > 
> > > > 
> > > > I've tried with proper `postmaster.conf` using `env GPSESSID=0000000000 GPERA=f6a2f666f34494b3_230907184019 $GPHOME/bin/pg_ctl -D /home/gpadmin/cbdb/gpAux/gpdemo/datadirs/standby -l /home/gpadmin/cbdb/gpAux/gpdemo/datadirs/standby/log/startup.log -w -t 600 -o " -p 7001" start` and `env GPSESSID=0000000000 GPERA=f6a2f666f34494b3_230907184019 $GPHOME/bin/pg_ctl -D /home/gpadmin/cbdb/gpAux/gpdemo/datadirs/qddir/demoDataDir-1 -l /home/gpadmin/cbdb/gpAux/gpdemo/datadirs/qddir/demoDataDir-1/log/startup.log -w -t 600 -o " -p 7000 -c gp_role=utility" start` can successfully create a singlenode deployment. but the recommended approach is still to use `gpinitsystem --singlenodeMode` to start one coordinator and one coordinator standby.
> > > 
> > > 
> > > can we change default `datadirs/qddir` to make singlenode dir ?
> > 
> > 
> > we can arbitrarily change dir of the coordinator and the coordinator standby. when using `gpinitsystem`, env `COORDINATOR_DATADIR` can be set, and when using `pg_ctl`, just change `-D` argument.
> 
> `COORDINATOR_DATADIR` can only change `datadirs`, but I want to rename `qddir` to singlenode dir.

`gpinitsystem` accepts a `clusterConfigFile` where you can arbitrarily set `COORDINATOR_DIRECTORY` and `DATA_DIRECTORY` to put cluster data anywhere you want. this file (and the paths) is generated by `demo_cluster.sh`.

If we want singlenode deployment to have `gpAux/gpdemo/datadirs/singlenodedir` instead of `gpAux/gpdemo/datadirs/qddir`, we need to take care of this in `demo_cluster.sh`. Is this what you want? I think this is a great idea.

yjhjstz:
> 
> `gpinitsystem` accepts a `clusterConfigFile` where you can arbitrarily set `COORDINATOR_DIRECTORY` and `DATA_DIRECTORY` to put cluster data anywhere you want. this file (and the paths) is generated by `demo_cluster.sh`.
> 
> If we want singlenode deployment to have `gpAux/gpdemo/datadirs/singlenodedir` instead of `gpAux/gpdemo/datadirs/qddir`, we need to take care of this in `demo_cluster.sh`. Is this what you want? I think it's actually a great idea.

yes, make different dir.

Ray-Eldath:
> yes, make different dir.

fixed.

yjhjstz (approved):

avamingli (approved):

avamingli (approved):

PR #233, opened 2023-10-11
Limit CI concurrency to one per PR

Ray-Eldath:
In current workflow, when newer commit comes, old pipeline won't get canceled, which wastes resources and cause unnecessary CI failure emails. This PR limits concurrency to one per PR. when newer commit comes, stale CI jobs will be automatically cancelled.

See: https://stackoverflow.com/a/72408109/8711684

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -4,7 +4,11 @@ on:
     types: [opened, synchronize, reopened, ready_for_review]
     branches:    
       - main
-  
+
+concurrency:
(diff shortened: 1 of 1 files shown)


Ray-Eldath:
I've tested in #206 that this approach takes effect. stale jobs get cancelled automatically.

image

avamingli:
Sounds good.

my-ship-it (approved):

baotingfang (approved):
LGTM

Ray-Eldath:
how about merge this to make our ci pipeline more efficient? :-) @my-ship-it

PR #282, opened 2023-11-03
Reduce external-fts CI pipeline to only icw-test

Ray-Eldath:
per discussion with @my-ship-it and @avamingli, Code taking care of external-fts are largely identical to that of internal-fts, so two redundant pipelines are removed. Besides, CI is really unstable these days, redundant pipeline can cause PR to be unnecessarily blocked.


my-ship-it (approved):
LGTM

avamingli (approved):

PR #291, opened 2023-11-08
Provide gpdemo.sh deployment script

Ray-Eldath:
Customers have complained that the deployment procedures of CBDB is tedious. In this PR, gpdemo.sh provides a default configuration to enable a one-click deployment experience. The default parameters are the same as "make create-demo-cluster", and all demo cluster Makefile jobs are now directly run "gpdemo.sh" instead.

---

### Packaging Instructions

Following files are necessary and need to be packaged into `.rpm` files if we need to ship this script in the releases:
- in directory `gpAux/gpdemo`:
  - `*.sh` (`demo_cluster.sh`, `gpdemo-defaults.sh`, `gpdemo.sh`, `probe_config.sh`)
  - `lalshell`
  - `README`

PR #303, opened 2023-11-16
[DO NOT REVIEW] test if singlenode introduces latent bugs

Ray-Eldath:
this pr tests if singlenode introduces latent & intermittent bugs that surface themselves in #291 because that pr failed for a stunning consecutive 4-5 times.


Ray-Eldath:
failure in icw-orca-test is due to incomplete revert. such a shame.

PR #339, opened 2023-12-18
Dispatch by shared memory

Ray-Eldath:
This PR makes `serializedPlantree` and `serializedQueryDispatchDesc` dispatched by shared memory instead of interconnect, so that they can be sent only once to writer QE, and synced inbetween reader QEs and writer QE on a segment through DSM. It has been discussed in https://github.com/orgs/cloudberrydb/discussions/243.

## Implementation Outline

- this PR mainly use **polling** on reader QE to wait shared plan get's dispatched from writer QE. this is the same mechanism as shared snapshot synchronization. to circumvent this we may need something special (e.g., signal) because seems current synchronization mechanisms (`barrier`, SharedLatch) cannot satisfy our requirement.
- to properly reclaim DSM segments for a query, a `reference` count is calculated on writer QE. and **the last** reader QE who reads it will reclaim the DSM. this is the same as parallel (`GpInsertParallelDSMHash`).
- to isolate states in each user connections, a shmem HTAB is used. also same as parallel.

## Prerequisites

This feature is only enabled when: **1)** current query is _not_ an extended query (cursor, or `Bind` messages, etc.), and **2)** there exists a gang in which **all** QEs are writer QE (notably, writer QE and writer gang are two different concepts).

#### Prerequisite 1
prerequisite 1 is a hard limit due to the way extended query (equery for short) works. during equery, there's always a live writer gang in which everyone's a writer QE (gang W). first a command `set gp_write_shared_snapshot=true` will be dispatched to gang W to force a shared snapshot sync, then the actual gang will be created in which everyone's reader (gang R) and the actual query gets dispatched to it. Immediately you can find out that when the actual query get's dispatched, no writer QE receives the plan (because all writer QEs are in gang W), so there's no one be responsible for shared query plan synchronization.

#### Prerequisite 2

prerequisite 2 is a tradeoff. Consider the following query plan:

image

In this plan, seg0 in slice1 is a writer that should receives full query text. but in slice2, seg0 is a reader that should receive `slimQueryText` (a `slimQueryText` is a query text w/o query plan and ddesc), and seg1 and seg2 are writer QEs. this means when dispatching to gang2, seg0 should receives slim query text (because the full plan can be synced to it from seg0 in slice1 which is a writer), but seg1 and seg2 should receives full query text. this poses challenges because on QD side, the current interface of cdb dispatcher limits **all dispatches happens on a per-gang basis** (`cdbdisp_dispatchToGang`) and plan cannot be changed from seg to seg in a gang. for the same reason, the reference count of a DSM segment cannot be dispatched from QD directly because it may be different from QE to QE, even they're in the same gang (consider a plan that have singleton reader).

We can surely workaround this by bringing more thorough refactor to cdb/dispatcher interfaces. but I don't think that's worth it though surely this's debatable.

#### Updatable Views

It may seem that the following invariant holds for any given query:

> For any segment that a plan touches, there's always a writer QE exists on that segment.

This is indeed true for many common queries, but unfortunately not all of them. Below is a counterexample:

image

#### `InitPlan`

If there's a`InitPlan` _at the root_ of a plan, there could be **two set of writer gang** created and **two rounds of dispatching happens** for the same query:

image

this is why we limit "same root" during `reference` calculation (https://github.com/Ray-Eldath/cloudberrydb/blob/dispatch-by-shmem/organized-and-unlogged/src/backend/utils/time/sharedqueryplan.c#L119-L121). Note that a `InitPlan` doesn't necessarily have to be at the root. It could be deep down the plan tree as well:

image

## Possible Outcome

All in all, though many requirements need to be meet for this feature to take effects, it is still very much **turned on** in most common queries (see tests for example). This is good news. On the bad side, I doubt whether this PR can make any noticeable performance improvements at all. On qd, we cannot completely get rid of libpq connections for now, and query dispatch is already **pipelined** to hide interconnect cost anyway. In the long run, if we are to "decentralize" QE by reassigning tasks (such as creating reader QEs, keepalive, etc.) to writer QE, this feature is a very good pathfinder and also a mandatory requisite. But if that's not the case, I doubt whether this feature alone worths the risk.

### DIFF
--- a/src/backend/cdb/dispatcher/cdbdisp_query.c
+++ b/src/backend/cdb/dispatcher/cdbdisp_query.c
@@ -104,7 +104,8 @@ static int fillSliceVector(SliceTable *sliceTable,
 				int len);
 
 static char *buildGpQueryString(DispatchCommandQueryParms *pQueryParms,
-				   int *finalLen);
+								int *finalLen,
+								bool withPlanAndDdesc);
 
 static DispatchCommandQueryParms *cdbdisp_buildPlanQueryParms(struct QueryDesc *queryDesc, bool planRequiresTxn);
 static DispatchCommandQueryParms *cdbdisp_buildUtilityQueryParms(struct Node *stmt, int flags, List *oid_assignments);
@@ -314,7 +315,7 @@ CdbDispatchSetCommand(const char *strCommand, bool cancelOnError)
 
 	ds = cdbdisp_makeDispatcherState(false);
 
-	queryText = buildGpQueryString(pQueryParms, &queryTextLength);
+	queryText = buildGpQueryString(pQueryParms, &queryTextLength, true);
 
 	primaryGang = AllocateGang(ds, GANGTYPE_PRIMARY_WRITER, cdbcomponent_getCdbComponentsList());
 
@@ -483,7 +484,7 @@ cdbdisp_dispatchCommandInternal(DispatchCommandQueryParms *pQueryParms,
 	 */
 	ds->destroyIdleReaderGang = true;
 
-	queryText = buildGpQueryString(pQueryParms, &queryTextLength);
+	queryText = buildGpQueryString(pQueryParms, &queryTextLength, true);
 
 	/*
 	 * Allocate a primary QE for every available segDB in the system.
@@ -848,7 +849,8 @@ fillSliceVector(SliceTable *sliceTbl, int rootIdx,
  */
 static char *
 buildGpQueryString(DispatchCommandQueryParms *pQueryParms,
-				   int *finalLen)
+                   int *finalLen,
+                   bool withPlanAndDdesc)
 {
 	const char *command = pQueryParms->strCommand;
 	int			command_len;
@@ -863,6 +865,7 @@ buildGpQueryString(DispatchCommandQueryParms *pQueryParms,
 	Oid			outerUserId = GetOuterUserId();
 	Oid			currentUserId = GetUserId();
 	int32		numsegments = getgpsegmentCount();
+	uint8       withPlanAndDdescValue = (uint8) withPlanAndDdesc;
 	StringInfoData resgroupInfo;
 
 	int			tmp,
@@ -906,13 +909,13 @@ buildGpQueryString(DispatchCommandQueryParms *pQueryParms,
 		sizeof(currentUserId) +
 		sizeof(n32) * 2 /* currentStatementStartTimestamp */ +
 		sizeof(command_len) +
+		sizeof(withPlanAndDdescValue) +
 		sizeof(plantree_len) +
 		sizeof(sddesc_len) +
 		sizeof(dtxContextInfo_len) +
 		dtxContextInfo_len +
 		command_len +
-		plantree_len +
-		sddesc_len +
+		(withPlanAndDdesc ? plantree_len + sddesc_len : 0) +
 		sizeof(numsegments) +
 		sizeof(resgroupInfo.len) +
 		resgroupInfo.len;
@@ -961,6 +964,10 @@ buildGpQueryString(DispatchCommandQueryParms *pQueryParms,
 	memcpy(pos, &tmp, sizeof(command_len));
 	pos += sizeof(command_len);
 
+	tmp = withPlanAndDdescValue;
+	memcpy(pos, &tmp, sizeof(withPlanAndDdescValue));
+	pos += sizeof(withPlanAndDdescValue);
+
 	tmp = htonl(plantree_len);
 	memcpy(pos, &tmp, sizeof(plantree_len));
 	pos += sizeof(plantree_len);
@@ -984,16 +991,19 @@ buildGpQueryString(DispatchCommandQueryParms *pQueryParms,
 	pos[command_len - 1] = '\0';
 	pos += command_len;
 
-	if (plantree_len > 0)
+	if (withPlanAndDdesc)
 	{
-		memcpy(pos, plantree, plantree_len);
-		pos += plantree_len;
-	}
+		if (plantree_len > 0)
+		{
+			memcpy(pos, plantree, plantree_len);
+			pos += plantree_len;
+		}
 
-	if (sddesc_len > 0)
-	{
-		memcpy(pos, sddesc, sddesc_len);
-		pos += sddesc_len;
+		if (sddesc_len > 0)
+		{
+			memcpy(pos, sddesc, sddesc_len);
+			pos += sddesc_len;
+		}
 	}
 
 	tmp = htonl(numsegments);
@@ -1044,6 +1054,8 @@ cdbdisp_dispatchX(QueryDesc* queryDesc,
 	int			rootIdx;
 	char	   *queryText = NULL;
 	int			queryTextLength = 0;
+	char       *slimQueryText = NULL; /* slimQueryText doesn't have plantree and ddesc which is dispatched to SharedQueryPlan readers */
+	int         slimQueryTextLength = 0;
 	struct SliceTable *sliceTbl;
 	struct EState *estate;
 	CdbDispatcherState *ds;
@@ -1083,7 +1095,8 @@ cdbdisp_dispatchX(QueryDesc* queryDesc,
 	sliceTbl->ic_instance_id = ++gp_interconnect_id;
 
 	pQueryParms = cdbdisp_buildPlanQueryParms(queryDesc, planRequiresTxn);
-	queryText = buildGpQueryString(pQueryParms, &queryTextLength);
+	queryText = buildGpQueryString(pQueryParms, &queryTextLength, true);
+	slimQueryText = buildGpQueryString(pQueryParms, &slimQueryTextLength, false);
 
 	/*
 	 * Allocate result array with enough slots for QEs of primary gangs.
@@ -1116,6 +1129,7 @@ cdbdisp_dispatchX(QueryDesc* queryDesc,
 		Gang	   *primaryGang = NULL;
 		ExecSlice  *slice;
 		int			si = -1;
+		bool        useFullQueryText = false;
 
 		Assert(sliceVector != NULL);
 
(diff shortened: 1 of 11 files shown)


yjhjstz:
```
else if (Gp_role == GP_ROLE_EXECUTE)
	{
		if (Gp_is_writer)
		{
			addSharedSnapshot("Writer qExec", gp_session_id);
		}
		else
		{
			/*
			 * NOTE: This assumes that the Slot has already been
			 *       allocated by the writer.  Need to make sure we
			 *       always allocate the writer qExec first.
			 */			 			
			lookupSharedSnapshot("Reader qExec", "Writer qExec", gp_session_id);
		}
	}
```
can reuse `SharedSnapshot` logic and code ?

src/backend/cdb/dispatcher/cdbdisp_query.c
```diff
@@ -906,13 +909,13 @@ buildGpQueryString(DispatchCommandQueryParms *pQueryParms,
 		sizeof(currentUserId) +
 		sizeof(n32) * 2 /* currentStatementStartTimestamp */ +
 		sizeof(command_len) +
+		sizeof(withPlanAndDdescValue) +
```
yjhjstz, line 912: can `withPlanAndDdescValue ` be 16-bit positive integers ?

src/backend/utils/time/sharedqueryplan.c
```diff
@@ -0,0 +1,203 @@
...
+		ResourceOwner oldowner;
+		Size size = serializedPlantreelen + serializedQueryDispatchDesclen;
+
+		entry = (SharedQueryPlanEntry *) hash_search(SharedQueryPlanHash,
+		                                             &gp_session_id,
+		                                             HASH_ENTER,
+		                                             &found);
+		Assert(!found);
+		if (!SharedQueryPlanResOwner)
+			SharedQueryPlanResOwner = ResourceOwnerCreate(NULL, "SharedQueryPlanResOwner");
```
yjhjstz, line 146: entry add a new var to store `ResourceOwner` , that's better ?

Ray-Eldath: the parent of `SharedQueryPlanResOwner` so it has a unlimited lifetime. this is to make `PrintDSMLeakWarning` **not** printing any warnings at transaction completion because the DSM's lifetime is attached to `entry->reference` that we already have a reclaim logic for. 

shared snapshot have the same approach to create a unbounded `DumpResOwner` which I guess is to do the same thing.

src/backend/cdb/dispatcher/cdbdisp_query.c
```diff
@@ -1146,9 +1160,19 @@ cdbdisp_dispatchX(QueryDesc* queryDesc,
...
+			cdbdisp_setDispatchParamsQueryText(ds, queryText, queryTextLength);
+		else
+			cdbdisp_setDispatchParamsQueryText(ds, slimQueryText, slimQueryTextLength);
+
 		if (Test_print_direct_dispatch_info)
-			elog(INFO, "(slice %d) Dispatch command to %s", slice->sliceIndex,
-						segmentsToContentStr(slice->segments));
+			elog(INFO, "(slice %d) Dispatch %s command to %s", slice->sliceIndex,
+			     useFullQueryText ? "full" : "slim",
+			     segmentsToContentStr(slice->segments));
```
avamingli, line 1175: Why change this?

avamingli:
> On the bad side, I doubt whether this PR can make any noticeable performance improvements at all. On qd, we cannot completely get rid of libpq connections for now, and query dispatch is already **pipelined** to hide interconnect cost anyway. In the long run, if we are to "decentralize" QE by reassigning tasks (such as creating reader QEs, keepalive, etc.) to writer QE, this feature is a very good pathfinder and also a mandatory requisite. But if that's not the case, I doubt whether this feature alone worths the risk.

Agree,  we have a long way to go, but this PR is a good start!  Thanks for your work and rich description.

src/backend/cdb/dispatcher/cdbdisp_query.c
Ray-Eldath: to make the dispatch logic tested. we can always remove this and all test cases changes.

avamingli: Please recover that. Direct dispatch is another thing.

Ray-Eldath:
> ```
> else if (Gp_role == GP_ROLE_EXECUTE)
> 	{
> 		if (Gp_is_writer)
> 		{
> 			addSharedSnapshot("Writer qExec", gp_session_id);
> 		}
> 		else
> 		{
> 			/*
> 			 * NOTE: This assumes that the Slot has already been
> 			 *       allocated by the writer.  Need to make sure we
> 			 *       always allocate the writer qExec first.
> 			 */			 			
> 			lookupSharedSnapshot("Reader qExec", "Writer qExec", gp_session_id);
> 		}
> 	}
> ```
> 
> can reuse `SharedSnapshot` logic and code ?

the original implementation was to use `SharedSnapShot`, but I quickly gave it up. `SharedSnapshot` served for a totally different purpose and its lifetime is tied to a transaction not a query. when dtxcontext is `AUTO_COMMIT_IMPLICIT`, txn lifetime _is_ query lifetime, but when you explicitly use txn with `BEGIN` stmt etc., they're different. we can of course make this work but I didn't go down that path.

also I can never figure out why they use a **array** to store slots discriminated by `gp_session_id`. use `ShmemSharedHTAB` as this PR and parallel can lead to a far simpler implementation. Do you have a theory on this? @yjhjstz

yjhjstz:
> the original implementation was to use `SharedSnapShot`, but I quickly gave it up. `SharedSnapshot` served for a totally different purpose and its lifetime is tied to a transaction not a query. when dtxcontext is `AUTO_COMMIT_IMPLICIT`, txn lifetime _is_ query lifetime, but when you explicitly use txn with `BEGIN` stmt etc., they're different. we can of course make this work but I didn't go down that path. I can try again if you think that's better.

I found they have the same logical except lifetime.

src/include/utils/sharedqueryplan.h
```diff
@@ -0,0 +1,29 @@
...
+	int        commandCount;
+} SharedQueryPlanEntry;
+
+extern void InitSharedQueryPlanHash(void);
+extern Size SharedQueryPlanHashSize(void);
+
+void ReadSharedQueryPlan(char *serializedPlantree, int serializedPlantreelen, char *serializedQueryDispatchDesc, int serializedQueryDispatchDesclen);
+void WriteSharedQueryPlan(const char *serializedPlantree, int serializedPlantreelen, const char *serializedQueryDispatchDesc, int serializedQueryDispatchDesclen, const SliceTable *sliceTable);
+
+extern void AtAbort_SharedQueryPlan(void);
```
yjhjstz, line 29: add an emtpy line.

src/backend/utils/time/sharedqueryplan.c
```diff
@@ -0,0 +1,203 @@
...
+			if (seg == NULL)
+				ereport(ERROR, errcode(ERRCODE_GP_INTERCONNECTION_ERROR),
+				        errmsg("Dispatch by shmem commandCount matched but dsm_attach returns NULL"),
+				        errdetail("May due to out of memory. Please retry. "
+				                  "commandCount = %d (local %d), handle = %u",
+				                  entry->commandCount, gp_command_count, entry->handle));
+
+			addr = dsm_segment_address(seg);
+			memcpy(serializedPlantree, addr, serializedPlantreelen);
+			memcpy(serializedQueryDispatchDesc, addr + serializedPlantreelen, serializedQueryDispatchDesclen);
```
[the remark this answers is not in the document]
Ray-Eldath: we can, but then we need to take `entry->reference--` out of `ReadSharedQueryPlan` and make it a separate function to be called _after_ the `serializedPlantree` and `serializedQueryDispatchDesc` be deserialized in `exec_mpp_query`. do you think this's ok?

src/backend/cdb/dispatcher/cdbdisp_query.c
Ray-Eldath: I further change this to a `char`.

src/backend/utils/time/sharedqueryplan.c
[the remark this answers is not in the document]
Ray-Eldath: I think we couldn't make it. `serializedPlantree` is a `const char *` returned from `pq_getmsgbytes`.

src/backend/utils/time/sharedqueryplan.c
Ray-Eldath: It requires another round of `hash_search` to find the entry and decrease `referenced` _after_ the deserialization of `serializedPlantree` and `serializedQueryDesc`. I think maybe two `memcpy` is still more efficient.

src/backend/utils/time/sharedqueryplan.c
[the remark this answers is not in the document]
Ray-Eldath: It requires another round of `hash_search` to find the entry and decrease `referenced` _after_ the deserialization of `serializedPlantree` and `serializedQueryDesc`. I think maybe two `memcpy` is still more efficient.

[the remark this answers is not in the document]
yjhjstz: yes, I worry about performance degradation.

src/backend/utils/time/sharedqueryplan.c
```diff
@@ -0,0 +1,203 @@
...
+			dsm_detach(seg);
+
+			entry->reference--;
```
fanfuxiaoran, line 62: A exclusive lock on entry to do "entry->reference--"?

src/backend/utils/time/sharedqueryplan.c
```diff
@@ -0,0 +1,203 @@
...
+}
+
+void
+WriteSharedQueryPlan(const char *serializedPlantree, int serializedPlantreelen, const char *serializedQueryDispatchDesc, int serializedQueryDispatchDesclen, const SliceTable *sliceTable)
+{
+	int i;
+	ExecSlice *slice = NULL;
+	ExecSlice *local = NULL;
+
+	LWLockAcquire(SharedQueryPlanLock, LW_EXCLUSIVE);
```
fanfuxiaoran, line 111: `SharedQueryPlanLock` is a global lock?

src/backend/utils/time/sharedqueryplan.c
```diff
@@ -0,0 +1,203 @@
...
+	                                             NULL);
+	if (entry)
+	{
+		destroy_entry(entry, false);
+	}
+	LWLockRelease(SharedQueryPlanLock);
+}
+
+void
+AtAbort_SharedQueryPlan(void)
```
fanfuxiaoran, line 200: Only removing the plan when aborting transaction? How about commit ?

my-ship-it:
Close the PR since no response for a long time.

PR #352, opened 2023-12-29
Add guc gp_enable_explain_motion_detail to show sender time when Expl…

songdongxiaoa2:
Show motion sender time when Explain Analyse.

When guc set, Explain Analyse result will be:
actual time=<receive_start>..<receive_end>/<sender_start>..<sender_end>

gp_enable_explain_motion_detail defaults to off.

---

### Change logs

QEs will now send Motion Sender instruments to QD.
QD will collect two instruments(sender and receiver) to MotionPlanState->instruments[].   
Add guc gp_enable_explain_motion_detail to show sender time.
Add test cases.

In plan on QD, Motion is a single node, so we need an array of two to collect both sender and receiver instruments.
Add guc to show this information only when needed.

Yes, the guc gp_enable_explain_motion_detail default to off. Show sending time by:
set gp_enable_explain_motion_detail to on
in client.

Added icw test case and passed in local.


my-ship-it:
Close the PR since no response for a long time.

PR #201, opened 2023-09-14
Fix unportable scripts on macOS

Ray-Eldath:
### Change logs

Fix unportable scripts on macOS

1. Regarding `greenplum_path.sh`: In some platforms (notably macOS), various Python installations may co-exist. When `demo_cluster.sh` creates segments using ssh, it may not run by the same Python installation as the coordinator because env are not passed through, causing the super weird `"Module 'pgdb' not found"` error. This PR fix this by probing the path of the Python installation and write it down in `greenplum_path.sh`.
2. `README.macOS.bash` and `gp_bash_functions.sh` are changed as well to properly create cluster on macOS.

### Why are the changes needed?

> [!IMPORTANT]
> **This PR is a by-product of #198. These two PR are necessary to support CBDB building and running on macOS platform.**

### Does this PR introduce any user-facing change?

Before this PR, user cannot run `make create-demo-cluster` on macOS. Now they can.

PR #363, opened 2024-01-11
Fix greenplum_path.sh change PATH

Ray-Eldath:
Now only override PATH when Python used during ./configure isn't included in.

### DIFF
--- a/gpMgmt/bin/generate-greenplum-path.sh
+++ b/gpMgmt/bin/generate-greenplum-path.sh
@@ -27,13 +27,20 @@ else
 fi
 EOF
 
+# only override PATH when Python used during ./configure isn't included in.
+if [[ $(python3 -c "import sys; print(sys.executable)") == "${WHICHPYTHON}" ]]; then
+  PATHSTR='${GPHOME}/bin:${PATH}'
+else
+  PATHSTR='${GPHOME}/bin:${PYTHONBINDIR}:${PATH}'
+fi
+
 cat <<EOF
 PYTHONBINDIR="$(dirname "${WHICHPYTHON}")"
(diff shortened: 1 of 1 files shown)


gpMgmt/bin/generate-greenplum-path.sh
```diff
@@ -27,13 +27,20 @@ else
...
+fi
+
 cat <<EOF
 PYTHONBINDIR="$(dirname "${WHICHPYTHON}")"
+PATH="${PATHSTR}"
 EOF
 
 cat <<"EOF"
 PYTHONPATH="${GPHOME}/lib/python"
-PATH="${GPHOME}/bin:${PYTHONBINDIR}:${PATH}"
```
yjhjstz, line 36: can put $PATH in front ? like `PATH="${PATH}:${GPHOME}/bin:${PYTHONBINDIR}"`

gfphoenix78: Probably no. The system `psql` like commands(comes from the standard postgres) will have higher precedence over the commands in cbdb.

gfphoenix78:
@Ray-Eldath Could you describe what's wrong in the current greenplum_path.sh?

Ray-Eldath:
Current greenplum_path.sh is a result of #201 to make greenplum_path.sh always using the Python *during `./configure`*.

But it introduced unnecessary changes to user's `$PATH`. Colleagues reported that they use multiple gcc toolchain versions, and `source greenplum.sh` accidently changes gcc version that they're using greenplum_path.sh because #201 changes `$PATH`.

I think we could just revert this portion of #201 for the sake of simplicity.

PR #386, opened 2024-03-01
Revert #201 partially

Ray-Eldath:
Current greenplum_path.sh is a result of #201 to make greenplum_path.sh always using the Python during `./configure`. But it inadvertently changes user's `PATH`. This portion of #201 is reverted, and it should be fine if people all use `pip install --user` to install dependencies.


my-ship-it (approved):
LGTM

Ray-Eldath:
@my-ship-it what about merge this? :)

PR #391, opened 2024-03-06
Add new callback 'scan_flags' for table access method

HuSen8891:
The table access method supports different features due to the different implementation, this callback is used to indicate what the AM can do, what features the AM can support.

### DIFF
--- a/src/include/access/tableam.h
+++ b/src/include/access/tableam.h
@@ -390,6 +390,13 @@ typedef struct TableAmRoutine
 											  ScanDirection direction,
 											  TupleTableSlot *slot);
 
+	/*
+	 * This callback is used to indicate what the AM can do, what features the
+	 * AM can support, return the flags represented the supported features of
+	 * scan.
+	 */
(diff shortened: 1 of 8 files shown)


src/include/access/tableam.h
```diff
@@ -390,6 +390,13 @@ typedef struct TableAmRoutine
 											  ScanDirection direction,
 											  TupleTableSlot *slot);
 
+	/*
+	 * This callback is used to indicate what the AM can do, what features the
+	 * AM can support, return the flags represented the supported features of
+	 * scan.
+	 */
+	int			(*scan_flags) (Relation rel);
```
my-ship-it, line 398: Should we init scan_flags for HeapAm/AOAm/AOCOAm ?

my-ship-it (approved):
LGTM

src/include/access/tableam.h
HuSen8891: yes, let me add the init methods.

avamingli (approved):

PR #406, opened 2024-04-10
Add GUC 'gp_random_insert_segments' to control the segments used for random distributed table insertion

foreyes:
Introduces the 'gp_random_insert_segments' GUC to reduce the generation of excessive fragmented files during the insertion of small amounts of data into clusters with a large number of segments (e.g., 1000 records into 100 segments).

Fragmented data insertion can significantly degrade performance, especially when using append-optimized or cloud-based storage. By introducing the 'gp_random_insert_segments' GUC, users can limit the number of segments used for data insertion in randomly distributed tables, which can significantly reduce fragmented files.

### DIFF
--- a/src/backend/utils/misc/guc_gp.c
+++ b/src/backend/utils/misc/guc_gp.c
@@ -146,6 +146,7 @@ int			gp_appendonly_compaction_threshold = 0;
 bool		enable_parallel = false;
 int			gp_appendonly_insert_files = 0;
 int			gp_appendonly_insert_files_tuples_range = 0;
+int			gp_random_insert_segments = 0;
 bool		gp_heap_require_relhasoids_match = true;
 bool		gp_local_distributed_cache_stats = false;
 bool		debug_xlog_record_read = false;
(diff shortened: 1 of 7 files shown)


src/backend/utils/misc/guc_gp.c
```diff
@@ -3215,6 +3216,16 @@ struct config_int ConfigureNamesInt_gp[] =
 		NULL, NULL, NULL
 	},
 
+	{
+		{"gp_random_insert_segments", PGC_USERSET, CUSTOM_OPTIONS,
```
my-ship-it, line 3220: Also update in src/include/utils/sync_guc_name.h?

src/backend/utils/misc/guc_gp.c
```diff
@@ -3215,6 +3216,16 @@ struct config_int ConfigureNamesInt_gp[] =
...
+			gettext_noop("Use limited number of segments for random distributed table insertion."),
+			NULL
+		},
+		&gp_random_insert_segments,
+		0, 0, INT_MAX,
```
gfphoenix78, line 3225: What happens if the `gp_random_insert_segments` is 0 or greater than the number of segments in the cluster?

foreyes: 0 means that the feature is disabled, greater than actual segments number will have any effect.

src/backend/utils/misc/guc_gp.c
foreyes: Already done.

my-ship-it:
Please add test case?

my-ship-it (approved):
LGTM

PR #501, opened 2024-07-07
Maintain materialized view data status.

avamingli:
A materialized view's data could be seen as up to date with base tables if there is no writable operations since last Refresh of the view.

If one of base tables in the query tree of a materailied view is modified, the view data is not up to data. And we could not use it to answer query.

This commit maintain the data satus of a materialized view and it applies to normal materialized view, IVM with defer refresh. IVM with immediately refresh is always up to date.

When a base table has writable operation, we try to update view data status as:

- 'u'(up to date):
    - Create Materialized View
    - Refresh

- 'e'(expired): 
  - Update base tables
  - Delete base tables 
  - Refresh With No Data 
  - Truncate base tables 
  - Create Materialized View With No Data

- 'i'(insert only): 
  - Insert info base tables 
  - Copy From
  - Copy From on Segments

- 'r'(up to date but reorganized): 
  - Cluster base tables
  -  Vacuum Full base tables

Insert, Update, Delete, Copy From operations take effect if the actual affected rows > 0, ex:
 ```sql insert into t1 select * from t2;```
We don't need to update view if t2 has zero rows.

The real status will be decided both by current and the status we try to mark as.Ex: we try to mark insert only on a expired status, it will not success.

This doesn't work on utility mode, if user modified data in that mode, should refresh views if want to use it to answer query.

AQUMV could use normal materialized views after this commit.

fix #ISSUE_Number

---

### Change logs

### Contributor's Checklist

- [ ] Document changes.


src/backend/catalog/gp_matview_aux.c
```diff
@@ -0,0 +1,444 @@
...
+
+	RangeTblEntry	*rte = rt_fetch(1, viewQuery->rtable);
+	if (rte->rtekind != RTE_RELATION)
+		return NIL;
+
+	/* Only support normal relation now. */
+	if (get_rel_relkind(rte->relid) != RELKIND_RELATION)
+		return NIL;
+
+	relids = list_make1_oid(rte->relid);
```
my-ship-it, line 94: We can only process one Range Table Entry?

src/backend/executor/execMain.c
```diff
@@ -972,6 +973,61 @@ standard_ExecutorRun(QueryDesc *queryDesc,
...
+		 * to avoid case writablte query may not
+		 * change data when success, ex:
+		 *  insert into t1 select * from t2;
+		 * When t2 has zero rows, don't need to update view status.
+		 *
+		 * NB: This can't handle well in utility mode, should REFRESH by user
+		 * after that.
+		 */
+		if ((GP_ROLE_DISPATCH == Gp_role && es_processed > 0) ||
+			(IS_SINGLENODE() && (operation != CMD_SELECT) && estate->es_processed > 0))
```
my-ship-it, line 995: We also need to catch for  'BEFORE ROW INSERT or INSTEAD OF ROW INSERT' triggers?

src/backend/catalog/gp_matview_aux.c
avamingli: Yeah, for now. 
AQUMV and Append Agg plan only could use one base table now.
We could store more complex query if we are ready to use them, ex: a join.

src/backend/executor/execMain.c
avamingli: NSTEAD OF ROW INSERT in only for  updatable view which is not well supported in CBDB, and we only support one base table to maintain view datastaus.
So it could be bypassed in CBDB or SingleNode mode.

BEFORE ROW INSERT
 * for triggers on segments, CBDB could not exec a non-SELECT function: `function cannot execute on a QE slice because it issues a non-SELECT statement  `
 * for triggers try to modify entry tables, most of them are catalog which is not allowed in production.
 
In SingleNode mode, all tables include user tables are on QD.
Need to verify BEFORE ROW INSERT triggers and etc.

my-ship-it (approved):
LGTM except one comment

src/backend/executor/execMain.c
avamingli: Verify that it could work with triggers  in SingleNode mode:
tri_t1 has trigers insert into tri_t2 and update tri_t3.
An insertion into t1 will update tri_t2, tri_t3 relative materialized views data status well.

```sql
create table tri_t1(a int, b int);
create table tri_t2(a int, b int);
create table tri_t3(a int, b int);
create materialized view tri_mv1 as select * from tri_t1;
create materialized view tri_mv2 as select * from tri_t2;
insert into tri_t3 values (1, 2);
create materialized view tri_mv3 as select * from tri_t3;
create function trigger_insert_tri_t2()
returns trigger AS
$$
begin
  execute 'insert into tri_t2 values(1, 1)';
  execute 'update tri_t3 set b = 10 where a = 1;';
  return NEW;
end;
$$
language plpgsql;
create trigger trigger_insert_tri_t2 before insert ON tri_t1
  for each row execute procedure trigger_insert_tri_t2();
select mvname, datastatus from gp_matview_aux where mvname in ('tri_mv1', 'tri_mv2', 'tri_mv3');
 mvname  | datastatus 
---------+------------
 tri_mv1 | u
 tri_mv2 | u
 tri_mv3 | u
(3 rows)

select * from tri_t1;
 a | b 
---+---
(0 rows)

select * from tri_t2;
 a | b 
---+---
(0 rows)

select * from tri_t3;
 a | b 
---+---
 1 | 2
(1 row)

insert into tri_t1 values(1, 2);
select * from tri_t1;
 a | b 
---+---
 1 | 2
(1 row)

-- should also insert data
select * from tri_t2;
 a | b 
---+---
 1 | 1
(1 row)

-- shoud be updated
select * from tri_t3;
 a | b  
---+----
 1 | 10
(1 row)

-- check mv status
select mvname, datastatus from gp_matview_aux where mvname in ('tri_mv1', 'tri_mv2', 'tri_mv3');
 mvname  | datastatus 
---------+------------
 tri_mv2 | i
 tri_mv3 | e
 tri_mv1 | i
(3 rows)
```
i -> insert only
e -> expired

src/backend/catalog/gp_matview_aux.c
my-ship-it: Sounds good.

PR #520, opened 2024-07-15
to enhance gp_dqa test stability, avoiding issues arising from imprecision of floating-point arithmetic

congxuebin:
ICW test case gp_dqa failed intermittently, as following

```
 select corr(distinct d, i) from dqa_t1;
         corr        
 --------------------
- 0.0824013341460019
+ 0.0824013341460018
 (1 row) 
```

It is not good practice to verify the fload8 data directly due to precision issues. Thus I replace all instances of "corr(distinct d, i)" with "to_char(corr(distinct d, i), '0.99999999999999')". The function to_char() is used to format the output of the correlation calculation into a string with a predefined format mask. In this instance, the format mask '0.99999999999999' specifies that the output should be rounded to 15 decimal places. By doing so, we are essentially truncating the precision of the floating-point number to a manageable level that minimizes the impact of rounding errors.

This modification ensures that the comparison of floating-point results in our test cases is conducted in a manner that is both practical and reliable.

### DIFF
--- a/src/test/isolation2/expected/fts_segment_reset.out
+++ b/src/test/isolation2/expected/fts_segment_reset.out
@@ -10,6 +10,12 @@
 -- start_ignore
 alter system set gp_fts_probe_interval to 10;
 ALTER
+-- Because after RESET, it still takes a little while for the primary
+-- to restart, and potentially makes FTS think it's in "recovery not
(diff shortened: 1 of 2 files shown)


avamingli:
Use `extra_float_digits` to control.

PR #521, opened 2024-07-15
make gp_dqa test case stable

congxuebin:
ICW test case gp_dqa failed intermittently, as following example:

```
select corr(distinct d, i) from dqa_t1;
         corr        
 --------------------
- 0.0824013341460019
+ 0.0824013341460018
 (1 row) 
```

It is not a good practice to verify the fload8 data directly due to precision issues. 
Thus setting GUC extra_float_digits=-1 to truncate the precision of the floating-point number to a manageable level that minimizes the impact of rounding errors.  The result of corresponding example would be changed to as following： 

```
       corr        
-------------------
 0.082401334146002
```

This modification ensures that the comparison of floating-point results in our test cases is conducted in a manner that is both practical and reliable.

### DIFF
--- a/src/test/regress/expected/gp_dqa.out
+++ b/src/test/regress/expected/gp_dqa.out
@@ -3,7 +3,7 @@
 -- are flowing from different segments in different order. Mask those
 -- differences by setting 'extra_float_digits'. This isn't enough for all of
 -- the queries, so a few also use TO_CHAR() to truncate the results further.
-set extra_float_digits=0;
+set extra_float_digits=-1;
 SET optimizer_trace_fallback to on;
 drop table if exists dqa_t1;
 NOTICE:  table "dqa_t1" does not exist, skipping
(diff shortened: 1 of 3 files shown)


avamingli:
Why create another pr?

avamingli:
> @avamingli Mingli, thanks for the suggestion. According to it I changed to use extra_float_digits. The old PR was automatically closed when I undo the earlier commit. The modification is totally different from the one earlier. Would it be OK if I use this new PR?

Next time,  could use git reset to rollback your commit, git rebase to squash commits and push force on your branch  if ready.
We should always keep context and all story in only one PR.

congxuebin:
@avamingli Got it. Will do. Thanks Mingli.

avamingli (approved):

congxuebin:
Unfortunately I forgot to close this PR before  another commit.  I think it's required to go through the CI process again.

congxuebin:
Hi Mingli @avamingli,
All tests have passed,  can you please help merge the PR?

avamingli:
Pushed.

PR #563, opened 2024-08-08
Support distributed by clause specification when Create MV if not exists

reshke:
$subj

For some reason, current grammatical parser does not accept Distributed By clause in case IF NOT EXISTS specified for materialized view, which I don't any any reason to disallow

### DIFF
--- a/src/test/regress/sql/matview.sql
+++ b/src/test/regress/sql/matview.sql
@@ -187,6 +187,11 @@ SELECT * FROM mvtest_mv_v_3;
 SELECT * FROM mvtest_mv_v_4;
 DROP TABLE mvtest_v CASCADE;
 
+-- Check that CREATE IF NOT EXISTS accept DISTRIBUTED BY
+CREATE MATERIALIZED VIEW IF NOT EXISTS mv_ine_distr (a, b) AS
+  SELECT generate_series(1, 10) a, generate_series(1, 10) b  DISTRIBUTED BY (b);
+DROP MATERIALIZED VIEW mv_ine_distr;
+
 -- Check that unknown literals are converted to "text" in CREATE MATVIEW,
 -- so that we don't end up with unknown-type columns.
 CREATE MATERIALIZED VIEW mv_unspecified_types AS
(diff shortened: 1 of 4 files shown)


reshke:
hmmm icw-test are failing, but they look unrelated to C.M.V, aren't they?

avamingli:
Hi, please provide cases to verify what you have done.

reshke:
> Hi, please provide cases to verify what you have done.

Hi, you mean some regression test-cases? They are as obvious as 
```
create materialized view if not exists mv1 as select 1;
``` 
(this fails on current HEAD)

I dont think we need anything more complicated here.

avamingli:
> > Hi, please provide cases to verify what you have done.
> 
> 
> 
> Hi, you mean some regression test-cases? They are as obvious as 
> 
> ```
> 
> create table t1(i int);
> 
> create materialized view if not exists mv1 as select i from t1 distributed by (i);
> 
> ``` 
> 
> (this fails on current HEAD)
> 
> 
> 
> I dont think we need anything more complicated here.
> 
> 
> 
> Will do

The gramma you changed is create **incremental** materialized view, not a normal one, right?

reshke:
Seems that matview_optimizer.out depends on matriew tests. Pushed commit should fix test

src/test/regress/sql/matview.sql
```diff
@@ -187,6 +187,11 @@ SELECT * FROM mvtest_mv_v_3;
 SELECT * FROM mvtest_mv_v_4;
 DROP TABLE mvtest_v CASCADE;
 
+-- Check that CREATE IF NOT EXISTS accept DISTRIBUTED BY
+CREATE MATERIALIZED VIEW IF NOT EXISTS mv_ine_distr (a, b) AS
+  SELECT generate_series(1, 10) a, generate_series(1, 10) b  DISTRIBUTED BY (b);
+DROP MATERIALIZED VIEW mv_ine_distr;
```
avamingli, line 193: `\d+ mv_ine_distr ` to display dist key info.

reshke: even though the main job of these test in to check that DISTRIBUTED BY and IF NOT EXISTS are accepted together, not the DISTRIBUTED BY value?
Well, maybe this is good for sake of sameness test set as whole. 
Pushed as requested

src/test/regress/sql/matview.sql
reshke, line 1: Sorry for noise, forgot this file last time

reshke:
Are there any open items/suggestion/issues?

avamingli (approved):

PR #576, opened 2024-08-16
Rephrase comment for deferred IVM case

reshke:
Have been studying AQUMV feature internals today. Noticed the typo in comment:

```...defered-fefresh...```

I decided to rephrase it a little bit. This way it is more clear IMO.

### DIFF
--- a/src/backend/optimizer/plan/aqumv.c
+++ b/src/backend/optimizer/plan/aqumv.c
@@ -174,7 +174,8 @@ answer_query_using_materialized_views(PlannerInfo *root,
 		/*
 		 * AQUMV
 		 * Currently the data of IVM is always up-to-date if there were.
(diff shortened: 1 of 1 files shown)


reshke:
I omitted the fact that deferred IVM intend to only SERVERLESS mode. I think this should be state in doc, no need to duplicate this in comment. Am I right?

avamingli:
> I omitted the fact that deferred IVM intend to only SERVERLESS mode. I think this should be state in doc, no need to duplicate this in comment. Am I right?

Hi, 
Put it into doc is good.
We use same kernel codes in SERVERLESS mode, and CBDB could also have deferred IVM in the future, I keep it here to remind us to correct the codes for that.
Thanks for the typos.

reshke:
> I keep it here to remind us to correct the codes for that. Thanks for the typos.

Ok, reverted this, lets keep it there

avamingli (approved):

PR #581, opened 2024-08-20
Change temp table name used to IMMV to some less frequent sequence of…

reshke:
Currently, if user run ```CREATE TEMP TABLE t();``` in their session, IVM will fail. Change its name with generated one, which has a little probability to cause problems. 

Also, i added some debug logging and CBDB_IVM_FIXME comment that I feel needed.

### DIFF
--- a/src/backend/commands/matview.c
+++ b/src/backend/commands/matview.c
@@ -214,7 +214,7 @@ static char *get_null_condition_string(IvmOp op, const char *arg1, const char *a
 						  const char* count_col);
 static void apply_old_delta(const char *matviewname, const char *deltaname_old,
 				List *keys);
-static void apply_old_delta_with_count(const char *matviewname, const char *deltaname_old,
+static void apply_old_delta_with_count(const char *matviewname, Oid matviewRelid, const char *deltaname_old,
 				List *keys, StringInfo aggs_list, StringInfo aggs_set,
 				const char *count_colname);
 static void apply_new_delta(const char *matviewname, const char *deltaname_new,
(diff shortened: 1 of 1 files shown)


yjhjstz:
think about schema to  solve this ?

reshke:
> think about schema to solve this ?

Temporary tables exist in a special schema, so a schema name cannot be given when creating a temporary table.[1]

[1] https://www.postgresql.org/docs/current/sql-createtable.html


```
reshke=# create temp table ex.t();
ERROR:  cannot create temporary relation in non-temporary schema
LINE 1: create temp table ex.t();
                          ^
```

yjhjstz:
> > think about schema to solve this ?
> 
> Temporary tables exist in a special schema, so a schema name cannot be given when creating a temporary table.[1]
> 
> [1] https://www.postgresql.org/docs/current/sql-createtable.html
> 
> ```
> reshke=# create temp table ex.t();
> ERROR:  cannot create temporary relation in non-temporary schema
> LINE 1: create temp table ex.t();
>                           ^
> ```

I may found a way out:
refer to `refresh_by_match_merge`,  make_new_heap of  RELPERSISTENCE_TEMP tablespace, that will not confilct.

reshke:
> > > think about schema to solve this ?
> > 
> > 
> > Temporary tables exist in a special schema, so a schema name cannot be given when creating a temporary table.[1]
> > [1] https://www.postgresql.org/docs/current/sql-createtable.html
> > ```
> > reshke=# create temp table ex.t();
> > ERROR:  cannot create temporary relation in non-temporary schema
> > LINE 1: create temp table ex.t();
> >                           ^
> > ```
> 
> I may found a way out: refer to `refresh_by_match_merge`, make_new_heap of RELPERSISTENCE_TEMP tablespace, that will not confilct.

I'm not quite following you suggestion here... First of all, there are no RELPERSISTENCE_TEMP usage inside `refresh_by_match_merge`? Did you mean `ExecRefreshMatView` func? If so,  it uses RELPERSISTENCE_TEMP, yes. It assigns this value to relpersistence under if-concurrent condition. Then it passes this variable to `make_new_heap`. In this function we have following: 

```

	if (relpersistence == RELPERSISTENCE_TEMP)
		namespaceid = LookupCreationNamespace("pg_temp");
	else
		namespaceid = RelationGetNamespace(OldHeap);

```
https://github.com/cloudberrydb/cloudberrydb/blob/dccc518bea74302c3aebbf446a6fa47f8510c12c/src/backend/commands/cluster.c#L756-L759


So, this is the same as CREATE TEMP TABLE $name.

Do I miss something?

yjhjstz:
yes, in `ExecRefreshMatView`,  I think make_new_heap will not conflict.

yjhjstz (approved):
LGTM

PR #590, opened 2024-08-23
Fix pxf_fragment.c compilation failure

### LINKED ISSUE #583: [Bug] Compilation Failure in pxf_fdw Due to PG_VERSION_NUM Update and --enable-pxf Option
### Cloudberry Database version

HEAD of main

### What happened

While building Cloudberry Database with the **--enable-pxf** configure option, I encountered a compilation failure in the **pxf_fdw** module. The issue appears to be related to the recent update of **PG_VERSION_NUM** to **140004**, reflecting the upgrade to PostgreSQL 14.4, rather than the Cloudberry Database version itself.

The problem is caused by conditional preprocessor directives that depend on **PG_VERSION_NUM**, leading to different compilation paths. For example, in **pxf_fragment.c**, a necessary header file is no longer included due to the updated version number:

```
#if (PG_VERSION_NUM <= 140000)
#include "commands/copyfrom_internal.h"
#endif
```

**Compilation Error:**
The error occurs during the compilation of the **ProcessRequest** function within **pxf_fragment.c**:

```
gcc -Wall -Wmissing-prototypes -Wpointer-arith -Werror=vla -Wendif-labels -Wmissing-format-attribute -Wimplicit-fallthrough=3 -Wcast-function-type -Wformat-security -fno-strict-aliasing -fwrapv -fexcess-precision=standard -Wno-unused-but-set-variable -Werror=implicit-fallthrough=3 -Wno-format-truncation -Wno-stringop-truncation -O3 -fPIC  -DUSE_INTERNAL_FTS=1  -Werror=uninitialized -Werror=implicit-function-declaration -Werror -fPIC -I. -I. -I../../src/include   -D_GNU_SOURCE -I/usr/include/libxml2  -I/usr/local/cloudberry/include  -c -o pxf_fragment.o pxf_fragment.c
pxf_fragment.c: In function ‘ProcessRequest’:
pxf_fragment.c:547:15: error: ‘RAW_BUF_SIZE’ undeclared (first use in this function); did you mean ‘BUFSIZ’?
  char  buffer[RAW_BUF_SIZE];
               ^~~~~~~~~~~~
               BUFSIZ
pxf_fragment.c:547:15: note: each undeclared identifier is reported only once for each function it appears in
pxf_fragment.c:547:8: error: unused variable ‘buffer’ [-Werror=unused-variable]
  char  buffer[RAW_BUF_SIZE];
        ^~~~~~
cc1: all warnings being treated as errors
```

_No response_

### How to reproduce

**Steps to Reproduce:**

1. Change Cloudberry Database version to 1.5.4 in configure.ac
2. Run autoconf to regenerate the configure script
3. Run configure with the --enable-pxf option
4. Compile the code.

### Operating System

Rocky Linux 9.4 (this issue is not OS dependent)

### Anything else

**Expected Outcome:**
The code should compile without errors.

**Actual Outcome:**
The compilation fails with the errors mentioned above.

**Additional Information:**
The issue stems from the changes in PG_VERSION_NUM and its impact on conditional preprocessor logic. The recent update to PostgreSQL 14.4 has introduced new conditions that were not present in earlier versions, leading to this failure.

This is a critical issue that needs to be addressed so we can include the **--enable-pxf** feature for our next open source build.

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://cloudberrydb.org/community/coc).

Terry1504:
pxf_fragment.c has a incorrect (seems typo) conditional preprocessor directives that depend on PG_VERSION_NUM, which makes copyfrom_internal.h is not included in the end.

Fix #583

### DIFF
--- a/gpcontrib/pxf_fdw/pxf_fragment.c
+++ b/gpcontrib/pxf_fdw/pxf_fragment.c
@@ -5,7 +5,7 @@
 #include "cdb/cdbtm.h"
 #include "cdb/cdbvars.h"
 #include "commands/copy.h"
-#if (PG_VERSION_NUM <= 140000)
+#if (PG_VERSION_NUM >= 140000)
 #include "commands/copyfrom_internal.h"
 #endif
 #include "common/jsonapi.h"


gpcontrib/pxf_fdw/pxf_fragment.c
```diff
@@ -5,7 +5,7 @@
 #include "cdb/cdbtm.h"
 #include "cdb/cdbvars.h"
 #include "commands/copy.h"
-#if (PG_VERSION_NUM <= 140000)
+#if (PG_VERSION_NUM >= 140000)
```
my-ship-it, line 8: Should be <= 150000?

Terry1504: I'm not quite sure.  Current PG version in CloudberryDB is 14.4, PG_VERSION_NUM is 140004.

Maybe we see what's going on at that time we bring in PG15?

This change is just it fixed this issue and it kept same way as code in pxf_fdw.c

#if (PG_VERSION_NUM >= 140000)
#include "commands/copyfrom_internal.h"
#include "commands/copyto_internal.h"
#endif

edespino: I am very confused how PG_VERSION_NUM is assigned. In configure.ac (see diff below), I changed the Cloudberry version to 1.6.3 and change PG_PACKAGE_VESION to 15.4.

After running `configure`, notice the value (`150003`) assigned to `PG_VERSION_NUM`. It seems as if it takes the 15 from PG_PACKAGE_VESION and the trailing two digits `03` from the last digit of the Cloudberry Database version:

```
src/include/pg_config.h:944:#define PG_VERSION_NUM 150003
```

Here is the only change I made to configure.ac:
```
diff --git a/configure.ac b/configure.ac
index 71e5e3c..1fe2c5c 100644
--- a/configure.ac
+++ b/configure.ac
@@ -20,8 +20,8 @@ m4_pattern_forbid(^PGAC_)dnl to catch undefined macros
 dnl The PACKAGE_VERSION from upstream PostgreSQL is maintained in the
 dnl PG_PACKAGE_VERSION variable, when merging make sure to update this
 dnl variable with the merge conflict from the AC_INIT() statement.
-AC_INIT([Cloudberry Database], [1.0.0], [info@cloudberrydb.org], [], [https://www.cloudberrydb.org/])
-[PG_PACKAGE_VERSION=14.4]
+AC_INIT([Cloudberry Database], [1.6.3], [info@cloudberrydb.org], [], [https://www.cloudberrydb.org/])
+[PG_PACKAGE_VERSION=15.4]
 AC_SUBST(PG_PACKAGE_VERSION)

 dnl m4_if(m4_defn([m4_PACKAGE_VERSION]), [2.69], [], [m4_fatal([Autoconf version 2.69 is required.
 ```

I believe your fix does address the compilation issue, but is there something fundamentally wrong in how PG_VERSION_NUM is assigned a value?

Terry1504: For above strange behavior  you described how PG_VERSION_NUM is assigned after change , I agreed it's strange for me too,  but I have no idea.

reshke: first of all, we do not need parentheses here
second, `commands/copyfrom_internal.h` was committed as part of https://github.com/postgres/postgres/commit/c532d15dddff14b01fe9ef1d465013cb8ef186df and still is part of PostgreSQL 
So, condition should be 
`#if PG_VERSION_NUM >= 140000`
As for PG_VERSION_NUM its value should come from PostgreSQL core major and minor version. So, given Cloudberry is based on 14.4 PG_VERSION_NUM should be equal to 140004
I didnt take a look on Cloudburry version generation, but if it results in anything different from 140004 then we have a bug here.

reshke (approved):
Generally speaking, provided fix is correct and fixes issue it aims to.

gpcontrib/pxf_fdw/pxf_fragment.c
edespino: @reshke Thank you for confirmation. As it stands today, the last digit of the Cloudberry Database version is used as the last digit assigned to PG_VERSION_NUM.  We have a bug.

edespino (approved):
+1

my-ship-it (approved):

PR #645, opened 2024-09-26
Adding the ability to create a non-distributed table (can be used for postgres extensions)

visill:
### Change logs

Add: create table table_name DISTRIBUTED BY LOCAL
Creates a table that is queried only on the coordinator

Some postgres extensions (sr_plan or yezzey) needs to create local table. Before we do this by C-functions. Now we can do using SQL syntax
 
No

It has regress test

### Contributor's Checklist

- [ ] Document changes.


wenchaozhang-123:
Should add test for join with other distribution table.

src/backend/parser/gram.y
```diff
@@ -6046,6 +6046,15 @@ DistributedBy:   DISTRIBUTED BY  '(' distributed_by_list ')'
 				distributedBy->keyCols = NIL;
 				$$ = (Node *)distributedBy;
 			}
+			| DISTRIBUTED LOCAL
+			{
+				DistributedBy *distributedBy = makeNode(DistributedBy);
+				distributedBy->ptype = POLICYTYPE_LOCAL;
```
reshke, line 6052: we did choose POLICYTYPE_LOCAL Instead of POLICYTYPE_ENRTY, because `getPolicyForDistributedBy` explicitly forbits creating relation with POLICYTYPE_ENRTY policy. We are unaware of reason however.

reshke (approved):

reshke:
> Should add test for join with other distribution table.

Maybe, but main reason for having LOCAL relation is that they act exactly like pg_catalog relations (so, used only on coordinator node).

reshke:
> > Should add test for join with other distribution table.
> 
> Maybe, but main reason for having LOCAL relation is that they act exactly like pg_catalog relations (so, used only on coordinator node).

for example, save-restore plan (sr_plan) extension uses relation to store query plans. Designed originally for postgres, this ext does work well in Greenplum (or its forks), because it tries to perform distributed query each time we access `sr_plans` table. 

https://github.com/yezzey-gp/ygp/blob/YGP_6.27_STABLE/gpcontrib/sr_plan/sr_plan--1.1--1.2.sql#L5-L15

wenchaozhang-123:
Squash commit message?

gfphoenix78:
Hi, @visill , thank you for your interesting about cloudberry. Could you give more details about why we should change the current data model? Coordinator in cloudberry inherits from greenplum. It doesn't process data scan, and tries to move the workload to the segments, so the database  has the ability to scale to a large data processing system. Local table seems violates our design, we need more discussion about why it's necessary.

Cloudberry has many extensions that were backported from PostgreSQL. Generally, our principle is to adapt the extensions to our distributed database, not the other way around.

avamingli:
> Cloudberry has many extensions that were backported from PostgreSQL. Generally, our principle is to adapt the extensions to our distributed database, not the other way around.

Absolutely right.

This is wrong way, you need to make your extensions compatible with CBDB kernel codes.
Greenplum/Cloudberry store catalog on master, other tables should be on segments in MPP mode.
Could you explain more why need create table on master?

Why your extensions need to store data on master, if it must be , it's a catalog.
If it needs to  work only on master, you need to change your extensions with Gp_role to do that.
A lot of extensions are modified to be compatible in MPP like that.

src/include/catalog/gp_distribution_policy.h
```diff
@@ -86,7 +86,8 @@ typedef enum GpPolicyType
 {
 	POLICYTYPE_PARTITIONED,		/* Tuples partitioned onto segment database. */
 	POLICYTYPE_ENTRY,			/* Tuples stored on entry database. */
-	POLICYTYPE_REPLICATED		/* Tuples stored a copy on all segment database. */
+	POLICYTYPE_REPLICATED,  	/* Tuples stored a copy on all segment database. */
+	POLICYTYPE_LOCAL            /* Tuples stored on coordinator */
```
weinan003, line 90: For the new policy type, is it any different from `POLICYTYPE_ENTRY`?

src/backend/parser/gram.y
my-ship-it: Possible we keep distributeBy as NULL like pg_class? So that planner could work properly

reshke:
> Could you explain more why need create table on master?

Because we need to store and access data on master, during planning phase. It is about `sr_plan` extension. 

> Why your extensions need to store data on master, if it must be , it's a catalog. If it needs to work only on master, you need to change your extensions with Gp_role to do that. A lot of extensions are modified to be compatible in MPP like that.

While this is generally true, it is very unhandy to do that. Today I bumped into problems with another extension - `pg_hint_plan`. This extension also uses table to store its planner 'hints'. This table should persist on master exclusively.  

So, the ability to force Greenplum to create relation without any distribution policy would help with at least two extensions now.

And yes, we can play with `Gp_role` inside extension installation script and all functions/sql/other that access this relation. Whilst it is achievable with `sr_plan`,  it is too much work in other cases, especially for https://github.com/ossc-db/pg_hint_plan

We also can execute `heap_create_with_catalog` from extension install script, to create all needed relation as we want (with NULL gp_policy). 
But specifying one proper option in `CREATE TABLE` sql is just much more simple.

reshke:
If one dislikes SQL approach, we can do it another way: using GUC to control transformDistributedBy behaviour when no distrib clause specified

src/backend/parser/gram.y
reshke: Will think about that

avamingli:
> > > We also can execute heap_create_with_catalog from extension install script, to create all needed relation as we want (with NULL gp_policy
> 
> > 
> 
> > That is forbidden in GPDB, all tables should have a distribution policy , you are passing an invalid param with that function.
> 
> 
> 
> It is impossible to forbid something for extension. extension install script is executed under superuser

I'm not talking about the permissions.
Superuser doesn't mean volatile our MPP design is right.
It's terrible that tables are created without a distribution policy entry in MPP by manual to avoid necessary checks like distribution policy, but not through SQL.

PR #653, opened 2024-10-01
Parallel DEDUP_SEMI and DEDUP_SEMI_REVERSE Join.(A new approach to process Semi Join Query in Parallel of MPP)

avamingli:
To handle semi join in parallel, we have enabled Parallel Semi join. 
This commit introduces a new approach to process Semi Join Query in Parallel of MPP, enable DEDUP_SEMI and DEDUP_SEMI_REVERSE join in parallel to handle semi join cases which could be a win in MPP mode.

Allow to use RowIdExpr in paths to process SEMI join query in parallel, and we could use Broadcast Motion in parallel too.
If one of that tables is relatively smaller than another, that will be a big win.

Even Redistribute both side(in parallel-oblivious plan) and unique it in upper plans, it may be a win compared to a SemiJoin Nodes.

---

**Performance**

image

Test of case[0] three times with 10 parallel number.

| plan | avg(ms) | 1st(ms) | 2nd(ms) | 3rd(ms) |
|--- | :---: | :---: | :---: | :---: | 
|Non-Parallell   | 32660 | 32655.304 | 32144.482 | 33181.489 |
|Parallell Semi Join  | 24670 | 24846.083 | 24134.965 | 25029.130 |
|Parallell DEDUP_SEMI Join  | 9249 | 7936.484 | 10333.245 | 9478.557 |
|Parallell DEDUP_SEMI_REVERSE Join  | 5448 | 6277.671 | 5950.229 | 4117.556 |

---
**Plan**

Both parallel-oblivious and parallel-aware are enabled, for examples of parallel-aware case:

**DEDUP_SEMI**

```sql
select sum(foo.a) from foo where exists (select 1 from bar where foo.a = bar.b);
                                       QUERY PLAN                                        
-----------------------------------------------------------------------------------------
 Finalize Aggregate
   ->  Gather Motion 6:1  (slice1; segments: 6)
         ->  Partial Aggregate
               ->  HashAggregate
                     Group Key: (RowIdExpr)
                     ->  Redistribute Motion 6:6  (slice2; segments: 6)
                           Hash Key: (RowIdExpr)
                           Hash Module: 3
                           ->  Parallel Hash Join
                                 Hash Cond: (foo.a = bar.b)
                                 ->  Broadcast Workers Motion 6:6  (slice3; segments: 6)
                                       ->  Parallel Seq Scan on foo
                                 ->  Parallel Hash
                                       ->  Parallel Seq Scan on bar
(15 rows)
```

**DEDUP_SEMI_REVERSE**

```sql
select sum(foo.a) from foo where exists (select 1 from bar where foo.a = bar.b);
                                           QUERY PLAN                                           
------------------------------------------------------------------------------------------------
 Finalize Aggregate
   ->  Gather Motion 6:1  (slice1; segments: 6)
         ->  Partial Aggregate
               ->  HashAggregate
                     Group Key: (RowIdExpr)
                     ->  Redistribute Motion 6:6  (slice2; segments: 6)
                           Hash Key: (RowIdExpr)
                           Hash Module: 3
                           ->  Parallel Hash Join
                                 Hash Cond: (bar.b = foo.a)
                                 ->  Parallel Seq Scan on bar
                                 ->  Parallel Hash
                                       ->  Broadcast Workers Motion 6:6  (slice3; segments: 6)
                                             ->  Parallel Seq Scan on foo
(15 rows)
```

---
For DEDUP_SEMI or DEDUP_SEMI_REVERSE join, each process need a unique RowIdExpr to identify unique rows, which is assigned with a baseline when building paths.

It's ok for non-parallel plan, but in parallel mode there are multiple processes on same segment, RowIdExpr in not unique then. To enable that, add ParallelWorkerNumberOfSlice to identify worker id of a parallel plan of a slice.

When rowidexpr is used, it's executed by 48 bits and left other 16 bits for segment_id. In parallel mode, we have to make more room for parallel worker id within segment_id's bits. This is done during planner with checks, in case that there are many segments with many parallel workers (which rarely happens).

---

**[0]** DDL & DML
```sql
create table foo(a int)  with(parallel_workers=10)  distributed randomly;
create table bar(b int)  with(parallel_workers=10)  distributed randomly;
insert into foo select i from generate_series(1,10000000)i;
insert into bar select i from generate_series(1,100000000)i;
analyze foo;
analyze bar;
select sum(foo.a) from foo where exists (select 1 from bar where foo.a = bar.b);
```

fix #ISSUE_Number

---

### Change logs

### Contributor's Checklist

- [ ] Document changes.


avamingli:
Another flaky failed cases of resgroup:
https://github.com/cloudberrydb/cloudberrydb/issues/651#issuecomment-2387565620

avamingli:
Nestloop seems not use Parallel, need a dig.

avamingli:
> Nestloop tends to use normal plan instead of Parallel, need a dig.

ENV problem, corrected.
And I have to clarify that even without Broadcast, Redistribute both side(in parallel-oblivious plan) and unique it in upper plans, it may be a win compared to a SemiJoin Node.
See parallel Nestloop test cases.

src/backend/executor/execExpr.c
```diff
@@ -1139,7 +1140,23 @@ ExecInitExprRec(Expr *node, ExprState *state,
...
+				/*
+				 * CBDB_PARALLEL
+				 * Planner have ensured that there is enough space for num of segments and parallel workers.
+				 * As we has not set ParallelWokerNumber yet now, use TotalParallelWorkerNumberOfSlice here
+				 * and keep bits space for ParallelWokerNumber.
+				 */
+				if (TotalParallelWorkerNumberOfSlice > 0)
+				{
+					int parallel_bits = pg_leftmost_one_pos32(TotalParallelWorkerNumberOfSlice) + 1;
+					int segs = getgpsegmentCount();
```
my-ship-it, line 1153: Both use getgpsegmentCount or dbid?

avamingli: Good idea!
Instead of total segments, use dbid for each query, we could make more room for parallel worker numbers.
If a query on a database with a small dbid, there could be more parallel workers for it~

avamingli: Oh, wait, there is a thinko.
We should use getgpsegmentCount() to check room, because on each segments , dbid is different.
But we should make sure all segments of a plan have enough room for segbits, so getgpsegmentCount is used for that.

avamingli: In fact, segs here  is a double check as we have already checked it during planner.
This check could be removed.

avamingli: AFAK, use a Max(DBID) is the right one in case that some segments down and recovered or CBDB drop some segments and then expanded.
@my-ship-it

avamingli: > AFAK, use a Max(DBID) is the right one in case that some segments down and recovered or CBDB drop some segments and then expanded.

 Codes updated as that.

my-ship-it (approved):

PR #670, opened 2024-10-15
Cherry pick Rows out in EXPLAIN ANALYZE

robozmey:
Cherry pick https://github.com/yezzey-gp/ygp/commit/1d4123048fbf5482c636eb193876d517edaaae81

### Change logs

Add "Rows out" print in cdbexplain_showExecStats

```
create table tt (a int, b int) with(parallel_workers=2) distributed by (a);
insert into tt select * from generate_series(1,1000)a,generate_series(1,1000)b;
set enable_parallel = on;
set max_parallel_workers_per_gather = 2;
set gp_enable_explain_rows_out = on;
```

```
explain (analyze) select * from tt where a > b;

                                                            QUERY PLAN                                                            
----------------------------------------------------------------------------------------------------------------------------------
 Gather Motion 6:1  (slice1; segments: 6)  (cost=0.00..548.21 rows=28700 width=8) (actual time=2.000..54.000 rows=499500 loops=1)
   Rows out: 499500.00 rows avg x 1 workers from 1 segments, 499500 rows max (seg-1 worker0), 499500 rows min (seg-1 worker0).
   ->  Parallel Seq Scan on tt  (cost=0.00..213.38 rows=4783 width=8) (actual time=0.000..18.000 rows=85668 loops=1)
         Filter: (a > b)
         Rows Removed by Filter: 82531
         Rows out: 83250.00 rows avg x 6 workers from 3 segments, 86550 rows max (seg2 worker4), 80436 rows min (seg1 worker3).
 Planning Time: 0.151 ms
   (slice0)    Executor memory: 31K bytes.
   (slice1)    Executor memory: 85K bytes avg x 6x(0) workers, 110K bytes max (seg0).
 Memory used:  128000kB
 Execution Time: 62.169 ms
(12 rows)
```

"Rows out" is useful for auto explain

This feature tests in regression `gp_explain` test

### Contributor's Checklist

- [ ] Document changes.

### DIFF
--- a/src/backend/cdb/cdbvars.c
+++ b/src/backend/cdb/cdbvars.c
@@ -278,6 +278,7 @@ int			gp_hashjoin_tuples_per_bucket = 5;
 int			gp_motion_slice_noop = 0;
 
 /* Apache Cloudberry Experimental Feature GUCs */
+bool		gp_enable_explain_rows_out = false;
 bool		gp_enable_explain_allstat = false;
 bool		gp_enable_motion_deadlock_sanity = false;	/* planning time sanity
 														 * check */
--- a/src/backend/commands/explain_gp.c
+++ b/src/backend/commands/explain_gp.c
@@ -1806,60 +1806,42 @@ cdbexplain_showExecStats(struct PlanState *planstate, ExplainState *es)
 	 */
 
 	if (gp_enable_explain_rows_out && es->analyze && ns->ninst > 0) {
-		double alltuples = 0;
-		double maxtuples = ns->insts[0].ntuples;
-		int maxseg = 0;
-		double mintuples = ns->insts[0].ntuples;
-		int minseg = 0;
+        double ntuples_max = ns->ntuples.vmax;
+        int ntuples_imax = ns->ntuples.imax;
+        double ntuples_min = ns->ntuples.vmax; /// TODO
+        int ntuples_imin = ns->ntuples.imax; /// TODO
+        double ntuples_avg = ns->ntuples.vsum / ns->ntuples.vcnt;
+        int ntuples_cnt = ns->ntuples.vcnt;
 
-		for (i = 0; i < ns->ninst; i++)
-		{
-			CdbExplain_StatInst *nsi = &ns->insts[i];
+        if (es->format == EXPLAIN_FORMAT_TEXT)
+        {
+            /*
+             * create a header for all stats: separate each individual stat by an
+             * underscore, separate the grouped stats for each node by a slash
+             */
+            appendStringInfoSpaces(es->str, es->indent * 2);
+            appendStringInfoString(es->str, "Rows out: ");
 
-			alltuples += nsi->ntuples;
-
-			if (nsi->ntuples > maxtuples) {
-				maxtuples = nsi->ntuples;
-				maxseg = ns->segindex0 + i;
-			}
-
-			if (nsi->ntuples < mintuples) {
-				mintuples = nsi->ntuples;
-				minseg = ns->segindex0 + i;
-			}
-		}
-
-		double avgtuples = alltuples / ns->ninst;
-
-		if (es->format == EXPLAIN_FORMAT_TEXT)
-		{
-			/*
-			 * create a header for all stats: separate each individual stat by an
-			 * underscore, separate the grouped stats for each node by a slash
-			 */
-			appendStringInfoSpaces(es->str, es->indent * 2);
-			appendStringInfoString(es->str, "Rows out: ");
-
-			appendStringInfo(es->str,
-								 "%0.2f rows avg x %d workers, %0.f rows max (seg%d), %0.f rows min (seg%d).\n",
-								 avgtuples,
-								 ns->ninst,
-								 maxtuples,
-								 maxseg,
-								 mintuples,
-								 minseg);
-		}
-		else {
-			// ExplainOpenGroup("Rows Out", NULL, false, es);
-			ExplainPropertyInteger("Workers", ns->ninst, es);
-			ExplainPropertyFloat("Average Rows", avgtuples, 1, es);
-			ExplainPropertyFloat("Max Rows", maxtuples, 0, es);
-			ExplainPropertyInteger("Max Rows Segment", maxseg, es);
-			ExplainPropertyFloat("Min Rows", mintuples, 0, es);
-			ExplainPropertyInteger("Min Rows Segment", minseg, es);
-			// ExplainCloseGroup("Rows out", NULL, false, es);
-		}
-	}
+            appendStringInfo(es->str,
+                                 "%.2f rows avg x %d workers, %.0f rows max (seg%d), %.0f rows min (seg%d).\n",
+                                 ntuples_avg,
+                                 ntuples_cnt,
+                                 ntuples_max,
+                                 ntuples_imax,
+                                 ntuples_min,
+                                 ntuples_imin);
+        }
+        else {
+            // ExplainOpenGroup("Rows Out", NULL, false, es);
+            ExplainPropertyInteger("Workers", NULL, ntuples_cnt, es);
+            ExplainPropertyFloat("Average Rows", NULL, ntuples_avg, 1, es);
+            ExplainPropertyFloat("Max Rows", NULL, ntuples_max, 0, es);
+            ExplainPropertyInteger("Max Rows Segment", NULL, ntuples_imax, es);
+            ExplainPropertyFloat("Min Rows", NULL, ntuples_min, 0, es);
+            ExplainPropertyInteger("Min Rows Segment", NULL, ntuples_imin, es);
+            // ExplainCloseGroup("Rows out", NULL, false, es);
+        }
+    }
 
 	/*
 	 * Dump stats for all workers.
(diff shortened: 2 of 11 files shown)


src/backend/utils/misc/guc_gp.c
```diff
@@ -772,6 +772,17 @@ struct config_bool ConfigureNamesBool_gp[] =
 		NULL, NULL, NULL
 	},
 
+	{
+		{"gp_enable_explain_rows_out", PGC_USERSET, CLIENT_CONN_OTHER,
+			gettext_noop("Experimental feature: print avg, min and max rows out in segments in EXPLAIN ANALYZE."),
```
reshke, line 777: it is not Experimental, is it feature from GP5

robozmey: Fixed

src/backend/cdb/cdbvars.c
```diff
@@ -277,6 +277,7 @@ int			gp_hashagg_groups_per_bucket = 5;
 int			gp_motion_slice_noop = 0;
 
 /* Cloudberry Database Experimental Feature GUCs */
+bool		gp_enable_explain_rows_out = false;
```
fanfuxiaoran, line 280: It is better  to send a param to explain command than using a guc to control if print out the "rows out"

src/include/cdb/cdbvars.h
```diff
@@ -626,6 +626,13 @@ extern bool gp_enable_agg_pushdown;
  */
 extern bool gp_enable_preunique;
 
+/* May Cloudberry print statistics as average, minimum and maximum rows out
+ * during EXPLAIN ANALYZE?
+ *
+ */
```
fanfuxiaoran, line 632: Cloudberry prints the maximum rows in current explain analyze result.

src/backend/commands/explain_gp.c
```diff
@@ -1703,6 +1703,66 @@ cdbexplain_showExecStats(struct PlanState *planstate, ExplainState *es)
...
+			ExplainPropertyFloat("Average Rows", avgtuples, 1, es);
+			ExplainPropertyFloat("Max Rows", maxtuples, 0, es);
+			ExplainPropertyInteger("Max Rows Segment", maxseg, es);
+			ExplainPropertyFloat("Min Rows", mintuples, 0, es);
+			ExplainPropertyInteger("Min Rows Segment", minseg, es);
+			// ExplainCloseGroup("Rows out", NULL, false, es);
+		}
+	}
+
 	/*
```
fanfuxiaoran, line 1766: Personally, I think it's better to compute the min/max/avg rows in 'cdbexplain_depositStatsToNode' function.
'CdbExplain_DepStatAcc' this struct currently maintains such info.

robozmey: Agree. Added min fields in CdbExplain_DepStatAcc and CdbExplain_Agg

src/backend/cdb/cdbvars.c
robozmey: Rows out feature created for auto_explain. If we add "rows out" as parameter of explain, then we'll need change auto_explain source, and auto_explain wont be compatible with vanilla

src/backend/commands/explain_gp.c
```diff
@@ -941,14 +941,20 @@ cdbexplain_collectStatsFromNode(PlanState *planstate, CdbExplain_SendStatCtx *ct
...
 	/* max's received inst in StatHdr */
 	CdbExplain_StatInst *rsimax;
 	/* max's inst in NodeSummary */
 	CdbExplain_StatInst *nsimax;
+	/* min's received StatHdr */
+	CdbExplain_StatHdr *rshmin;
+	/* min's received inst in StatHdr */
+	CdbExplain_StatInst *rsimin;
+	/* min's inst in NodeSummary */
+	CdbExplain_StatInst *nsimin;
```
fanfuxiaoran, line 957: The new added 3 fields `rshmin, rsimin,nsimin` seems not used.

robozmey: Removed

src/backend/cdb/cdbvars.c
fanfuxiaoran:
```
 auto_explain wont be compatible with vanilla
```
Hmm... cannot understand. Could you give more details? Is auto_explain aslo include in vanilla?

robozmey: auto_explain is included in vanilla https://github.com/greenplum-db/gpdb-archive/tree/main/contrib/auto_explain

Also GUC gp_enable_explain_rows_out is made by analogy with GUC gp_enable_explain_allstat

fanfuxiaoran (dismissed):

my-ship-it (dismissed):

src/backend/commands/explain_gp.c
```diff
@@ -1716,6 +1716,48 @@ cdbexplain_showExecStats(struct PlanState *planstate, ExplainState *es)
...
+                                 "%.2f rows avg x %d workers, %.0f rows max (seg%d), %.0f rows min (seg%d).\n",
+                                 ntuples_avg,
+                                 ntuples_cnt,
+                                 ntuples_max,
+                                 ntuples_imax,
+                                 ntuples_min,
+                                 ntuples_imin);
+        }
+        else {
+            // ExplainOpenGroup("Rows Out", NULL, false, es);
```
avamingli, line 1750: Please remove unused codes

avamingli (commented):
Do we have a **Parallel Plan** test(ex: cbdb_parallel.sql) to ensure the output is as expected?

src/backend/commands/explain_gp.c
robozmey: Removed

robozmey:
> Do we have a **Parallel Plan** test(ex: cbdb_parallel.sql) to ensure the output is as expected?

I'm not sure what we need **Parallel Plan** test, but I expanded `gp_explain.sql` test

avamingli:
> > Do we have a **Parallel Plan** test(ex: cbdb_parallel.sql) to ensure the output is as expected?
> 
> I'm not sure what we need **Parallel Plan** test, but I expanded `gp_explain.sql` test

In GPDB, there is one process executed the Slice on each segments. And explain use the statistic of that gang of QEs to compute the results.
In CBDB, we have Parallel feature. For a parallel plan, there may be multiple QEs execute the same Slice on each segments.
We call them parallel workers.

See CBDB style README https://github.com/apache/cloudberry/blob/11333c0b4d3a4962b0a6610ceb5b6d7a12e45ec4/src/backend/optimizer/README.cbdb.parallel for more details.
And most cases could be found in https://github.com/apache/cloudberry/blob/11333c0b4d3a4962b0a6610ceb5b6d7a12e45ec4/src/test/regress/sql/cbdb_parallel.sql

This is a significant difference between CBDB and GPDB.

In parallel plan cases, what do the statistic results should be? Do current codes compute results right or it's already expected?
Maybe the current codes don't need to do anything else more, but we should be clear about that.
In any cases, there should be test cases to verify the explain output.

fanfuxiaoran:
>> In parallel plan cases, what do the statistic results should be? Do current codes compute results right or it's already as expected?
Maybe the current codes don't need to do anything else more, but we should be clear about that.

Yes, agree! Looks we don't a lot things for the parallel workers. Cloudberry uses motion to gather the slice result both from workers on different segments and from the parallel worker on same segment . So the EXPLAIN ANALYZE command can still work properly to collect results from all workers.

But we should confirm the EXPLAIN ANALYZE output format for the plan contains parallel workers.  what should it be like?
```
 Rows out: 333333.33 rows avg x 3 workers, 1000000 rows max (seg1), 0 rows min (seg0).
```
Do we need to add the parallel worker info into this line? like 
```
 Rows out: 333333.33 rows avg x 3 workers from xx segments?, 1000000 rows max (seg1 worker?), 0 rows min (seg0 worker?).
```

avamingli:
@fanfuxiaoran Thanks for your analysis and verification.

>
> Do we need to add the parallel worker info into this line? like
> 
> ```
>  Rows out: 333333.33 rows avg x 3 workers from xx segments?, 1000000 rows max (seg1 worker?), 0 rows min (seg0 wo
> ```

Sounds good, with a worker info is more reasonable of parallel plan.

robozmey:
I added test for parallel plan (cbdb_plan.sql)

I added segment count, and worker ids for max/min

`Rows out: 83250.00 rows avg x 6 workers from 3 segments, 87051 rows max (seg2), 80349 rows min (seg1).`

robozmey:
> > > In parallel plan cases, what do the statistic results should be? Do current codes compute results right or it's already as expected?
> > > Maybe the current codes don't need to do anything else more, but we should be clear about that.
> 
> Yes, agree! Looks we don't a lot things for the parallel workers. Cloudberry uses motion to gather the slice result both from workers on different segments and from the parallel worker on same segment . So the EXPLAIN ANALYZE command can still work properly to collect results from all workers.
> 
> But we should confirm the EXPLAIN ANALYZE output format for the plan contains parallel workers. what should it be like?
> 
> ```
>  Rows out: 333333.33 rows avg x 3 workers, 1000000 rows max (seg1), 0 rows min (seg0).
> ```
> 
> Do we need to add the parallel worker info into this line? like
> 
> ```
>  Rows out: 333333.33 rows avg x 3 workers from xx segments?, 1000000 rows max (seg1 worker?), 0 rows min (seg0 worker?).
> ```

I added test for parallel plan (cbdb_plan.sql)

I added segment count, and worker ids for max/min

`Rows out: 83250.00 rows avg x 6 workers from 3 segments, 86144 rows max (seg2 worker4), 80391 rows min (seg1 worker2).`

leborchuk:
Need to rework it completely

PR #682, opened 2024-10-18
Fast path to REFRESH materialized view.

avamingli:
We already have the ability to track the data status for some materialized views, aware whether its data is up to date or not.
And we could avoid doing the real REFRESH if the data of view is up to date.

The no-refreshed data should be the logically same as after a real REFRESH when there is no data changed since latest REFRESH command.
In that case we may save a lot (read data from view query, compute and write into view table), ex: a cron task REFRESH view takes a long time and much resource periodically or executed manually by users each time.

New GUC: **gp_enable_refresh_fast_path**

Set this feature default to true, but let users decide if they intend to do a real REFRESH.

### Performance

If the fast path is chosen, we always return immediately and almost do nothing.
And the cost we save depends on the amount of data, the resource we use to compute and the time we read and write back to view table.

```sql
insert into t1 select i from generate_series(1, 100000000) i;
create materialized view mv2 as select * from t1 where a > 1 with no data;
refresh materialized view mv2;
REFRESH MATERIALIZED VIEW
Time: 194061.961 ms (03:14.062)

set gp_enable_refresh_fast_path = on;

refresh materialized view mv2;
REFRESH MATERIALIZED VIEW
Time: 4.617 ms
```

fix #ISSUE_Number

---

### Change logs

### Contributor's Checklist

- [ ] Document changes.

### DIFF
--- a/src/backend/commands/matview.c
+++ b/src/backend/commands/matview.c
@@ -388,6 +388,27 @@ ExecRefreshMatView(RefreshMatViewStmt *stmt, const char *queryString,
 	relowner = matviewRel->rd_rel->relowner;
 
 	/*
+	 * Fastpath to REFRESH a view:
+	 * avoid do the real REFRESH if the data of view
+	 * is up to date. The data should be the logically same as after
+	 * REFRESH when there is data changed since latest REFRESH.
+	 * In that case we may save a lot, ex: a cron task REFRESH view periodically
+	 * or manually executed by users each time.
+	 *
+	 * Set this feature default to true, but let uesrs decide if they intend
+	 * to do a real REFRESH.
+	 */
+	if (gp_enable_refresh_fast_path &&
+		!RelationIsIVM(matviewRel) &&
+		!stmt->skipData &&
+		MatviewIsGeneralyUpToDate(matviewOid))
+	{
+		ObjectAddressSet(address, RelationRelationId, matviewOid);
+		table_close(matviewRel, NoLock);
+		return address;
+	}
+
+	/*
 	 * Switch to the owner's userid, so that any functions are run as that
 	 * user.  Also lock down security-restricted operations and arrange to
 	 * make GUC variable changes local to this command.
--- a/src/backend/commands/matview.c
+++ b/src/backend/commands/matview.c
@@ -388,27 +388,6 @@ ExecRefreshMatView(RefreshMatViewStmt *stmt, const char *queryString,
 	relowner = matviewRel->rd_rel->relowner;
 
 	/*
-	 * Fastpath to REFRESH a view:
-	 * avoid do the real REFRESH if the data of view
-	 * is up to date. The data should be the logically same as after
-	 * REFRESH when there is data changed since latest REFRESH.
-	 * In that case we may save a lot, ex: a cron task REFRESH view periodically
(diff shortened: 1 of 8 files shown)


src/backend/commands/matview.c
```diff
@@ -387,6 +387,27 @@ ExecRefreshMatView(RefreshMatViewStmt *stmt, const char *queryString,
...
+	 * avoid do the real REFRESH if the data of view
+	 * is up to date. The data should be the logically same as after
+	 * REFRESH when there is data changed since latest REFRESH.
+	 * In that case we may save a lot, ex: a cron task REFRESH view periodically
+	 * or manually executed by users each time.
+	 *
+	 * Set this feature default to true, but let uesrs decide if they intend
+	 * to do a real REFRESH.
+	 */
+	if (gp_enable_refresh_fast_path &&
```
yjhjstz, line 401: those code should put after line440,  to do some check, like `CONCURRENTLY ` ?

avamingli: Good catch.

avamingli: Fixed.

avamingli: And fix by rollback User context, GUCs

src/backend/commands/matview.c
```diff
@@ -468,6 +468,34 @@ ExecRefreshMatView(RefreshMatViewStmt *stmt, const char *queryString,
...
```
yjhjstz, line 482: also !oldPopulated need to refresh.

avamingli: It's included in `MatviewIsGeneralyUpToDate()`

yjhjstz: OK ,better to add a test to cover this case.

avamingli: The feature is on by default, so there are many cases have tested no-populated mvs, example:
https://github.com/cloudberrydb/cloudberrydb/blob/74808061fa642568f714f32743f3b64fe6ccd3a2/src/test/regress/expected/incremental_matview.out#L15

```sql
CREATE INCREMENTAL MATERIALIZED VIEW mv_ivm_1 AS SELECT i,j,k FROM mv_base_a a INNER JOIN mv_base_b b USING(i) WITH NO DATA;
NOTICE:  Table doesn't have 'DISTRIBUTED BY' clause -- Using column(s) named 'i' as the Cloudberry Database data distribution key for this table.
HINT:  The 'DISTRIBUTED BY' clause determines the distribution of data. Make sure column(s) chosen are the optimal data distribution key to minimize skew.
SELECT * FROM mv_ivm_1 ORDER BY 1,2,3;
ERROR:  materialized view "mv_ivm_1" has not been populated
HINT:  Use the REFRESH MATERIALIZED VIEW command.
REFRESH MATERIALIZED VIEW mv_ivm_1;
SELECT * FROM mv_ivm_1 ORDER BY 1,2,3;
 i | j  |  k  
---+----+-----
 1 | 10 | 101
 2 | 20 | 102
 3 | 30 | 103
 4 | 40 | 104
(4 rows)
```

yjhjstz:
LGTM

yjhjstz (approved):

PR #705, opened 2024-11-07
[AQUMV] Answer Aggregation Query Directly.

avamingli:
This commits enable answer query which has aggregation directly. Use the results of view has aggregations to avoid compute those from origin table.
This may lead to significant efficiency gains if the SQL has a large amount of data. 
AQUMV will always return results immediately.

If we have a valid view like:
```sql
create materialized view mv as
  select sum(c1) as mc1, count(c2) as mc2, avg(c3) as mc3, count(*) as
mc4
  from t where c1 > 90;
```
SQL: 
```sql
select count(*), sum(c1), count(c2), avg(c3), abs(count(*) - 21) from t where c1 > 90;
```
Could be rewritten to:
```sql
select mc4, mc1, mc2, mc3, abs((mc4 - 21)) from mv;
```
Plan:
```sql
explain(verbose, costs off)
select count(*), sum(c1), count(c2), avg(c3), abs(count(*) - 21) from t where c1 > 90;
                                    QUERY PLAN
------------------------------------------------------------------------
   Output: mc4, mc1, mc2, mc3, (abs((mc4 - 21)))
   ->  Seq Scan on mv
         Output: mc4, mc1, mc2, mc3, abs((mc4 - 21))
 Settings: enable_answer_query_using_materialized_views = 'on',
optimizer = 'off'
(6 rows)
```
View query with Group By is not supported yet.

### HAVING clause process:

If some HAVING quals only exist in origin query and they could be computed from view query's target list, then we could keep them like post_quals.But as the view has aggregations, the additional quals should be moved to WHERE instead of HAVING.
```sql
create table t(c1 int, c2 int, c3 int, c4 int);
create materialized view mv as
  select sum(c1) as mc1, count(c2) as mc2, avg(c3) as mc3, count(*) as
mc4
  from t where c1 > 90;
```
SQL:
```sql
select count(*), sum(c1) from t where c1 > 90 having abs(count(*) - 21)
> 0 and 2 > 1 and avg(c3) > 97;
```
Could be rewritten to (The HAVING clause has been rewritten to WHERE clause):
```sql
select mc4, mc1 from mv where mc3 > 97 and abs(mc4 - 21) > 0;
```
Plan:
```sql
explain(verbose, costs off)
select count(*), sum(c1) from t where c1 > 90 having abs(count(*) - 21)
> 0 and 2 > 1 and avg(c3) > 97;
                                    QUERY PLAN
------------------------------------------------------------------------
   Output: mc4, mc1
   ->  Seq Scan on aqumv.mv
         Output: mc4, mc1
         Filter: ((mv.mc3 > '97'::numeric) AND (abs((mv.mc4 - 21)) > 0))
(7 rows)
```

There are two additional HAVING quals:
Expression: 2 > 1 (would be eliminated during planner). Expression: abs(count(*) - 21) > 0, it could be computed from view as:
```sql
	abs(mc4 - 21) > 0
```
And the new one is put to WHERE clause and acts as a Filter finally.

### ORDER BY clause:

There is a trick for ORDER BY for both origin query and view query. As we has no Groupy By curretly, the aggregation results would be either one or zero rows that make the Order By clause pointless.
We could avoid considering the sort columns if it's a junk for view matching.

### LIMIT clause:

As we have no group by for view with aggs now, the final result would be either one or zero row.
LIMIT, OFFSET clause of origin query could be applied to view if there are consts.
```sql
create incremental materialized view mv as
  select sum(c1) as mc1, count(c2) as mc2, avg(c3) as mc3, count(*) as
mc4
  from t where c1 > 90;
```
Query:
```sql
  select count(*), sum(c1) from t where c1 > 90 limit 2;
```

Could be rewritten to:
```sql
  select mc4, mc1 from mv limit 2;
```

fix #ISSUE_Number

---

### Change logs

### Contributor's Checklist

- [ ] Document changes.

### DIFF
--- a/src/backend/optimizer/plan/aqumv.c
+++ b/src/backend/optimizer/plan/aqumv.c
@@ -122,7 +122,6 @@ answer_query_using_materialized_views(PlannerInfo *root, AqumvContext aqumv_cont
 	RelOptInfo *current_rel = aqumv_context->current_rel;
 	query_pathkeys_callback qp_callback = aqumv_context->qp_callback;
 	void *qp_extra = aqumv_context->qp_extra;
-	List *raw_processed_tlist = aqumv_context->raw_processed_tlist;
 	Node *raw_havingQual = aqumv_context->raw_havingQual;
 
 	Query   		*parse = root->parse; /* Query of origin SQL. */
@@ -359,27 +358,9 @@ answer_query_using_materialized_views(PlannerInfo *root, AqumvContext aqumv_cont
 			 * As we has no Groupy By here, the aggregation results would be either one or
 			 * zero rows that make the Order By clause pointless, except that there were
 			 * SRF.
-			 * We could avoid considering the sort columns if it's a junk for view matching.
-			 * This in-place updates raw_processed_tlist.
 			 */
 			if (parse->sortClause != NIL || viewQuery->sortClause != NIL)
 			{
-				ListCell *lc;
-				ListCell *lcc;
-
-				foreach (lc, raw_processed_tlist)
-				{
-					TargetEntry *tle = (TargetEntry *) lfirst(lc);
-					if (!tle->resjunk || (0 == tle->ressortgroupref))
-						continue;
-
-					foreach (lcc, parse->sortClause)
-					{
-						SortGroupClause *srt = (SortGroupClause *) lfirst(lcc);
-						if (tle->ressortgroupref == srt->tleSortGroupRef)
-							foreach_delete_current(raw_processed_tlist, lc);
-					}
-				}
 				/* Earse view's sort caluse, it's ok to let alone view's target list. */
 				viewQuery->sortClause = NIL;
 			}
@@ -432,7 +413,7 @@ answer_query_using_materialized_views(PlannerInfo *root, AqumvContext aqumv_cont
 				continue;
 			subroot->hasHavingQual = false;
 
-			if(!aqumv_process_targetlist(context, raw_processed_tlist, &mv_final_tlist))
+			if(!aqumv_process_targetlist(context, aqumv_context->raw_processed_tlist, &mv_final_tlist))
 				continue;
 
 			viewQuery->targetList = mv_final_tlist;
--- a/src/backend/optimizer/plan/aqumv.c
+++ b/src/backend/optimizer/plan/aqumv.c
@@ -30,6 +30,7 @@
 #include "optimizer/planner.h"
 #include "optimizer/prep.h"
 #include "optimizer/tlist.h"
+#include "optimizer/transform.h"
 #include "parser/analyze.h"
 #include "parser/parsetree.h"
 #include "parser/parse_node.h"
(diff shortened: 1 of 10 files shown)


avamingli:
While develop #685, SRF also need to be considered here: https://github.com/apache/cloudberry/pull/685#issuecomment-2453949643
Will fix it later.

src/backend/optimizer/plan/aqumv.c
```diff
@@ -293,69 +325,195 @@ answer_query_using_materialized_views(PlannerInfo *root,
...
+				viewQuery->limitOffset = copyObject(parse->limitOffset);
+				viewQuery->limitOption = parse->limitOption;
+			}
+
+			preprocess_qual_conditions(subroot, (Node *) viewQuery->jointree);
+
+			if(!aqumv_process_from_quals(parse->jointree->quals, viewQuery->jointree->quals, &post_quals))
+				continue;
+
+			if (post_quals != NIL)
```
fanfuxiaoran, line 418: Why `post_quals` here cannot apply to the viewQuery?

src/backend/optimizer/plan/aqumv.c
```diff
@@ -293,69 +325,195 @@ answer_query_using_materialized_views(PlannerInfo *root,
...
-		 * NB: Update processed_tlist again in case that tlist has been changed. 
-		 */
-		preprocess_targetlist(subroot);
+			/*
+			 * There is a trick for ORDER BY for both origin query and view query.
+			 * As we has no Groupy By here, the aggregation results would be either one or
+			 * zero rows that make the Order By clause pointless.
+			 * We could avoid considering the sort columns if it's a junk for view matching.
+			 * This in-place update raw_processed_tlist.
+			 */
```
fanfuxiaoran, line 356: Does the below code block mean that if the TargetEntry in the `raw_processed_tlist` is `resjunk` and used for `order by` clause, then delete it from the list?
Beside the `order by` clause, can the TargetEntry be used in other caluses? such as `where`?

avamingli: > Beside the order by clause, can the TargetEntry be used in other caluses? such as where?

Can't be, if a column is only referred in `WHERE` and not in `SELECT`,  that column is not a TargetEntry.

You could gdb to see parse tree of an example SQL: `select a from t1 where b > 0`;

src/backend/optimizer/plan/aqumv.c
avamingli: We answer `aggregation` query directly, post_quals can't apple then.

fanfuxiaoran: Got it . Exactly, the `post_quals` cannot apply on the agg result.
Thanks for your explanation ,

src/backend/optimizer/plan/aqumv.c
fanfuxiaoran: You're right, only `sort` related columns appear in target list. 
But I think we don't need to optimize sort clauses here, as the pr https://github.com/apache/cloudberry/pull/685 has 
done that kind of work, the unused sort clauses from the query tree has been deleted,  they will not appear in raw_processed_tlist.

avamingli: Make sense, this PR comes earlier than #685 . 
Will pick this up after that one is merged, thanks~

avamingli:
Update: 
 Disable query has SRF to AQUMV and Remove raw_processed_tlist elimination.
 The origin SQL has removed useless ORDER BY clause, only process view query is enough.

src/backend/optimizer/plan/planner.c
```diff
@@ -1890,7 +1906,14 @@ grouping_planner(PlannerInfo *root, double tuple_fraction)
...
 		{
-			current_rel = answer_query_using_materialized_views(root, current_rel, standard_qp_callback, &qp_extra);
+			/* Now it's ok to set other fields. */
+			aqumv_context->current_rel = current_rel;
+			aqumv_context->qp_callback = standard_qp_callback;
+			aqumv_context->qp_extra = &qp_extra;
+
+			/* Do the real work. */
+			current_rel = answer_query_using_materialized_views(root, aqumv_context);
+			/* parse tree may be rewriten. */
```
yjhjstz, line 1916: here can deep free aqumv_context ?

avamingli: Good point, done.

avamingli: After a consideration, I think there is need to alloc it on heap.
There are just several pointers in fields. Done that at [35c6dbc](https://github.com/apache/cloudberry/pull/705/commits/35c6dbc7593f2c007402191548bfd1f6e766d093)

yjhjstz (approved):

my-ship-it (approved):

PR #719, opened 2024-11-20
Remove DockerHub auth from workflow

edespino:
Temporarily remove DOCKERHUB credentials since they're not being passed to workflows. Containers are public, so auth isn't required. This unblocks dev work while we investigate the credentials issue.

This fixes the workflow build error:

`Error: The template is not valid. .github/workflows/build-cloudberry.yml (Line: 190, Col: 19): Unexpected value '',.github/workflows/build-cloudberry.yml (Line: 191, Col: 19): Unexpected value ''
`

### DIFF
--- a/.github/workflows/build-cloudberry.yml
+++ b/.github/workflows/build-cloudberry.yml
@@ -58,10 +58,6 @@
 #    - Aggregates status from all jobs
 #    - Provides failure notifications if needed
 #
-# Required Secrets:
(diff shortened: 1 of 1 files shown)


edespino:
@yaooqinn & @tuhaihe Thank you!

reshke:
I'm too late to the party, bur still, approved + thanks for your contribution

PR #724, opened 2024-11-21
Push the runtime filter from HashJoin down to SeqScan.

zhangyue-hashdata:
Push the runtime filter from HashJoin down to SeqScan.

+----------+  AttrFilter   +------+  ScanKey   +---------+
| HashJoin | ------------> | Hash | ---------> | SeqScan |
+----------+               +------+            +---------+

If "gp_enable_runtime_filter_pushdown" is on, three steps will be run:

Step 1. In ExecInitHashJoin(), try to find the mapper between the var in
        hashclauses and the var in SeqScan. If found we will save the mapper in
        AttrFilter and push them to Hash node;

Step 2. We will create the range/bloom filters in AttrFilter during building
        hash table, and these filters will be converted to the list of ScanKey
        and pushed down to Seqscan when the building finishes;

Step 3. ScanKeys will be used to filter slot in Seqscan.

TODO:
1. support singlenode or utility mode;
2. support null value filter;
3. support Motion, SharedScan as target node;
4. support join qual like: t1.c1 = t2.c1+5;

perf:
CPU E5-2680 v2 10 cores, memory 32GB, 3 segments
1. tpcds 10s off: 939s on: 779s 17%
2. tpcds 100s off: 5270s on: 4365s 17%

tpcds 10s details
| NO.  |    off      |    on    |     |
|------|------------|-----------|-----|
|1|2959.5|1312.9|
|2|4522.2|2995.9|
|3|11924.1|9170.6|
|4|1678.4|1653.4|
|5-1|17433.1|14723.7|
|5-2|17244.1|14499.7|
|6|5541.8|4443.8|
|7|3144.0|1726.7|
|8|3895.7|2010.0|
|9|5991.1|3270.7|
|10|29981.7|19975.5|
|11|3113.9|2293.0|
|12|2166.9|1543.3|
|13|1258.8|726.0|
|14|11745.1|7037.3|
|15|3878.6|2568.8|
|16|5420.8|3200.1|
|17|3291.2|2117.3|
|18|2103.7|979.8|
|19|4400.1|2371.1|
|20|14048.3|11135.6|
|21|7739.5|7489.6|
|22|5524.1|3083.1|
|23|5978.4|4041.7|
|24|2355.7|1993.0|
|25|5561.1|3355.0|
|26|6973.0|3180.3|
|27|3536.0|2134.0|
|28|6899.6|3141.5|
|29|6065.5|3766.1|
|30|15443.3|15137.7|
|31|9197.8|6387.3|
|32|3345.3|1815.2|
|33|2488.1|2451.7|
|34|10840.0|9291.0|
|35|52991.2|52504.5|
|36|3386.1|3381.4|
|37|1834.0|994.2|
|38|10047.6|8915.8|
|39|3035.8|1961.5|
|40|1591.5|795.5|
|41|9284.0|7623.0|
|42|10255.5|10051.7|
|43|45356.5|44808.1|
|44|1989.3|1015.8|
|45|3021.2|1222.5|
|46|19865.7|18714.8|
|47|8438.0|4972.7|
|48|4520.7|3117.3|
|49|6244.7|6150.8|
|50|5926.0|4983.4|
|51|20423.2|20287.7|
|52|84.5|85.5|
|53|9273.4|6599.0|
|54|10214.5|8094.6|
|55|38920.5|36340.8|
|56|4248.9|2761.0|
|57|1980.7|1264.9|
|58|3954.6|1806.1|
|59|2982.5|1234.5|
|60|4099.5|1852.3|
|61|3080.7|1237.5|
|62|167.6|162.6|
|63|691.0|661.1|
|64|1283.0|549.9|
|65|2033.5|1017.7|
|66|8816.5|5342.4|
|67|10124.9|6664.9|
|68-1|53775.3|53665.4|
|68-2|55517.4|54534.5|
|69-1|20955.6|18578.4|
|69-2|17315.6|15543.9|
|70|6269.5|6134.9|
|71|13798.9|12300.0|
|72|5199.1|2861.3|
|73|4017.0|2236.1|
|74|5578.1|5341.3|
|75|1056.9|687.7|
|76|16922.5|16252.2|
|77|10491.0|8369.5|
|78|4590.9|2885.0|
|79|3660.0|2081.5|
|80|967.8|516.1|
|81|6620.6|1254.1|
|82|3072.2|1214.4|
|83|5111.3|2710.2|
|84|7458.4|7442.9|
|85|2989.0|1955.6|
|86|2546.8|1522.6|
|87|45634.8|43461.0|
|88|3555.6|1858.1|
|89|5630.1|4286.9|
|90|5321.3|2992.1|
|91|4748.6|4522.8|
|92-1|6264.3|4859.7|
|92-2|6355.4|5334.0|
|93|31174.3|30890.3|
|94|3971.3|3955.2|
|95|5097.4|3004.3|
|96|1999.7|1100.9|
|97|5055.0|2974.5|
|98|7319.9|5527.3|
|99|1895.3|1838.4|
| total|  939       |  779      |   ~17%  |

tpcds 100s details

| NO.  |    off      |    on    |     |
|------|------------|-----------|-----|
|1|18410.5|8966.4|
|2|29278.7|17942.0|
|3|68332.8|52880.0|
|4|9753.3|9171.7|
|5-1|13129.4|11131.0|
|5-2|13198.6|11041.2|
|6|33375.2|27429.7|
|7|19269.8|10077.7|
|8|22145.1|12336.3|
|9|64808.8|56969.8|
|10|241285.3|161582.3|
|11|12075.9|10860.6|
|12|9563.9|8114.5|
|13|2687.4|1484.8|
|14|8932.8|5325.5|
|15|28111.9|20017.6|
|16|30523.0|19822.1|
|17|3168.7|3126.4|
|18|8063.8|5925.4|
|19|19826.5|10147.2|
|20|85671.7|80051.7|
|21|42448.4|38791.5|
|22|27746.9|15857.7|
|23|38362.9|25001.5|
|24|14143.9|13982.5|
|25|38145.1|18836.1|
|26|21448.6|12437.5|
|27|22101.6|14043.4|
|28|20168.0|12036.7|
|29|25105.1|13663.3|
|30|129563.3|129378.8|
|31|13768.9|7181.7|
|32|19039.5|11568.3|
|33|10040.4|9460.8|
|34|66348.7|57294.4|
|35|202825.5|197756.5|
|36|28703.6|28536.2|
|37|7065.0|6596.1|
|38|82447.6|73465.7|
|39|16568.5|10866.6|
|40|5632.2|2801.4|
|41|65710.0|54836.4|
|42|63198.1|61843.4|
|43|154716.5|151916.7|
|44|10148.0|5314.8|
|45|18080.4|8917.2|
|46|92069.5|84997.4|
|47|29166.7|21165.4|
|48|23630.6|20430.5|
|49|43623.1|43465.4|
|50|33574.3|24155.7|
|51|140174.1|136511.3|
|52|121.1|108.4|
|53|61777.5|53161.0|
|54|82215.5|70357.6|
|55|28835.3|26623.2|
|56|24767.9|15975.6|
|57|11274.8|7005.6|
|58|20008.8|11185.9|
|59|19055.5|8743.5|
|60|30663.6|13416.8|
|61|18951.8|8765.2|
|62|205.8|170.3|
|63|22338.1|13134.9|
|64|5539.4|3077.0|
|65|9817.0|5337.0|
|66|52164.9|26808.1|
|67|20848.3|9974.1|
|68-1|462977.0|458799.8|
|68-2|468924.8|464181.1|
|69-1|109512.7|103476.2|
|69-2|96662.2|88990.1|
|70|29996.5|28361.0|
|71|90191.5|75458.1|
|72|35578.6|20232.3|
|73|23311.9|13611.2|
|74|33299.4|30755.5|
|75|4249.5|3892.5|
|76|107381.3|102310.4|
|77|83335.9|72471.8|
|78|24223.1|15703.0|
|79|22379.5|14098.0|
|80|2753.5|1177.6|
|81|39643.7|15849.0|
|82|18515.7|8568.5|
|83|21661.9|11949.9|
|84|55856.0|55736.4|
|85|15100.4|9251.1|
|86|14383.5|9051.7|
|87|136009.8|134747.4|
|88|20819.7|12613.0|
|89|40187.3|31457.0|
|90|34545.7|12278.3|
|91|34661.6|32747.8|
|92-1|102286.6|31937.6|
|92-2|103260.1|31959.5|
|93|240605.7|235582.7|
|94|26150.7|25874.4|
|95|31199.3|20272.8|
|96|4154.5|2226.1|
|97|26195.2|17051.1|
|98|37977.6|24553.3|
|99|16624.3|16603.3|
| total|  5270       |  4365      |   ~17%  |

Fixes #ISSUE_Number

### Type of Change
- [ ] Documentation update

### Test Plan

### Impact

**Performance:**

**User-facing changes:**

**Dependencies:**

### Checklist

### Additional Context

⚠️ **To skip CI:** Add `[skip ci]` to your PR title. Only use when necessary! ⚠️

---

### DIFF
--- a/src/backend/executor/nodeHash.c
+++ b/src/backend/executor/nodeHash.c
@@ -101,7 +101,10 @@ static bool ExecParallelHashTuplePrealloc(HashJoinTable hashtable,
 										  size_t size);
 static void ExecParallelHashMergeCounters(HashJoinTable hashtable);
 static void ExecParallelHashCloseBatchAccessors(HashJoinTable hashtable);
-
+static void AddTupleValuesIntoRF(HashState *node, TupleTableSlot *slot);
+static void PushdownRuntimeFilter(HashState *node);
+static void FreeRuntimeFilter(HashState *node);
+static void ResetRuntimeFilter(HashState *node);
 
 /* ----------------------------------------------------------------
  *		ExecHash
@@ -192,7 +195,15 @@ MultiExecPrivateHash(HashState *node)
 	{
 		slot = ExecProcNode(outerNode);
 		if (TupIsNull(slot))
+		{
+			if (gp_enable_runtime_filter_pushdown && node->filters)
+				PushdownRuntimeFilter(node);
 			break;
+		}
+
+		if (gp_enable_runtime_filter_pushdown && node->filters)
+			AddTupleValuesIntoRF(node, slot);
+
 		/* We have to compute the hash value */
 		econtext->ecxt_outertuple = slot;
 		bool hashkeys_null = false;
@@ -334,6 +345,7 @@ MultiExecParallelHash(HashState *node)
 				slot = ExecProcNode(outerNode);
 				if (TupIsNull(slot))
 					break;
+
 				econtext->ecxt_outertuple = slot;
 				if (ExecHashGetHashValue(node, hashtable, econtext, hashkeys,
 										 false, hashtable->keepNulls,
@@ -511,6 +523,9 @@ ExecEndHash(HashState *node)
 	 */
 	outerPlan = outerPlanState(node);
 	ExecEndNode(outerPlan);
+
+	if (node->filters)
+		FreeRuntimeFilter(node);
 }
 
 
@@ -2519,6 +2534,9 @@ ExecReScanHash(HashState *node)
 	 */
 	if (node->ps.lefttree->chgParam == NULL)
 		ExecReScan(node->ps.lefttree);
+
+	if (gp_enable_runtime_filter_pushdown && node->filters)
+		ResetRuntimeFilter(node);
 }
 
 
@@ -4124,4 +4142,143 @@ get_hash_mem(void)
 	mem_limit = Min(mem_limit, (size_t) MAX_KILOBYTES);
 
 	return (int) mem_limit;
+}
+
+/*
+ * Convert AttrFilter to ScanKeyData and send these runtime filters to the
+ * target node(seqscan).
+ */
+void
+PushdownRuntimeFilter(HashState *node)
+{
+	ListCell	*lc;
+	List		*scankeys;
+	ScanKey		sk;
+	AttrFilter	*attr_filter;
+
+	foreach (lc, node->filters)
+	{
+		scankeys = NIL;
+
+		attr_filter = lfirst(lc);
+		if (!IsA(attr_filter->target, SeqScanState) || attr_filter->empty)
+			continue;
+
+		/* bloom filter */
+		sk = (ScanKey)palloc0(sizeof(ScanKeyData));
+		sk->sk_flags    = SK_BLOOM_FILTER;
+		sk->sk_attno    = attr_filter->lattno;
+		sk->sk_subtype  = INT8OID;
+		sk->sk_argument = PointerGetDatum(attr_filter->blm_filter);
+		scankeys = lappend(scankeys, sk);
+
+		/* range filter */
+		sk = (ScanKey)palloc0(sizeof(ScanKeyData));
+		sk->sk_flags    = 0;
+		sk->sk_attno    = attr_filter->lattno;
+		sk->sk_strategy = BTGreaterEqualStrategyNumber;
+		sk->sk_subtype  = INT8OID;
+		sk->sk_argument = attr_filter->min;
+		scankeys = lappend(scankeys, sk);
+
+		sk = (ScanKey)palloc0(sizeof(ScanKeyData));
+		sk->sk_flags    = 0;
+		sk->sk_attno    = attr_filter->lattno;
+		sk->sk_strategy = BTLessEqualStrategyNumber;
+		sk->sk_subtype  = INT8OID;
+		sk->sk_argument = attr_filter->max;
+		scankeys = lappend(scankeys, sk);
+
+		/* append new runtime filters to target node */
+		SeqScanState *sss = castNode(SeqScanState, attr_filter->target);
+		sss->filters = list_concat(sss->filters, scankeys);
+	}
+}
+
+static void
+AddTupleValuesIntoRF(HashState *node, TupleTableSlot *slot)
+{
+	Datum val;
+	bool  isnull;
+	ListCell	*lc;
+	AttrFilter	*attr_filter;
+
+	foreach (lc, node->filters)
+	{
+		attr_filter = (AttrFilter *) lfirst(lc);
+
+		val = slot_getattr(slot, attr_filter->rattno, &isnull);
+		if (isnull)
+			continue;
+
+		attr_filter->empty = false;
+
+		if ((int64_t)val < (int64_t)attr_filter->min)
+			attr_filter->min = val;
+
+		if ((int64_t)val > (int64_t)attr_filter->max)
+			attr_filter->max = val;
+
+		if (attr_filter->blm_filter)
+			bloom_add_element(attr_filter->blm_filter, (unsigned char *)&val, sizeof(Datum));
+	}
+}
+
+void
+FreeRuntimeFilter(HashState *node)
+{
+	ListCell	*lc;
+	AttrFilter	*attr_filter;
+
+	if (!node->filters)
+		return;
+
+	foreach (lc, node->filters)
+	{
+		attr_filter = lfirst(lc);
+		if (attr_filter->blm_filter)
+			bloom_free(attr_filter->blm_filter);
+	}
+
+	list_free_deep(node->filters);
+	node->filters = NIL;
+}
+
+void
+ResetRuntimeFilter(HashState *node)
+{
+	ListCell		*lc;
+	AttrFilter		*attr_filter;
+	SeqScanState	*sss;
+
+	if (!node->filters)
+		return;
+
+	foreach (lc, node->filters)
+	{
+		attr_filter = lfirst(lc);
+		attr_filter->empty  = true;
+
+		if (IsA(attr_filter->target, SeqScanState))
+		{
+			sss = castNode(SeqScanState, attr_filter->target);
+			if (sss->filters)
+			{
+				list_free_deep(sss->filters);
+				sss->filters = NIL;
+			}
+		}
+
+		if (attr_filter->blm_filter)
+			bloom_free(attr_filter->blm_filter);
+
+		attr_filter->blm_filter = bloom_create_aggresive(node->ps.plan->plan_rows,
+														 work_mem,
+														 random());
+
+		StaticAssertDecl(sizeof(LONG_MAX) == sizeof(Datum), "sizeof(LONG_MAX) should be equal to sizeof(Datum)");
+		StaticAssertDecl(sizeof(LONG_MIN) == sizeof(Datum), "sizeof(LONG_MIN) should be equal to sizeof(Datum)");
+		attr_filter->min    = LONG_MAX;
+		attr_filter->max    = LONG_MIN;
+	}
 }
--- a/src/backend/executor/nodeHashjoin.c
+++ b/src/backend/executor/nodeHashjoin.c
@@ -118,6 +118,8 @@
 #include "executor/nodeRuntimeFilter.h"
 #include "miscadmin.h"
 #include "pgstat.h"
+#include "utils/guc.h"
+#include "utils/fmgroids.h"
 #include "utils/memutils.h"
 #include "utils/sharedtuplestore.h"
 
@@ -161,6 +163,19 @@ static void ReleaseHashTable(HashJoinState *node);
 static void SpillCurrentBatch(HashJoinState *node);
 static bool ExecHashJoinReloadHashTable(HashJoinState *hjstate);
 static void ExecEagerFreeHashJoin(HashJoinState *node);
+static void CreateRuntimeFilter(HashJoinState* hjstate);
+static bool IsEqualOp(Expr *expr);
+static bool CheckEqualArgs(Expr *expr, AttrNumber *lattno, AttrNumber *rattno);
+static bool CheckTargetNode(PlanState *node,
+							AttrNumber attno,
+							AttrNumber *lattno);
+static List *FindTargetNodes(HashJoinState *hjstate,
+							 AttrNumber attno,
+							 AttrNumber *lattno);
+static AttrFilter *CreateAttrFilter(PlanState *target,
+									AttrNumber lattno,
+									AttrNumber rattno,
+									double plan_rows);
 extern bool Test_print_prefetch_joinqual;
 
 
@@ -998,6 +1013,11 @@ ExecInitHashJoin(HashJoin *node, EState *estate, int eflags)
 		hstate->rfstate = rfstate;
 		ExecInitRuntimeFilterFinish(rfstate, hstate->ps.plan->plan_rows);
 	}
+
+	if (Gp_role == GP_ROLE_EXECUTE
+		&& gp_enable_runtime_filter_pushdown
+		&& !estate->useMppParallelMode)
+		CreateRuntimeFilter(hjstate);
 
 	return hjstate;
 }
@@ -2155,4 +2175,294 @@ ExecHashJoinInitializeWorker(HashJoinState *state,
 		state->worker_id = ParallelWorkerNumber + 1;
 		ExecSetExecProcNode(&state->js.ps, ExecParallelHashJoin);
 	}
+}
+
+/*
+ * Find "inner var = outer var" in hj->hashclauses and create runtime filter
+ * for it.
+ */
+void
+CreateRuntimeFilter(HashJoinState* hjstate)
+{
+	AttrNumber lattno, rattno;
+	Expr		*expr;
+	JoinType	jointype;
+	HashJoin	*hj;
+	HashState	*hstate;
+	AttrFilter	*attr_filter;
+	ListCell	*lc;
+	List		*targets;
+
+	/*
+	 * A build-side Bloom filter tells us if a row is definitely not in the build
+	 * side. This allows us to early-eliminate rows or early-accept rows depending
+	 * on the type of join.
+	 * Left Outer Join and Full Outer Join output all rows, so a build-side Bloom
+	 * filter would only allow us to early-output. Left Antijoin outputs only if
+	 * there is no match, so again early output. We don't implement early output
+	 * for now.
+	 * So it's only applicatable for inner, right and semi join.
+	 */
+	jointype = hjstate->js.jointype;
+	if (jointype != JOIN_INNER &&
+		jointype != JOIN_RIGHT &&
+		jointype != JOIN_SEMI)
+		return;
+
+	hstate = castNode(HashState, innerPlanState(hjstate));
+	hstate->filters = NIL;
+
+	/*
+	 * check and initialize the runtime filter for all hash conds in
+	 * hj->hashclauses
+	 */
+	hj = castNode(HashJoin, hjstate->js.ps.plan);
+	foreach (lc, hj->hashclauses)
+	{
+		expr = (Expr *)lfirst(lc);
+
+		if (!IsEqualOp(expr))
+			continue;
+
+		lattno = -1;
+		rattno = -1;
+		if (!CheckEqualArgs(expr, &lattno, &rattno))
+			continue;
+
+		if (lattno < 1 || rattno < 1)
+			continue;
+
+		targets = FindTargetNodes(hjstate, lattno, &lattno);
+		if (lattno == -1 || targets == NULL)
+			continue;
+
+		foreach(lc, targets)
+		{
+			PlanState *target = lfirst(lc);
+			Assert(IsA(target, SeqScanState));
+
+			attr_filter = CreateAttrFilter(target, lattno, rattno,
+					hstate->ps.plan->plan_rows);
+			if (attr_filter->blm_filter)
+				hstate->filters = lappend(hstate->filters, attr_filter);
+			else
+				pfree(attr_filter);
+		}
+	}
+}
+
+static bool
+IsEqualOp(Expr *expr)
+{
+	Oid funcid = InvalidOid;
+
+	if (!IsA(expr, OpExpr) && !IsA(expr, FuncExpr))
+		return false;
+
+	if (IsA(expr, OpExpr))
+		funcid = ((OpExpr *)expr)->opfuncid;
+	else if (IsA(expr, FuncExpr))
+		funcid = ((FuncExpr *)expr)->funcid;
+	else
+		return false;
+
+	if (funcid == F_INT2EQ  || funcid == F_INT4EQ  || funcid == F_INT8EQ
+		|| funcid == F_INT24EQ || funcid == F_INT42EQ
+		|| funcid == F_INT28EQ || funcid == F_INT82EQ
+		|| funcid == F_INT48EQ || funcid == F_INT84EQ
+	   )
+		return true;
+
+	return false;
+}
+
+/*
+ * runtime filters which can be pushed down:
+ * 1. hash expr MUST BE equal op;
+ * 2. args MUST BE Var node;
+ * 3. the data type MUST BE integer;
+ */
+static bool
+CheckEqualArgs(Expr *expr, AttrNumber *lattno, AttrNumber *rattno)
+{
+	Var		*var;
+	List	*args;
+
+	if (lattno == NULL || rattno == NULL)
+		return false;
+
+	if (IsA(expr, OpExpr))
+		args = ((OpExpr *)expr)->args;
+	else if (IsA(expr, FuncExpr))
+		args = ((FuncExpr *)expr)->args;
+	else
+		return false;
+
+	if (!args || list_length(args) != 2)
+		return false;
+
+	/* check the first arg */
+	if (!IsA(linitial(args), Var))
+		return false;
+
+	var = linitial(args);
+	if (var->varno == INNER_VAR)
+		*rattno = var->varattno;
+	else if (var->varno == OUTER_VAR)
+		*lattno = var->varattno;
+	else
+		return false;
+
+	/* check the second arg */
+	if (!IsA(lsecond(args), Var))
+		return false;
+
+	var = lsecond(args);
+	if (var->varno == INNER_VAR)
+		*rattno = var->varattno;
+	else if (var->varno == OUTER_VAR)
+		*lattno = var->varattno;
+	else
+		return false;
+
+	return true;
+}
+
+static bool
+CheckTargetNode(PlanState *node, AttrNumber attno, AttrNumber *lattno)
+{
+	Var *var;
+	TargetEntry *te;
+
+	if (!IsA(node, SeqScanState))
+		return false;
+
+	te = (TargetEntry *)list_nth(node->plan->targetlist, attno - 1);
+	if (!IsA(te->expr, Var))
+		return false;
+
+	var = castNode(Var, te->expr);
+
+	/* system column is not allowed */
+	if (var->varattno <= 0)
+		return false;
+
+	*lattno = var->varattno;
+
+	return true;
+}
+
+/*
+ * it's just allowed like this:
+ *   HashJoin
+ *      ... a series of HashJoin nodes
+ *        HashJoin
+ *          SeqScan <- target
+ */
+static List *
+FindTargetNodes(HashJoinState *hjstate, AttrNumber attno, AttrNumber *lattno)
+{
+	Var *var;
+	PlanState *child, *parent;
+	TargetEntry *te;
+	List *targetNodes;
+
+	parent = (PlanState *)hjstate;
+	child  = outerPlanState(hjstate);
+	Assert(child);
+
+	*lattno = -1;
+	targetNodes = NIL;
+	while (true)
+	{
+		/* target is seqscan */
+		if ((IsA(parent, HashJoinState) || IsA(parent, ResultState)) && IsA(child, SeqScanState))
+		{
+			/*
+			 * hashjoin
+			 *   seqscan
+			 * or
+			 * hashjoin
+			 *   result
+			 *     seqscan
+			 */
+			if (!CheckTargetNode(child, attno, lattno))
+				return NULL;
+
+			targetNodes = lappend(targetNodes, child);
+			return targetNodes;
+		}
+		else if (IsA(parent, AppendState) && child == NULL)
+		{
+			/*
+			 * append
+			 *   seqscan on t1_prt_1
+			 *   seqscan on t1_prt_2
+			 *   ...
+			 */
+			AppendState *as = castNode(AppendState, parent);
+			for (int i = 0; i < as->as_nplans; i++)
+			{
+				child = as->appendplans[i];
+				if (!CheckTargetNode(child, attno, lattno))
+					return NULL;
+
+				targetNodes = lappend(targetNodes, child);
+			}
+
+			return targetNodes;
+		}
+
+		/*
+		 * hashjoin
+		 *   result (hash filter)
+		 *     seqscan on t1, t1 is replicated table
+		 * or
+		 * hashjoin
+		 *   append
+		 *     seqscan on t1_prt_1
+		 *     seqscan on t1_prt_2
+		 *     ...
+		 */
+		if (!IsA(child, HashJoinState) && !IsA(child, ResultState) && !IsA(child, AppendState))
+			return NULL;
+
+		/* child is hashjoin, result or append node */
+		te = (TargetEntry *)list_nth(child->plan->targetlist, attno - 1);
+		if (!IsA(te->expr, Var))
+			return NULL;
+
+		var = castNode(Var, te->expr);
+		if (var->varno == INNER_VAR)
+			return NULL;
+
+		attno = var->varattno;
+
+		/* find at child node */
+		parent = child;
+		child  = outerPlanState(parent);
+	}
+
+	return NULL;
+}
+
+static AttrFilter*
+CreateAttrFilter(PlanState *target, AttrNumber lattno, AttrNumber rattno,
+				 double plan_rows)
+{
+	AttrFilter *attr_filter = palloc0(sizeof(AttrFilter));
+	attr_filter->empty  = true;
+	attr_filter->target = target;
+
+	attr_filter->lattno = lattno;
+	attr_filter->rattno = rattno;
+
+	attr_filter->blm_filter = bloom_create_aggresive(plan_rows, work_mem, random());
+
+	StaticAssertDecl(sizeof(LONG_MAX) == sizeof(Datum), "sizeof(LONG_MAX) should be equal to sizeof(Datum)");
+	StaticAssertDecl(sizeof(LONG_MIN) == sizeof(Datum), "sizeof(LONG_MIN) should be equal to sizeof(Datum)");
+	attr_filter->min = LONG_MAX;
+	attr_filter->max = LONG_MIN;
+
+	return attr_filter;
 }
--- a/src/backend/executor/nodeSeqscan.c
+++ b/src/backend/executor/nodeSeqscan.c
@@ -43,6 +43,9 @@
 
 static TupleTableSlot *SeqNext(SeqScanState *node);
 
+static bool PassByBloomFilter(SeqScanState *node, TupleTableSlot *slot);
+static ScanKey ScanKeyListToArray(List *keys, int *num);
+
 /* ----------------------------------------------------------------
  *						Scan Support
  * ----------------------------------------------------------------
@@ -71,13 +74,24 @@ SeqNext(SeqScanState *node)
 
 	if (scandesc == NULL)
 	{
+		int nkeys = 0;
+		ScanKey keys = NULL;
+
+		/*
+		 * Just when gp_enable_runtime_filter_pushdown enabled and
+		 * node->filter_in_seqscan is false means scankey need to be pushed to
+		 * AM.
+		 */
+		if (gp_enable_runtime_filter_pushdown && !node->filter_in_seqscan)
+			keys = ScanKeyListToArray(node->filters, &nkeys);
+
 		/*
 		* We reach here if the scan is not parallel, or if we're serially
 		* executing a scan that was planned to be parallel.
 		*/
 		scandesc = table_beginscan_es(node->ss.ss_currentRelation,
 									  estate->es_snapshot,
-									  0, NULL,
+									  nkeys, keys,
 									  NULL,
 									  &node->ss.ps);
 		node->ss.ss_currentScanDesc = scandesc;
@@ -86,8 +100,22 @@ SeqNext(SeqScanState *node)
 	/*
 	 * get the next tuple from the table
 	 */
-	if (table_scan_getnextslot(scandesc, direction, slot))
-		return slot;
+	if (node->filter_in_seqscan && node->filters)
+	{
+		while (table_scan_getnextslot(scandesc, direction, slot))
+		{
+			if (!PassByBloomFilter(node, slot))
+				continue;
+
+			return slot;
+		}
+	}
+	else
+	{
+		if (table_scan_getnextslot(scandesc, direction, slot))
+			return slot;
+	}
+
 	return NULL;
 }
 
@@ -190,6 +218,15 @@ ExecInitSeqScanForPartition(SeqScan *node, EState *estate,
 	 */
 	scanstate->ss.ps.qual =
 		ExecInitQual(node->plan.qual, (PlanState *) scanstate);
+
+	/*
+	 * check scan slot with bloom filters in seqscan node or not.
+	 */
+	if (gp_enable_runtime_filter_pushdown
+		&& !estate->useMppParallelMode)
+	{
+		scanstate->filter_in_seqscan = true;
+	}
 
 	return scanstate;
 }
@@ -363,4 +400,72 @@ ExecSeqScanInitializeWorker(SeqScanState *node,
 		scandesc = table_beginscan_parallel(node->ss.ss_currentRelation, pscan);
 	}
 	node->ss.ss_currentScanDesc = scandesc;
+}
+
+/*
+ * Returns true if the element may be in the bloom filter.
+ */
+static bool
+PassByBloomFilter(SeqScanState *node, TupleTableSlot *slot)
+{
+	ScanKey	sk;
+	Datum	val;
+	bool	isnull;
+	ListCell *lc;
+	bloom_filter *blm_filter;
+
+	/*
+	 * Mark that the pushdown runtime filter is actually taking effect.
+	 */
+	if (node->ss.ps.instrument &&
+		!node->ss.ps.instrument->prf_work &&
+		list_length(node->filters))
+		node->ss.ps.instrument->prf_work = true;
+
+	foreach (lc, node->filters)
+	{
+		sk = lfirst(lc);
+		if (sk->sk_flags != SK_BLOOM_FILTER)
+			continue;
+
+		val = slot_getattr(slot, sk->sk_attno, &isnull);
+		if (isnull)
+			continue;
+
+		blm_filter = (bloom_filter *)DatumGetPointer(sk->sk_argument);
+		if (bloom_lacks_element(blm_filter, (unsigned char *)&val, sizeof(Datum)))
+		{
+			InstrCountFilteredPRF(node, 1);
+			return false;
+		}
+	}
+
+	return true;
+}
+
+/*
+ * Convert the list of ScanKey to the array, and append an emtpy ScanKey as
+ * the end flag of the array.
+ */
+static ScanKey
+ScanKeyListToArray(List *keys, int *num)
+{
+	ScanKey sk;
+
+	if (list_length(keys) == 0)
+		return NULL;
+
+	Assert(num);
+	*num = list_length(keys);
+
+	sk = (ScanKey)palloc(sizeof(ScanKeyData) * (*num + 1));
+	for (int i = 0; i < *num; ++i)
+		memcpy(&sk[i], list_nth(keys, i), sizeof(ScanKeyData));
+
+	/*
+	 * SK_EMPYT means the end of the array of the ScanKey
+	 */
+	sk[*num].sk_flags = SK_EMPYT;
+
+	return sk;
 }
--- a/src/test/regress/expected/gp_runtime_filter.out
+++ b/src/test/regress/expected/gp_runtime_filter.out
@@ -1,3 +1,7 @@
+-- start_matchignore
+-- m/^.*Extra Text:.*/
+-- m/^.*Buckets:.*/
+-- end_matchignore
 -- Disable ORCA
 SET optimizer TO off;
 -- Test Suit 1: runtime filter main case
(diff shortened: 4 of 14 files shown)


src/backend/executor/nodeHash.c
```diff
@@ -4126,3 +4151,138 @@ get_hash_mem(void)
...
+		sk->sk_flags    = 0;
+		sk->sk_attno    = attr_filter->lattno;
+		sk->sk_strategy = BTLessEqualStrategyNumber;
+		sk->sk_subtype  = INT8OID;
+		sk->sk_argument = attr_filter->max;
+		scankeys = lappend(scankeys, sk);
+
+		/* append new runtime filters to target node */
+		SeqScanState *sss = castNode(SeqScanState, attr_filter->target);
+		sss->filters = list_concat(sss->filters, scankeys);
```
yjhjstz, line 4202: can we merge filter here on the same attno ?

src/backend/executor/nodeSeqscan.c
```diff
@@ -365,3 +398,60 @@ ExecSeqScanInitializeWorker(SeqScanState *node,
...
+	bloom_filter *blm_filter;
+
+	foreach (lc, node->filters)
+	{
+		sk = lfirst(lc);
+		if (sk->sk_flags != SK_BLOOM_FILTER)
+			continue;
+
+		val = slot_getattr(slot, sk->sk_attno, &isnull);
+		if (isnull)
```
yjhjstz, line 421:
```c++
CREATE TABLE distinct_1(a int);
CREATE TABLE distinct_2(a int);
INSERT INTO distinct_1 VALUES(1),(2),(NULL);
INSERT INTO distinct_2 VALUES(1),(NULL);
SELECT * FROM distinct_1, distinct_2 WHERE distinct_1.a IS NOT DISTINCT FROM distinct_2.a;
```
test got wrong result.

src/backend/executor/nodeSeqscan.c
```diff
@@ -87,8 +101,17 @@ SeqNext(SeqScanState *node)
...
 	 * get the next tuple from the table
 	 */
-	if (table_scan_getnextslot(scandesc, direction, slot))
+	while (table_scan_getnextslot(scandesc, direction, slot))
+	{
+		if (TupIsNull(slot))
+			return slot;
+
+		if (node->filter_in_seqscan && node->filters &&
+			!PassByBloomFilter(node, slot))
```
yjhjstz, line 110: tpcds 1TB, bloom filter will lose efficacy or create failed due to large rows ?

src/test/regress/expected/gp_runtime_filter.out
```diff
@@ -250,6 +250,60 @@ SELECT COUNT(*) FROM dim_rf
...
+SELECT t1.c3 FROM t1, t2 WHERE t1.c2 = t2.c2;
+                                        QUERY PLAN                                         
+-------------------------------------------------------------------------------------------
+ Gather Motion 3:1  (slice1; segments: 3) (actual rows=0 loops=1)
+   ->  Hash Join (never executed)
+         Hash Cond: (t1.c2 = t2.c2)
+         Extra Text: (seg2)   Hash chain length 8.0 avg, 8 max, using 4 of 524288 buckets.
+         ->  Seq Scan on t1 (actual rows=1 loops=1)
+         ->  Hash (actual rows=32 loops=1)
+               Buckets: 524288  Batches: 1  Memory Usage: 4098kB
```
yjhjstz, line 299: how to debug and get pushdown scankey here ?

src/backend/executor/nodeHashjoin.c
```diff
@@ -2157,3 +2174,225 @@ ExecHashJoinInitializeWorker(HashJoinState *state,
...
+
+	if (!args || list_length(args) != 2)
+		return false;
+
+	match = false;
+	foreach (lc, args)
+	{
+		match = false;
+
+		if (!IsA(lfirst(lc), Var))
```
fanfuxiaoran, line 2301: Could it support other expression, whose one arg is the column attr, and the other is a const?

fanfuxiaoran:
Looks interesting. And I have some questions to discuss.
- Beside the seqscan, can the runtime filter apply to other types of scan? such as the index scan.
- Looks only when the `hashjoin` node and `seqscan` node run in the same process can use the runtime filter. Which means the tables should have same distributed policy on the join columns or one of the table is replicated.

src/backend/executor/nodeHashjoin.c
```diff
@@ -2157,3 +2174,225 @@ ExecHashJoinInitializeWorker(HashJoinState *state,
...
+			*lattno = var->varattno;
+			return child;
+		}
+
+		/*
+		 * hashjoin
+		 *   result (hash filter)
+		 *     seqscan on t1, t1 is replicated table
+		 */
+		if (!IsA(child, HashJoinState) && !IsA(child, ResultState))
```
yjhjstz, line 2361:
```c++
Hash Join  (cost=0.00..4019.55 rows=37 width=9) (actual time=3203.012..9927.435 rows=1399 loops=1)
                                                   Hash Cond: (web_sales_1_prt_2.ws_item_sk = item.i_item_sk)
                                                   Join Filter: (web_sales_1_prt_2.ws_ext_discount_amt > ((1.3 * avg(web_sales_1_prt_2_1.ws_ext_discount_amt))))
                                                   Rows Removed by Join Filter: 4763
                                                   Extra Text: (seg2)   Hash chain length 1.0 avg, 1 max, using 198 of 2097152 buckets.
                                                   ->  Append  (cost=0.00..676.44 rows=2399189 width=13) (actual time=16.899..5572.473 rows=3090021 loops=1)
                                                         ->  Seq Scan on web_sales_1_prt_2  (cost=0.00..676.44 rows=2399189 width=13) (actual time=16.895..1138.267 rows=662
149 loops=1)
                                                         ->  Seq Scan on web_sales_1_prt_3  (cost=0.00..676.44 rows=2399189 width=13) (actual time=8.947..1102.409 rows=6621
36 loops=1)
                                                         ->  Seq Scan on web_sales_1_prt_4  (cost=0.00..676.44 rows=2399189 width=13) (actual time=8.822..1100.839 rows=6621
48 loops=1)
                                                         ->  Seq Scan on web_sales_1_prt_5  (cost=0.00..676.44 rows=2399189 width=13) (actual time=11.391..1083.785 rows=662
179 loops=1)
                                                         ->  Seq Scan on web_sales_1_prt_6  (cost=0.00..676.44 rows=2399189 width=13) (actual time=13.030..649.141 rows=4414
09 loops=1)
                                                         ->  Seq Scan on web_sales_1_prt_7  (cost=0.00..676.44 rows=2399189 width=13) (never executed)
                                                         ->  Seq Scan on web_sales_1_prt_others  (cost=0.00..676.44 rows=2399189 width=13) (actual time=1.213..3.203 rows=17
88 loops=1)
                                                   ->  Hash  (cost=2432.09..2432.09 rows=109 width=12) (actual time=3177.768..3177.770 rows=198 loops=1)
                                                         Buckets: 2097152  Batches: 1  Memory Usage: 16392kB
                                                         ->  Broadcast Motion 3:3  (slice3; segments: 3)  (cost=
```
need to consider partitioned table .

src/backend/executor/nodeSeqscan.c
zhangyue-hashdata: I will fix it.

src/backend/executor/nodeHashjoin.c
zhangyue-hashdata: I will try.

zhangyue-hashdata:
> Looks interesting. And I have some questions to discuss.
> 
> * Beside the seqscan, can the runtime filter apply to other types of scan? such as the index scan.
> * Looks only when the `hashjoin` node and `seqscan` node run in the same process can use the runtime filter. Which means the tables should have same distributed policy on the join columns or one of the table is replicated.

src/backend/executor/nodeHash.c
```diff
@@ -4126,3 +4151,138 @@ get_hash_mem(void)
...
+			}
+		}
+
+		if (attr_filter->blm_filter)
+			bloom_free(attr_filter->blm_filter);
+
+		attr_filter->blm_filter     = bloom_create_aggresive(node->ps.plan->plan_rows,
+		                                             work_mem, random());
+		attr_filter->min    = LLONG_MAX;
+		attr_filter->max    = LLONG_MIN;
```
gfphoenix78, line 4286: LLONG_MAX, LLONG_MIN are platform-spec value, i.e. the bound value for `unsigned long long`, which may not be exactly the same width as Datum. For safety, static assert could be considered.

src/backend/executor/nodeHashjoin.c
```diff
@@ -2157,3 +2174,225 @@ ExecHashJoinInitializeWorker(HashJoinState *state,
...
+	JoinType	jointype;
+	HashJoin	*hj;
+	HashState	*hstate;
+	PlanState	*target;
+	AttrFilter	*attr_filter;
+	ListCell	*lc;
+
+	/*
+	 * Only applicatable for inner, right and semi join,
+	 */
```
gfphoenix78, line 2196: Could you give a little more explain about why these join types are supported and others are not?

src/backend/executor/nodeHashjoin.c
```diff
@@ -2157,3 +2174,225 @@ ExecHashJoinInitializeWorker(HashJoinState *state,
...
+	Var		*var;
+	bool	match;
+	List	*args;
+	ListCell *lc;
+
+	if (lattno == NULL || rattno == NULL)
+		return false;
+
+	if (!IsA(expr, OpExpr) && !IsA(expr, FuncExpr))
+		return false;
```
gfphoenix78, line 2284: These 2 lines duplicate with the following if-elseif-else code, could be deleted.

src/backend/executor/nodeHashjoin.c
```diff
@@ -2157,3 +2174,225 @@ ExecHashJoinInitializeWorker(HashJoinState *state,
...
+			*rattno = var->varattno;
+		else if (var->varno == OUTER_VAR)
+			*lattno = var->varattno;
+		else
+			break;
+
+		match = true;
+	}
+
+	return match;
```
gfphoenix78, line 2315: The match flag gets the code hard(several modifications) to read. The break statement could be replaced by `return false;`. If the foreach loop ends, all conditions match, so returns true.

src/backend/executor/nodeSeqscan.c
```diff
@@ -87,8 +101,17 @@ SeqNext(SeqScanState *node)
 	/*
 	 * get the next tuple from the table
 	 */
-	if (table_scan_getnextslot(scandesc, direction, slot))
+	while (table_scan_getnextslot(scandesc, direction, slot))
+	{
+		if (TupIsNull(slot))
+			return slot;
```
gfphoenix78, line 107: Will it be true?

src/backend/executor/nodeSeqscan.c
```diff
@@ -365,3 +398,60 @@ ExecSeqScanInitializeWorker(SeqScanState *node,
...
+	*num = list_length(keys);
+
+	sk = (ScanKey)palloc(sizeof(ScanKeyData) * (*num + 1));
+	for (int i = 0; i < *num; ++i)
+		memcpy(&sk[i], list_nth(keys, i), sizeof(ScanKeyData));
+
+	/*
+	 * SK_EMPYT means the end of the array of the ScanKey
+	 */
+	sk[*num].sk_flags = SK_EMPYT;
```
gfphoenix78, line 454: How to check the boundary of the `ScanKey` array in rescan? In normal rescan, the number of `ScanKey`s is the same as begin_scan. If the number of `ScanKey`s is larger in rescan than that in begin_scan, the boundary value might be invalid and dangerous to access.

src/test/regress/expected/gp_runtime_filter.out
gfphoenix78: The test case should cover that the hash join node or result node is the child of the parent hash join.

gfphoenix78: > how to debug and get pushdown scankey here ?

Add debug message to dump the stats about how many tuples are filter out?

src/backend/executor/nodeHash.c
zhangyue-hashdata: 1. Combining Bloom filters will result in a higher False Positive Rate (FPR) compared to using each of the individual Bloom filters separately, so it is not recommended;
2. There is the same problem to combine range filters like combining Bloom filters;
3. There is only one Bloom filter and one range filter on the same attribute in many cases;

src/backend/executor/nodeSeqscan.c
zhangyue-hashdata: It seems that slot is never NULL here, so `Assert(!TupIsNull(slot));` is better or remove them.

src/backend/executor/nodeHashjoin.c
zhangyue-hashdata: fix it

src/backend/executor/nodeHash.c
zhangyue-hashdata: I see `StaticAssertDecl(SIZEOF_DATUM == 8, "sizeof datum is not 8");` in postgres.h, so it's better to use INT64_MAX/INT64_MIN here.

src/backend/executor/nodeHash.c
yjhjstz:
```sql
create table rt2(a int, b int);
create table rt3(a int, b int);
insert into t1 select i, i from generate_series(1, 100000) i;
insert into t1 select i, i+1 from generate_series(1, 10) i;
insert into rt1 select i, i+1 from generate_series(1, 10) i;
insert into rt2 select i, i+1 from generate_series(1, 10000) i;
insert into rt3 select i, i+1 from generate_series(1, 10) i;
analyze t1;
analyze rt1;
analyze rt2;
analyze rt3;

explain analyze select * from rt1 join t1 on rt1.a = t1.b join rt3 on rt3.a = t1.b;

postgres=# explain select * from rt1 join t1 on rt1.a = t1.b join rt3 on rt3.a = t1.b;
                                   QUERY PLAN                                   
--------------------------------------------------------------------------------
 Gather Motion 3:1  (slice1; segments: 3)  (cost=2.45..428.51 rows=17 width=24)
   ->  Hash Join  (cost=2.45..428.29 rows=6 width=24)
         Hash Cond: (t1.b = rt1.a)
         ->  Hash Join  (cost=1.23..427.00 rows=6 width=16)
               Hash Cond: (t1.b = rt3.a)
               ->  Seq Scan on t1  (cost=0.00..342.37 rows=33337 width=8)
               ->  Hash  (cost=1.10..1.10 rows=10 width=8)
                     ->  Seq Scan on rt3  (cost=0.00..1.10 rows=10 width=8)
         ->  Hash  (cost=1.10..1.10 rows=10 width=8)
               ->  Seq Scan on rt1  (cost=0.00..1.10 rows=10 width=8)
(11 rows)

```
you can try this case, will got two range filters.

src/backend/executor/nodeSeqscan.c
yjhjstz, line 110: what's difference with `gp_enable_runtime_filter`'s bloom filter ?

src/backend/executor/nodeHash.c
zhangyue-hashdata: got it

src/backend/executor/nodeHash.c
yjhjstz: use  LONG_MAX, LONG_MIN instead ?

src/backend/executor/nodeHashjoin.c
```diff
@@ -162,6 +164,16 @@ static void ReleaseHashTable(HashJoinState *node);
 static void SpillCurrentBatch(HashJoinState *node);
 static bool ExecHashJoinReloadHashTable(HashJoinState *hjstate);
 static void ExecEagerFreeHashJoin(HashJoinState *node);
+static void CreateRuntimeFilter(HashJoinState* hjstate);
+static bool IsEqualOp(Expr *expr);
+static bool CheckEqualArgs(Expr *expr, AttrNumber *lattno, AttrNumber *rattno);
+static PlanState *FindTargetAttr(HashJoinState *hjstate,
```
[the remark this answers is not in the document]
zhangyue-hashdata: good idea, i will fix it.

src/backend/executor/nodeHash.c
zhangyue-hashdata: fix with https://github.com/apache/cloudberry/pull/724/commits/99eabb224f89dcdfd7117ef37a5986ab8001abdf

src/backend/executor/nodeHashjoin.c
zhangyue-hashdata: fix with https://github.com/apache/cloudberry/pull/724/commits/99eabb224f89dcdfd7117ef37a5986ab8001abdf

src/backend/executor/nodeSeqscan.c
zhangyue-hashdata: fix with https://github.com/apache/cloudberry/pull/724/commits/99eabb224f89dcdfd7117ef37a5986ab8001abdf

avamingli:
There are codes changed  in MultiExecParallelHash, please add some parallel tests with runtime filter.

zhangyue-hashdata:
> There are codes changed in MultiExecParallelHash, please add some parallel tests with runtime filter.

got it.

src/backend/executor/nodeHashjoin.c
zhangyue-hashdata: support partitioned table in https://github.com/apache/cloudberry/pull/724/commits/c70109269c71e616734a1083159eaf44d6362e48

src/test/regress/expected/gp_runtime_filter.out
zhangyue-hashdata: Add debug message in https://github.com/apache/cloudberry/pull/724/commits/274d8aa1ff0b33a4d24e3a7765b493904c61dd60

src/backend/executor/nodeSeqscan.c
zhangyue-hashdata: **Similarities:**
In this feature, the creation and testing of the Bloom filter use the same implementation as that of the gp_enable_runtime_filter feature.

**Differences:**

- The gp_enable_runtime_filter feature decides during the optimizer phase whether to use a Bloom filter to filter data. It inserts the functionality of filtering data with a Bloom filter as an nodeRuntimeFilter node into the left subtree of a Hash Join.

- In contrast, this new feature does not insert an node. Instead, it directly pushes down the Bloom filter and other filters to the SeqScan or even the Access Method (AM), applying the filters closer to where the data is generated. This approach enhances filtering efficiency compared to gp_enable_runtime_filter.

By pushing the filters closer to the data source, this new method achieves more efficient data filtering before the data reaches higher-level operators, potentially reducing the amount of data that needs to be processed by subsequent operations in the query execution plan.

zhangyue-hashdata: Yes, when creating the Bloom filter, the system evaluates the estimated number of rows that this hash join will process and the amount of available memory during the execution plan generation. It determines whether using a Bloom filter for filtering data would be effective based on this evaluation. If it is assessed that the Bloom filter would not sufficiently enhance performance, then the Bloom filter will not be created.

src/backend/executor/nodeHashjoin.c
zhangyue-hashdata: I think that expressions like `t1.c1 = 5` should be pushed down by the optimizer to operators such as SeqScan for early processing. Therefore, this feature does not handle expressions of the form `t1.c1 = 5`.

zhangyue-hashdata:
> Looks interesting. And I have some questions to discuss.
> 
> * Beside the seqscan, can the runtime filter apply to other types of scan? such as the index scan.
> * Looks only when the `hashjoin` node and `seqscan` node run in the same process can use the runtime filter. Which means the tables should have same distributed policy on the join columns or one of the table is replicated.

* Beside the seqscan, can the runtime filter apply to other types of scan? such as the index scan.
> Theoretically, it is feasible to apply runtime filters to operators such as Index Scan. However, because Index Scan already reduces data volume by leveraging an optimized storage structure, the performance gains from applying runtime filters to Index Scan would likely be minimal. Thus, I think that applying runtime filters to Index Scan would not yield significant performance benefits.

> In subsequent work, when we discover that other scan operators can achieve notable performance improvements from pushdown runtime filters, we will support these operators. Our focus will be on operators where runtime filters can substantially decrease the amount of data processed early in the query execution, leading to more pronounced performance enhancements.

* Looks only when the `hashjoin` node and `seqscan` node run in the same process can use the runtime filter. Which means the tables should have same distributed policy on the join columns or one of the table is replicated.
> Yes, the current pushdown runtime filter only supports in-process pushdown, which means that the Hash Join and SeqScan need to be within the same process. The design and implementation of cross-process pushdown runtime filters are much more complex.

> This limitation arises because coordinating and sharing data structures like Bloom filters or other runtime filters across different processes involves additional challenges such as inter-process communication (IPC), synchronization, and ensuring consistency and efficiency of the filters across process boundaries. Addressing these issues requires a more sophisticated design that can handle the complexities of distributed computing environments.

src/backend/executor/nodeHashjoin.c
fanfuxiaoran: Sorry, I didn't make it clear.  I don't mean the predication on the var.  like the below sql
```
 EXPLAIN (ANALYZE, COSTS OFF, SUMMARY OFF, TIMING OFF)
SELECT t1.c3 FROM t1, t2 WHERE t1.c2 = (t2.c2 + 10);
                                        QUERY PLAN
-------------------------------------------------------------------------------------------
 Gather Motion 3:1  (slice1; segments: 3) (actual rows=0 loops=1)
   ->  Hash Join (actual rows=0 loops=1)
         Hash Cond: (t1.c2 = (t2.c2 + 10))
         Extra Text: (seg2)   Hash chain length 8.0 avg, 8 max, using 4 of 524288 buckets.
         ->  Seq Scan on t1 (actual rows=128 loops=1)
         ->  Hash (actual rows=32 loops=1)
               Buckets: 524288  Batches: 1  Memory Usage: 4098kB
               ->  Seq Scan on t2 (actual rows=32 loops=1)
(9 rows)
```
As `t2.c2 + 10` is not a `Var` but a `T_OpExpr` , the runtime filter cannot handle it.
Could we just iterate the expression tree and check if it only contains var and const ?

avamingli:
Hi, with gp_enable_runtime_filter_pushdown = on, execute SQL below will get a crash:

```sql
gpadmin=# show gp_enable_runtime_filter_pushdown;
 gp_enable_runtime_filter_pushdown
-----------------------------------
 on
(1 row)
```
```sql
CREATE TABLE test_tablesample (dist int, id int, name text) WITH (fillfactor=10) DISTRIBUTED BY (dist);
-- use fillfactor so we don't have to load too much data to get multiple pages

-- Changed the column length in order to match the expected results based on relation's blocksz
INSERT INTO test_tablesample SELECT 0, i, repeat(i::text, 875) FROM generate_series(0, 9) s(i) ORDER BY i;
INSERT INTO test_tablesample SELECT 3, i, repeat(i::text, 875) FROM generate_series(10, 19) s(i) ORDER BY i;
INSERT INTO test_tablesample SELECT 5, i, repeat(i::text, 875) FROM generate_series(20, 29) s(i) ORDER BY i;
EXPLAIN (COSTS OFF)
  SELECT id FROM test_tablesample TABLESAMPLE SYSTEM (50) REPEATABLE (2);
FATAL:  Unexpected internal error (assert.c:48)
DETAIL:  FailedAssertion("IsA(planstate, SeqScanState)", File: "explain.c", Line: 4154)
server closed the connection unexpectedly
        This probably means the server terminated abnormally
        before or while processing the request.
The connection to the server was lost. Attempting reset: Succeeded.
psql (14.4, server 14.4)
```

image

zhangyue-hashdata:
> ```sql
> ```sql
> gpadmin=# show gp_enable_runtime_filter_pushdown;
>  gp_enable_runtime_filter_pushdown
> -----------------------------------
>  on
> (1 row)
> ```
> 
> 
>     
>       
>     
> 
>       
>     
> 
>     
>   
> ```sql
> CREATE TABLE test_tablesample (dist int, id int, name text) WITH (fillfactor=10) DISTRIBUTED BY (dist);
> -- use fillfactor so we don't have to load too much data to get multiple pages
> 
> -- Changed the column length in order to match the expected results based on relation's blocksz
> INSERT INTO test_tablesample SELECT 0, i, repeat(i::text, 875) FROM generate_series(0, 9) s(i) ORDER BY i;
> INSERT INTO test_tablesample SELECT 3, i, repeat(i::text, 875) FROM generate_series(10, 19) s(i) ORDER BY i;
> INSERT INTO test_tablesample SELECT 5, i, repeat(i::text, 875) FROM generate_series(20, 29) s(i) ORDER BY i;
> EXPLAIN (COSTS OFF)
>   SELECT id FROM test_tablesample TABLESAMPLE SYSTEM (50) REPEATABLE (2);
> FATAL:  Unexpected internal error (assert.c:48)
> DETAIL:  FailedAssertion("IsA(planstate, SeqScanState)", File: "explain.c", Line: 4154)
> server closed the connection unexpectedly
>         This probably means the server terminated abnormally
>         before or while processing the request.
> The connection to the server was lost. Attempting reset: Succeeded.
> psql (14.4, server 14.4)
> ```
> ```

Thanks, I'll reproduce the issue and fix it.

fanfuxiaoran:
Thanks for your detailed explanation. 

> > Looks interesting. And I have some questions to discuss.
> > 
> > * Beside the seqscan, can the runtime filter apply to other types of scan? such as the index scan.
> > * Looks only when the `hashjoin` node and `seqscan` node run in the same process can use the runtime filter. Which means the tables should have same distributed policy on the join columns or one of the table is replicated.
> 
> * Beside the seqscan, can the runtime filter apply to other types of scan? such as the index scan.
> 
> > Theoretically, it is feasible to apply runtime filters to operators such as Index Scan. However, because Index Scan already reduces data volume by leveraging an optimized storage structure, the performance gains from applying runtime filters to Index Scan would likely be minimal. Thus, I think that applying runtime filters to Index Scan would not yield significant performance benefits.
> 
Make sense. When doing hashjoin, index scan or index only scan are often not used on probe node.

> > In subsequent work, when we discover that other scan operators can achieve notable performance improvements from pushdown runtime filters, we will support these operators. Our focus will be on operators where runtime filters can substantially decrease the amount of data processed early in the query execution, leading to more pronounced performance enhancements.
> 
> * Looks only when the `hashjoin` node and `seqscan` node run in the same process can use the runtime filter. Which means the tables should have same distributed policy on the join columns or one of the table is replicated.
> 
> > Yes, the current pushdown runtime filter only supports in-process pushdown, which means that the Hash Join and SeqScan need to be within the same process. The design and implementation of cross-process pushdown runtime filters are much more complex.
> 
> > This limitation arises because coordinating and sharing data structures like Bloom filters or other runtime filters across different processes involves additional challenges such as inter-process communication (IPC), synchronization, and ensuring consistency and efficiency of the filters across process boundaries. Addressing these issues requires a more sophisticated design that can handle the complexities of distributed computing environments.

Exactly, and if  there is any lock  used to solve the problem may even lead bad performance.

fanfuxiaoran:
```
 explain analyze
SELECT count(t1.c3) FROM t1, t3 WHERE t1.c1 = t3.c1 ;
                                                              QUERY PLAN

-----------------------------------------------------------------------------------------
----------------------------------------------
 Finalize Aggregate  (cost=1700.07..1700.08 rows=1 width=8) (actual time=32119.566..32119
.571 rows=1 loops=1)
   ->  Gather Motion 3:1  (slice1; segments: 3)  (cost=1700.02..1700.07 rows=3 width=8) (
actual time=30.967..32119.550 rows=3 loops=1)
         ->  Partial Aggregate  (cost=1700.02..1700.03 rows=1 width=8) (actual time=32119
.131..32119.135 rows=1 loops=1)
               ->  Hash Join  (cost=771.01..1616.68 rows=33334 width=4) (actual time=14.0
59..32116.962 rows=33462 loops=1)
                     Hash Cond: (t3.c1 = t1.c1)
                     Extra Text: (seg0)   Hash chain length 1.0 avg, 3 max, using 32439 o
f 524288 buckets.
                     ->  Seq Scan on t3  (cost=0.00..387.34 rows=33334 width=4) (actual t
ime=0.028..32089.490 rows=33462 loops=1)
                     ->  Hash  (cost=354.34..354.34 rows=33334 width=8) (actual time=13.2
57..13.259 rows=33462 loops=1)
                           Buckets: 524288  Batches: 1  Memory Usage: 5404kB
                           ->  Seq Scan on t1  (cost=0.00..354.34 rows=33334 width=8) (ac
tual time=0.180..4.877 rows=33462 loops=1)
 Planning Time: 0.227 ms
```
runtime_filter has been pushed down to t3 table seqscan, but 'explain analyze' doesn't print them out.

```
\d t1
                 Table "public.t1"
 Column |  Type   | Collation | Nullable | Default
--------+---------+-----------+----------+---------
 c1     | integer |           |          |
 c2     | integer |           |          |
 c3     | integer |           |          |
 c4     | integer |           |          |
 c5     | integer |           |          |
Checksum: t
Indexes:
    "t1_c2" btree (c2)
Distributed by: (c1)
```
```
 \d t3
                 Table "public.t3"
 Column |  Type   | Collation | Nullable | Default
--------+---------+-----------+----------+---------
 c1     | integer |           |          |
 c2     | integer |           |          |
 c3     | integer |           |          |
 c4     | integer |           |          |
 c5     | integer |           |          |
Distributed by: (c1)
```

zhangyue-hashdata:
> ```
>  explain analyze
> SELECT count(t1.c3) FROM t1, t3 WHERE t1.c1 = t3.c1 ;
>                                                               QUERY PLAN
> 
> -----------------------------------------------------------------------------------------
> ----------------------------------------------
>  Finalize Aggregate  (cost=1700.07..1700.08 rows=1 width=8) (actual time=32119.566..32119
> .571 rows=1 loops=1)
>    ->  Gather Motion 3:1  (slice1; segments: 3)  (cost=1700.02..1700.07 rows=3 width=8) (
> actual time=30.967..32119.550 rows=3 loops=1)
>          ->  Partial Aggregate  (cost=1700.02..1700.03 rows=1 width=8) (actual time=32119
> .131..32119.135 rows=1 loops=1)
>                ->  Hash Join  (cost=771.01..1616.68 rows=33334 width=4) (actual time=14.0
> 59..32116.962 rows=33462 loops=1)
>                      Hash Cond: (t3.c1 = t1.c1)
>                      Extra Text: (seg0)   Hash chain length 1.0 avg, 3 max, using 32439 o
> f 524288 buckets.
>                      ->  Seq Scan on t3  (cost=0.00..387.34 rows=33334 width=4) (actual t
> ime=0.028..32089.490 rows=33462 loops=1)
>                      ->  Hash  (cost=354.34..354.34 rows=33334 width=8) (actual time=13.2
> 57..13.259 rows=33462 loops=1)
>                            Buckets: 524288  Batches: 1  Memory Usage: 5404kB
>                            ->  Seq Scan on t1  (cost=0.00..354.34 rows=33334 width=8) (ac
> tual time=0.180..4.877 rows=33462 loops=1)
>  Planning Time: 0.227 ms
> ```
> 
> runtime_filter has been pushed down to t3 table seqscan, but 'explain analyze' doesn't print them out.
> 
> ```
> \d t1
>                  Table "public.t1"
>  Column |  Type   | Collation | Nullable | Default
> --------+---------+-----------+----------+---------
>  c1     | integer |           |          |
>  c2     | integer |           |          |
>  c3     | integer |           |          |
>  c4     | integer |           |          |
>  c5     | integer |           |          |
> Checksum: t
> Indexes:
>     "t1_c2" btree (c2)
> Distributed by: (c1)
> ```
> 
> ```
>  \d t3
>                  Table "public.t3"
>  Column |  Type   | Collation | Nullable | Default
> --------+---------+-----------+----------+---------
>  c1     | integer |           |          |
>  c2     | integer |           |          |
>  c3     | integer |           |          |
>  c4     | integer |           |          |
>  c5     | integer |           |          |
> Distributed by: (c1)
> ```

Thanks for your test case. Based on these, I rewrote code to ensure that debug info are always displayed even when the number of filtered rows is zero. And add the test case into gp_runtime_filter.sql too.
fix in https://github.com/apache/cloudberry/commit/98dac6dfc7d5e44e111aa16bdf5948d07ee2eb00

zhangyue-hashdata:
> Hi, with gp_enable_runtime_filter_pushdown = on, execute SQL below will get a crash:
> 
> ```sql
> gpadmin=# show gp_enable_runtime_filter_pushdown;
>  gp_enable_runtime_filter_pushdown
> -----------------------------------
>  on
> (1 row)
> ```
> 
> ```sql
> CREATE TABLE test_tablesample (dist int, id int, name text) WITH (fillfactor=10) DISTRIBUTED BY (dist);
> -- use fillfactor so we don't have to load too much data to get multiple pages
> 
> -- Changed the column length in order to match the expected results based on relation's blocksz
> INSERT INTO test_tablesample SELECT 0, i, repeat(i::text, 875) FROM generate_series(0, 9) s(i) ORDER BY i;
> INSERT INTO test_tablesample SELECT 3, i, repeat(i::text, 875) FROM generate_series(10, 19) s(i) ORDER BY i;
> INSERT INTO test_tablesample SELECT 5, i, repeat(i::text, 875) FROM generate_series(20, 29) s(i) ORDER BY i;
> EXPLAIN (COSTS OFF)
>   SELECT id FROM test_tablesample TABLESAMPLE SYSTEM (50) REPEATABLE (2);
> FATAL:  Unexpected internal error (assert.c:48)
> DETAIL:  FailedAssertion("IsA(planstate, SeqScanState)", File: "explain.c", Line: 4154)
> server closed the connection unexpectedly
>         This probably means the server terminated abnormally
>         before or while processing the request.
> The connection to the server was lost. Attempting reset: Succeeded.
> psql (14.4, server 14.4)
> ```
> 
> image

hanks for your test case. I fix it in https://github.com/apache/cloudberry/commit/98dac6dfc7d5e44e111aa16bdf5948d07ee2eb00 And add the test case into gp_runtime_filter.sql too.
fix in https://github.com/apache/cloudberry/commit/98dac6dfc7d5e44e111aa16bdf5948d07ee2eb00

src/backend/executor/nodeHashjoin.c
zhangyue-hashdata: Add more message to explain why just inner, right and semi join are allowed with runtime filter.
fix it in https://github.com/apache/cloudberry/commit/98dac6dfc7d5e44e111aa16bdf5948d07ee2eb00

Smyatkin-Maxim:
Hi @zhangyue-hashdata 
I see that previous runtime filter implementation relies on some cost model at try_runtime_filter(). Do I understand it correctly, that this PR does not do any cost evaluation?
Also for TPC-H/TPC-DS can you provide results for each query separately?

Asking mostly out of curiosity, I see here are quite a few reviewers here already :)

src/backend/executor/nodeSeqscan.c
fanfuxiaoran: > It determines whether using a Bloom filter for filtering data would be effective based on this evaluation

That makes sense, but where is related code, I just didn't see them in this pr. 
Does it compares the number of rows between the output of hashtable and data in the probe table? If the rows of the hashtable are far less than that of the probe table , then use the runtime filter?

zhangyue-hashdata:
> Hi @zhangyue-hashdata I see that previous runtime filter implementation relies on some cost model at try_runtime_filter(). Do I understand it correctly, that this PR does not do any cost evaluation? Also for TPC-H/TPC-DS can you provide results for each query separately?
> 
> Asking mostly out of curiosity, I see here are quite a few reviewers here already :)

Basically, you're correct. Because our goal is to filter out as much data as possible right at the point of data generation. However, this will lead to very complex evaluations, so we only made a simple estimation based on rows and work memory when creating the Bloom filter.
Furthermore, I have placed the detailed test results for TPC-DS 10s in PR description.

zhangyue-hashdata:
> There are codes changed in MultiExecParallelHash, please add some parallel tests with runtime filter.

fix it in https://github.com/apache/cloudberry/pull/724/commits/7ab040ae178e9bb616bd75e0488aa6a2293d4183

src/backend/executor/nodeSeqscan.c
```diff
@@ -87,8 +101,14 @@ SeqNext(SeqScanState *node)
 	/*
 	 * get the next tuple from the table
 	 */
-	if (table_scan_getnextslot(scandesc, direction, slot))
+	while (table_scan_getnextslot(scandesc, direction, slot))
+	{
+		if (node->filter_in_seqscan && node->filters &&
```
yjhjstz, line 106:
```c++
if (!node->filter_in_seqscan ||  !node->filters)
{
     if (table_scan_getnextslot(scandesc, direction, slot))
         return slot;
}
else 
{
     while (table_scan_getnextslot(scandesc, direction, slot))
     {
            .....
     }
}
```

this make origin path more efficient and readable ?

zhangyue-hashdata: good idea! I fix in https://github.com/apache/cloudberry/pull/724/commits/bcf93e66f72a35db0ab5694a92c59c768423b1ab

yjhjstz:
from tpcds 10s details table, there are some bad cases.

zhangyue-hashdata:
> from tpcds 10s details table, there are some bad cases.

**21,24,30,42,49,54,68-1,99**

I retested these SQL statements that exhibited performance regression, and the latest test results show no noticeable performance difference when toggling **gp_enable_runtime_filter_pushdown**. So, I speculate that the performance regression in these SQL statements might be associated with testing method. Previously I tested by running the entire suite of 99 TPC-DS queries with **gp_enable_runtime_filter_pushdown** enabled, and then again with it disabled.

Therefore, a more appropriate method would be to execute the same SQL statement multiple times with **gp_enable_runtime_filter_pushdown** both enabled and disabled, respectively, and then take the average of those runs for comparison. I will follow this testing method for retesting and observe if there's any performance regression.

zhangyue-hashdata:
> > from tpcds 10s details table, there are some bad cases.
> 
> **21,24,30,42,49,54,68-1,99**
> 
> I retested these SQL statements that exhibited performance regression, and the latest test results show no noticeable performance difference when toggling **gp_enable_runtime_filter_pushdown**. So, I speculate that the performance regression in these SQL statements might be associated with testing method. Previously I tested by running the entire suite of 99 TPC-DS queries with **gp_enable_runtime_filter_pushdown** enabled, and then again with it disabled.
> 
> Therefore, a more appropriate method would be to execute the same SQL statement multiple times with **gp_enable_runtime_filter_pushdown** both enabled and disabled, respectively, and then take the average of those runs for comparison. I will follow this testing method for retesting and observe if there's any performance regression.

I have retested the performance of tpcds 10s using the previously mentioned testing method. Please see the description part for the latest results.

yjhjstz:
> I have retested the performance of tpcds 10s using the previously mentioned testing method. Please see the description part for the latest results

cool, what about tpcds 100 sf ?

zhangyue-hashdata:
> > I have retested the performance of tpcds 10s using the previously mentioned testing method. Please see the description part for the latest results
> 
> cool, what about tpcds 100 sf ?

Please see the description part for the results of tpcds 100s.

src/backend/executor/nodeSeqscan.c
```diff
@@ -87,8 +100,22 @@ SeqNext(SeqScanState *node)
 	/*
 	 * get the next tuple from the table
 	 */
-	if (table_scan_getnextslot(scandesc, direction, slot))
-		return slot;
+	if (node->filter_in_seqscan && node->filters)
+	{
+		while (table_scan_getnextslot(scandesc, direction, slot))
+		{
+			if (!PassByBloomFilter(node, slot))
```
yjhjstz, line 107: Is bloom filter works on tpc-ds 1000G ?

zhangyue-hashdata: Based on the test results of TPC-DS 10G and 100G, I estimate that there is a high possibility of observing a similar performance improvement margin in the TPC-DS 1000G performance test as well.

yjhjstz: but bloom_create_aggresive maybe return NULL, buffer limit 1M ~2MB

src/backend/executor/nodeHashjoin.c
zhangyue-hashdata: I got it. This scenario should be supported at a later date. However, since this is a type of scenario that might require a significant amount of code to implement, I think it would be best to support it at an appropriate time in the future.

src/backend/executor/nodeSeqscan.c
```diff
@@ -244,12 +281,27 @@ void
 ExecReScanSeqScan(SeqScanState *node)
 {
 	TableScanDesc scan;
+	ScanKey keys;
 
 	scan = node->ss.ss_currentScanDesc;
 
+	/*
+	 * Clear all the pushdown scan keys.
```
yjhjstz, line 289: why clear all ?

src/backend/executor/nodeSeqscan.c
zhangyue-hashdata: > but bloom_create_aggresive maybe return NULL, buffer limit 1M ~2MB

There are two factors that will ensure the Bloom filter is created successfully with a high probability. The first is that Bloom filters are typically built on small tables. Even for tables as large as 1TB or 10TB, the data volume of TPC-DS small tables is generally not very large. The second factor is the number of segments; the average amount of data from small tables distributed across each segment will also not be excessive. Considering these two factors together, for a data volume of 10TB, it is highly likely that the creation of the Bloom filter will not fail.

src/backend/executor/nodeSeqscan.c
zhangyue-hashdata: I revert the fix with https://github.com/apache/cloudberry/pull/724/commits/05cfbdbc25f50f6ec164fb7229ee8140a27de84f

src/backend/executor/nodeHashjoin.c
```diff
@@ -2156,3 +2176,293 @@ ExecHashJoinInitializeWorker(HashJoinState *state,
...
+			continue;
+
+		if (lattno < 1 || rattno < 1)
+			continue;
+
+		targets = FindTargetNodes(hjstate, lattno, &lattno);
+		if (lattno == -1 || targets == NULL)
+			continue;
+
+		foreach(lc, targets)
```
Smyatkin-Maxim, line 2239: Is having the same listcell variable here intended?
I guess it should be working, but i`m not sure how it gonna behave if at some point we add a break statement into the inner loop. And I'm not sure if this code will continue working in future PG releases :)

Smyatkin-Maxim:
@zhangyue-hashdata have you tried benchmarks on the recent builds? The last time I ran TPC-H against cloudberry, 12/22 queries used legacy optimizer. I believe now, after cherry-picks from gpdb master, more queries should be using ORCA optimizer. I wonder if there is still the same benefit from runtime bloom filters.

Btw, can you guys share TPC-DS and TPC-H toolkit you're using to benchmark cloudberry?

zhangyue-hashdata:
> @zhangyue-hashdata have you tried benchmarks on the recent builds? The last time I ran TPC-H against cloudberry, 12/22 queries used legacy optimizer. I guess for TPC-DS it's even worse. But now, after cherry-picks from gpdb master, more queries should be using ORCA optimizer. I wonder if there is still the same benefit from runtime bloom filters.
> 
> Btw, can you guys share TPC-DS and TPC-H toolkit you're using to benchmark cloudberry?

not yet recently, here is toolkit for TPC-DS, hope helpful to you :-)
[toolkit.tar.gz](https://github.com/user-attachments/files/19459216/toolkit.tar.gz)

src/backend/executor/nodeHashjoin.c
zhangyue-hashdata: > Is having the same listcell variable here intended? I guess it should be working, but i`m not sure how it gonna behave if at some point we add a break statement into the inner loop. And I'm not sure if this code will continue working in future PG releases :)

At the commit **Support partition table**, I changed the return type of `FindTargetNodes()` from `PlanState *` to `List *` for partition table.

src/backend/executor/nodeSeqscan.c
zhangyue-hashdata: Currently, there are no tables in CBDB that support runtime filter pushdown to AM. So runtime filter pushdown to AM will not be supported temporarily, and filtering operations can be performed only in SeqScan.

src/backend/executor/nodeHash.c
zhangyue-hashdata: It may be a better choice to let the user of the runtime filter decide whether to merge. :-)

src/backend/executor/nodeHash.c
```diff
@@ -511,6 +523,9 @@ ExecEndHash(HashState *node)
 	 */
 	outerPlan = outerPlanState(node);
 	ExecEndNode(outerPlan);
+
+	if (gp_enable_runtime_filter_pushdown && node->filters)
```
gfphoenix78, line 527: Is the if condition necessary here? We should always free the filter if the filter is not null

src/backend/executor/nodeHashjoin.c
```diff
@@ -999,6 +1014,11 @@ ExecInitHashJoin(HashJoin *node, EState *estate, int eflags)
 		ExecInitRuntimeFilterFinish(rfstate, hstate->ps.plan->plan_rows);
 	}
 
+	if (Gp_role == GP_ROLE_EXECUTE
+		&& gp_enable_runtime_filter_pushdown
+		&& !estate->useMppParallelMode)
+		CreateRuntimeFilter(hjstate);
```
gfphoenix78, line 1020: Will we support runtime filter in singlenode or utility mode?

gfphoenix78: If yes, we should also add tests in src/test/singlenode_regress/

src/backend/executor/nodeHash.c
```diff
@@ -4125,3 +4143,142 @@ get_hash_mem(void)
...
+		scankeys = lappend(scankeys, sk);
+
+		/* append new runtime filters to target node */
+		SeqScanState *sss = castNode(SeqScanState, attr_filter->target);
+		sss->filters = list_concat(sss->filters, scankeys);
+	}
+}
+
+static void
+BuildRuntimeFilter(HashState *node, TupleTableSlot *slot)
```
gfphoenix78, line 4199: The function name is not very good to me. `BuildXXX` looks do something only one pass, but it actually runs for every tuples.

zhangyue-hashdata: `AddTupleValuesIntoRF` should be better, I fix in [3828a4d](https://github.com/apache/cloudberry/pull/724/commits/3828a4d743d77c0c39501b8419640065291c343a)

src/backend/executor/nodeHashjoin.c
zhangyue-hashdata: This support will be added to the TODO list.

src/backend/executor/nodeHash.c
zhangyue-hashdata: fix in [3828a4d](https://github.com/apache/cloudberry/pull/724/commits/3828a4d743d77c0c39501b8419640065291c343a)

gfphoenix78 (approved):
LGTM

yjhjstz (approved):
LGTM

yjhjstz (approved):

yjhjstz:
@zhangyue-hashdata squash commits into one commit .

PR #778, opened 2024-12-13
Enhance Dynamic Table auto-refresh task.

avamingli:
This commit implements changes to the auto-refresh task for Dynamic Tables, specifically:

1. Forbid users from altering the "AS" part of the ALTER TASK command. The SQL must always be a REFRESH command to maintain the integrity of the Dynamic Table design (an auto-refreshing materialized view). Other ALTER TASK subcommands remain allowed, enabling users to control other aspects as needed.
2. Modify the SQL from "REFRESH DYNAMIC TABLE xxx" to "REFRESH DYNAMIC TABLE xxx CONCURRENTLY" to allow refreshing the materialized view without locking out concurrent selects.

These changes aim to improve the usability and performance of Dynamic Tables in our system.

Fixes #ISSUE_Number

### Type of Change
- [ ] Documentation update

### Test Plan

### Impact

**Performance:**

**User-facing changes:**

**Dependencies:**

### Checklist

### CI Skip Instructions

---

### DIFF
--- a/src/test/regress/expected/dynamic_table.out
+++ b/src/test/regress/expected/dynamic_table.out
@@ -156,14 +156,14 @@ SELECT * FROM t1 WHERE a = 2;
 (2 rows)
 
 -- test DROP DYNAMIC TABLE
-SELECT schedule, command FROM pg_task WHERE jobname LIKE 'gp_dynamic_table_refresh%' AND command LIKE '%dt0';
+SELECT schedule, command FROM pg_task WHERE jobname LIKE 'gp_dynamic_table_refresh%' AND command LIKE '%dt0%';
  schedule  |                    command                     
 -----------+------------------------------------------------
  5 * * * * | REFRESH DYNAMIC TABLE dynamic_table_schema.dt0
 (1 row)
 
 DROP DYNAMIC TABLE dt0;
-SELECT schedule, command FROM pg_task WHERE jobname LIKE 'gp_dynamic_table_refresh%' AND command LIKE '%dt0';
+SELECT schedule, command FROM pg_task WHERE jobname LIKE 'gp_dynamic_table_refresh%' AND command LIKE '%dt0%';
  schedule | command 
 ----------+---------
 (0 rows)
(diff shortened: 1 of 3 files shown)


src/test/regress/expected/dynamic_table.out
```diff
@@ -156,24 +156,24 @@ SELECT * FROM t1 WHERE a = 2;
 (2 rows)
 
 -- test DROP DYNAMIC TABLE
-SELECT schedule, command FROM pg_task WHERE jobname LIKE 'gp_dynamic_table_refresh%' AND command LIKE '%dt0';
- schedule  |                    command                     
------------+------------------------------------------------
- 5 * * * * | REFRESH DYNAMIC TABLE dynamic_table_schema.dt0
+SELECT schedule, command FROM pg_task WHERE jobname LIKE 'gp_dynamic_table_refresh%' AND command LIKE '%dt0%';
```
yjhjstz, line 159: add a test case that: insert rows, wait pg_task refresh, then select * from dt0 ;

avamingli: Thank you for the suggested test case. I'm okay with that.

However, I would like to clarify that there are already cases tested the REFRESH commands in existing cases, and the pg_task executes the same command. If there were any issues with pg_task not executing the command as expected, it would indicate a bug within pg_task, not with the dynamic table functionality itself.

src/backend/commands/createas.c
```diff
@@ -1846,7 +1846,7 @@ create_dynamic_table_auto_refresh_task(ParseState *pstate, Relation DynamicTable
 	dtname = quote_qualified_identifier(get_namespace_name(RelationGetNamespace(DynamicTableRel)),
 											 RelationGetRelationName(DynamicTableRel));
 	resetStringInfo(&buf);
-	appendStringInfo(&buf, "REFRESH DYNAMIC TABLE %s", dtname);
+	appendStringInfo(&buf, "REFRESH DYNAMIC TABLE %s CONCURRENTLY", dtname);
```
yjhjstz, line 1849: normal matview support CONCURRENTLY ?

avamingli: Yes, normal materialized views do support the CONCURRENTLY option. According to the [PostgreSQL docs](https://www.postgresql.org/docs/14/sql-refreshmaterializedview.html):

> CONCURRENTLY allows the materialized view to be refreshed without locking out concurrent selects. While a standard refresh might block other connections reading from the view, using CONCURRENTLY can be faster, especially when a small number of rows are affected

yjhjstz: but gp need UniqueIndex? I checked `ExecRefreshMatView`.

avamingli: If there is no unique index, adding the CONCURRENTLY option poses no risk. It could improve performance if users decide to create a unique index later, so we should include this option.

avamingli: Oh, what if there is no unique index with CONCURRENTLY, I will have a check

avamingli: You'r right!
> HINT:  Create a unique index with no WHERE clause on one or more columns of the materialized view.

We should not add this options.

avamingli: Done.

src/test/regress/expected/dynamic_table.out
avamingli: Done.

yjhjstz (approved):
LGTM

my-ship-it (approved):

Issue #765, opened 2024-12-11
Issue with Core Dumps in Cloudberry Tests


avamingli:
### Apache Cloudberry version

Main

### What happened

Hi Team,

I have a concern regarding the Cloudberry tests. Even when all tests pass, a core dump is generated on the system if it's enabled. 

<img width="1788" alt="image" src="https://github.com/user-attachments/assets/5a4c46dc-e28c-4dc2-b219-b63132466d72">

Typically, a core dump indicates a crash, suggesting that something has gone wrong. However, since all tests are passing, it appears that this might be an expected behavior intended for testing purposes.
I believe this issue may stem from Greenplum, as I've noticed similar problems reported there for quite some time(That issue was created by me, but no response and closed as an expected one).

Core dump files can be quite large and consume significant disk space. I believe we should avoid generating them if possible. It would be beneficial to identify the specific case that triggers the core dump and explore alternative methods for testing such scenarios.

### How to reproduce

Run regression test.

### Operating System

Ubuntu.

### Are you willing to submit PR?

- [ ] Yes, I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/apache/cloudberry/blob/main/CODE_OF_CONDUCT.md).

yjhjstz:
which test case caused core ？

avamingli:
> which test case caused core ？

Unfortunately, I don't know which specific test case is causing the core dump at this point; identifying that is the key of addressing this issue.
The call stack indicates that the core dump is triggered by an auxiliary process, but it's unclear which test case activates that process

edespino:
Thanks for raising this concern about core dumps being generated during test execution. This is definitely something we should address to improve our test infrastructure.

## Proposed Solution

We could enhance the test framework by adding an optional core file existence check. Here's how it would work:

```yaml
- test: ic-resgroup-v2
  make_target: installcheck-resgroup-v2
  make_directory: --directory=src/test/isolation2
  num_primary_mirror_pairs: 3
  perform_core_file_existence_check: true
```

When `perform_core_file_existence_check` is enabled, the framework would:
1. Monitor for core file generation during test execution
2. Automatically fail tests if unexpected core files are detected
3. Capture stack trace information for debugging purposes
4. Include clear failure messages indicating why the test failed (core file found)

This approach would help us:
- Identify potential issues early in the development cycle
- Manage disk space more effectively by avoiding unnecessary core dumps
[...]

Issue #782, opened 2024-12-16
[Bug] Core Dump in `mirror_replay` Test Suite During Execution


edespino:
### Apache Cloudberry version

main branch

### What happened

The `mirror_replay` test suite is consistently generating a core dump during execution. This test is part of the `greenplum_schedule` running under the `ic-good-opt-off` (`make -c src/test/regress installcheck-good`) test matrix configuration. From the core dump's stack , the issue occurs specifically during the append-only segment file handling in the startup process.

**Environment**

Project: Apache Cloudberry
Test Suite: mirror_replay
Schedule: greenplum_schedule
Test Matrix Config: ic-good-opt-off
Build Type: Debug build with the following configuration:
```
--enable-debug
--enable-profiling
--enable-cassert
--enable-debug-extensions
```

**Stack Trace**
The core dump stack trace indicates the crash occurs during append-only segment file handling:

```
Thread 1 (Thread 0x7f9cf7a5ed00 (LWP 8442)):
#0  0x00007f9cf8f11a6c in __pthread_kill_implementation () from /lib64/libc.so.6
#1  0x00007f9cf8ec4686 in raise () from /lib64/libc.so.6
#2  0x00007f9cf8eae833 in abort () from /lib64/libc.so.6
#3  0x00007f9cf9ca28bf in errfinish (filename=<optimized out>, filename@entry=0x7f9cfa27ef7a "xlogutils.c", lineno=lineno@entry=103, funcname=funcname@entry=0x7f9cfa27f060 <__func__.5> "log_invalid_page") at elog.c:819
#4  0x00007f9cf97272d6 in log_invalid_page (present=false, blkno=1, forkno=MAIN_FORKNUM, node=...) at xlogutils.c:103
#5  XLogAOSegmentFile (rnode=..., segmentFileNum=1) at xlogutils.c:567
#6  0x00007f9cf9d590a6 in ao_truncate_replay (record=<optimized out>, record=<optimized out>) at cdbappendonlyxlog.c:177
#7  0x00007f9cf971b7e5 in StartupXLOG () at xlog.c:7824
#8  0x00007f9cf9a6d124 in StartupProcessMain () at startup.c:267
#9  0x00007f9cf9767e52 in AuxiliaryProcessMain (argc=<optimized out>, argc@entry=2, argv=<optimized out>, argv@entry=0x7ffd0b1cc490) at bootstrap.c:483
[...]

PR #784, opened 2024-12-16
Enhance Build Pipeline with Debug and Core Analysis Support

edespino:
Fixes: #765 

Adds comprehensive debug build support and automated core dump analysis to the Cloudberry build pipeline. Key features:

- Debug build capability with preserved symbols and debug-specific RPMs
- Automated core dump detection and analysis during test execution
- Core file correlation with test failures
- Enhanced test result reporting with core dump status
- Improved artifact management for debug builds

The changes enable better debugging of test failures and provide more detailed information about process crashes during testing.

I expect the `ic-good-opt-off` group to fail as the `mirror_replay` test suite is dumping a core. I have filed the following for that issue: Core Dump in mirror_replay Test Suite During Execution #782

After the checks have run and `ic-good-opt-off`  has failed as expected with core dump, I plan to enable the skip ci optionis PR to allow it to be committed after receiving approval.

Benign edit to re-trigger the workflow.

### DIFF
--- a/.github/workflows/build-cloudberry.yml
+++ b/.github/workflows/build-cloudberry.yml
@@ -41,8 +41,9 @@
 #
 # 2. **Build Job**:
 #    - Configures and builds Apache Cloudberry.
+#    - Supports debug build configuration via ENABLE_DEBUG flag.
 #    - Runs unit tests and verifies build artifacts.
-#    - Creates RPM packages, source tarballs, and logs.
+#    - Creates RPM packages (regular or debug), source tarballs, and logs.
 #    - **Key Artifacts**: RPM package, source tarball, build logs.
 #
 # 3. **RPM Install Test Job**:
@@ -54,10 +55,12 @@
 #    - Executes a test matrix to validate different scenarios.
 #    - Creates a demo cluster and runs installcheck tests.
 #    - Parses and reports test results, including failed and ignored tests.
+#    - Detects and analyzes any core dumps generated during tests.
 #    - **Key Features**:
 #      - Regression diffs are displayed if found, aiding quick debugging.
 #      - Both failed and ignored test names are logged and reported.
-#    - **Key Artifacts**: Test logs, regression files, test summaries.
+#      - Core dumps are analyzed using GDB for stack traces.
+#    - **Key Artifacts**: Test logs, regression files, test summaries, core analyses.
 #
 # 5. **Report Job**:
 #    - Aggregates job results into a final report.
@@ -71,8 +74,8 @@
 #   - CPU: Recommended 4+ cores.
 #
 # Triggers:
-# - Push to `build-devel` branch.
-# - Pull requests to `build-devel` branch.
+# - Push to `main` branch.
+# - Pull requests to `main` branch.
 # - Manual workflow dispatch.
 #
 # Container Images:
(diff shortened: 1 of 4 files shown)


edespino:
> If we do this with a skip CI tag, any later pull requests will fail and cannot be merged. I’m not optimistic that we can fix the core dump issue #782 in a short time.
> 
> How about we configure the system to allow the mirror_replay case for now? This way, we can separate the core dumps and report them independently. When someone eventually fixes the mirror_replay issue, we can update the configuration to set the expected core dump case to NULL, ensuring that nothing is blocked.

The best I can do is to disable (ignore) the `mirror_replay` test for now. Is this what you meant? I do not have a way to exclude the core checking for a specific test suite within a test schedule. Will disabling `mirror_replay` suffice for now?

avamingli:
> The best I can do is to disable (ignore) the `mirror_replay` test for now. Is this what you meant?

This is indeed a way to address the issue at the moment. However, it means we will miss out on a test case. 
On the positive side, it will allow us to promptly detect core dumps when they occur.

cc: @my-ship-it

edespino:
There are a couple of options:

1. I could place `mirror_replay` into it's own schedule, add a makefile target for it, enhance the core file check to allow a non-failure state at the matrix test group level. We could then unwind when it is fixed.
2. Instead of signaling a failure, I could simply flag core files (if found) as a warning and not a fatal scenario. I can add this as an annotation in the UI and a yellow warning symbol in the summary.

Personally,  I prefer option 1 over 2.  I do not know if developers tend to review the details of the workflow action only when there are CI check status failures. A core file warning condition could go unnoticed if the test case passes as is the case with `mirror_replay`.

edespino:
> There are a couple of options:
> 
> 1. I could place `mirror_replay` into it's own schedule, add a makefile target for it, enhance the core file check to allow a non-failure state at the matrix test group level. We could then unwind when it is fixed.
> 2. Instead of signaling a failure, I could simply flag core files (if found) as a warning and not a fatal scenario. I can add this as an annotation in the UI and a yellow warning symbol in the summary.
> 
> Personally, I prefer option 1 over 2. I do not know if developers tend to review the details of the workflow action only when there are CI check status failures. A core file warning condition could go unnoticed if the test case passes as is the case with `mirror_replay`.

@avamingli & @my-ship-it Any thoughts on which approach to take? I have other test suites I would like to enable that is blocked on getting this work completed.

my-ship-it:
> There are a couple of options:
> 
> 1. I could place `mirror_replay` into it's own schedule, add a makefile target for it, enhance the core file check to allow a non-failure state at the matrix test group level. We could then unwind when it is fixed.
> 2. Instead of signaling a failure, I could simply flag core files (if found) as a warning and not a fatal scenario. I can add this as an annotation in the UI and a yellow warning symbol in the summary.
> 
> Personally, I prefer option 1 over 2. I do not know if developers tend to review the details of the workflow action only when there are CI check status failures. A core file warning condition could go unnoticed if the test case passes as is the case with `mirror_replay`.

Hey Ed, thanks for making the enhancement.
I prefer option 1, in addition, it's better we can add a FIXME to remind us to fix this problem later.

edespino:
@avamingli & @my-ship-it Please check out the workflow. You will see a new test group `ic-fixme` which contains `mirror_replay`. It succeeds but the core does not cause the test to fail. You will see that warnings are available with the test summary. Let me know what you think.

my-ship-it (approved):

avamingli (approved):

PR #793, opened 2024-12-17
Cherry-pick: COPY FORM doesn't throw out ERROR: extra data after last expected column (7X) 

hyongtao-db:
Long log:
In pg-12, there is no function `InitCopyFromDispatchSplit()`,
so attr_count must be > 0. But in GP7, we may get `[NULL]`in 
QE when all fields are processed in the QD. So the condition 
checking should change from `attr_count > 0` to `attr_count >= 0`.
But before it, `attr_count = list_length(attnumlist);`,
we can ensure that `attr_count >= 0` consistently,
so delete this bug-related condition `attr_count > 0`.

Due to code refactoring in gp-14, the related code is moved 
from `copy.c` to `copyfrom.c`, so modify the related code in
`copyfrom.c`.


avamingli:
Critical fix cherry-pick! Your efforts are greatly appreciated.

yjhjstz (approved):
LGTM

my-ship-it (approved):

PR #863, opened 2025-01-13
FIX: cherry-pick missing icw test cases

jiaqizho:
Fixes #ISSUE_Number

### Type of Change
- [ ] Documentation update

### Test Plan

### Impact

**Performance:**

**User-facing changes:**

**Dependencies:**

### Checklist

### CI Skip Instructions

---

### DIFF
--- a/src/test/regress/output/external_table_optimizer.source
+++ b/src/test/regress/output/external_table_optimizer.source
@@ -1829,9 +1829,6 @@ NOTICE:  found 2 data formatting errors (2 or more input rows), rejected related
 ERROR:  segment reject limit reached, aborting operation
 DETAIL:  Last error was: invalid input syntax for type integer: "error_1", column i  (seg0 slice1 127.0.0.1:40000 pid=12226)
 CONTEXT:  External table exttab_udfs_2, line 7 of file://@hostname@@abs_srcdir@/data/exttab_more_errors.data, column i
-SQL statement "INSERT INTO exttab_udfs_insert_2
-  SELECT i, j from exttab_udfs_2"
-PL/pgSQL function exttab_udfs_func2() line 25 at SQL statement
 SELECT COUNT(*) > 0 FROM
 (
 SELECT * FROM gp_read_error_log('exttab_udfs_1')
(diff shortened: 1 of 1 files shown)


avamingli:
The external_table_optimizer.out file is generated at runtime by output files and should be included in the .gitignore file as part of the external table. It even shouldn’t be committed to the repository.

While I believe the optimizer file usually should not be generated, there may be some differences during cherry-picking that lead to the creation of external_table_optimizer.out. However, we should avoid changing this file, as it is auto-generated by output files.

The right approach would simply be to remove it.

jiaqizho:
> The external_table_optimizer.out file is generated at runtime by output files and should be included in the .gitignore file as part of the external table. It even shouldn’t be committed to the repository.
> 
> While I believe the optimizer file usually should not be generated, there may be some differences during cherry-picking that lead to the creation of external_table_optimizer.out. However, we should avoid changing this file, as it is auto-generated by output files.
> 
> The right approach would simply be to remove it.

I don't get it, there are no `external_table_optimizer.out`  in main branch. And current i changed `external_table_optimizer.source`.

yjhjstz (approved):

avamingli:
> > The external_table_optimizer.out file is generated at runtime by output files and should be included in the .gitignore file as part of the external table. It even shouldn’t be committed to the repository.
> > While I believe the optimizer file usually should not be generated, there may be some differences during cherry-picking that lead to the creation of external_table_optimizer.out. However, we should avoid changing this file, as it is auto-generated by output files.
> > The right approach would simply be to remove it.
> 
> I don't get it, there are no `external_table_optimizer.out` in main branch. And current i changed `external_table_optimizer.source`.

Sorry, I mean remove external_table_optimizer.source
origin commit https://github.com/greenplum-db/gpdb-archive/commit/8a9fc72abd54ff23ca44e26496cc6bbc8901c07a didn't introduce that, shall we copy one?

avamingli:
> > > The external_table_optimizer.out file is generated at runtime by output files and should be included in the .gitignore file as part of the external table. It even shouldn’t be committed to the repository.
> > > While I believe the optimizer file usually should not be generated, there may be some differences during cherry-picking that lead to the creation of external_table_optimizer.out. However, we should avoid changing this file, as it is auto-generated by output files.
> > > The right approach would simply be to remove it.
> > 
> > 
> > I don't get it, there are no `external_table_optimizer.out` in main branch. And current i changed `external_table_optimizer.source`.
> 
> Sorry, I mean remove external_table_optimizer.source origin commit [greenplum-db/gpdb-archive@8a9fc72](https://github.com/greenplum-db/gpdb-archive/commit/8a9fc72abd54ff23ca44e26496cc6bbc8901c07a) didn't introduce that, shall we copy one?

Get it, Orca generated different plan.

avamingli (approved):

PR #917, opened 2025-02-11
Lock leaf partitions for Insert Statement when GDD disabled.

avamingli:
Greenplum will lock relation on QD during parse-analyze stage and then dispatch the plan to QEs to execute. Note the dispatch is async. For insert statement on root, we cannot predict which leaf partitions will be inserted. Thus, if we never lock leaf partitions on QD, this leads to async dispatch to QEs, means, some of the session's QEs will execute insert and hold locks, some of the session's QEs might be blocked by other sessions. The problematical lock pattern in MPP database  is: without locks on QD and first time hold lock is on QEs.

An global deadlock case is:

First, create a partition table

```sql
create table rank (id int, year int)
distributed randomly
partition by range (year)
(start (2006) end (2009) every (1));
```

Second, in a new session (named `sess1`)

```sql
set gp_vmem_idle_resource_timeout = 1000000; -- make sure IDLE QEs not recycled
select pg_backend_pid(); -- find the QD's process id
select sess_id from pg_stat_activity where pid = pg_backend_pid(); -- this is to find the session id
select * from gp_dist_random('gp_id'); -- this is to create a gang
-- now run ps -ef | grep postgres | grep con<session id> we can get the QEs' pids

-- Use GDB to attach to one of this session's  QE
-- In GDB, set a break point at exec_mpp_query
-- (gdb) b exec_mpp_query
-- then continue
-- (gdb) c

-- now we run insert into this partition table by directly inserting into the root
insert into rank select i,i%3+2006 from generate_series(1, 1000)i;
-- the above insert will hang, because we set a break point on one of its QE
```

Third, start a new session (named `sess2`)

```sql
truncate rank_1_prt_2; -- truncate a leaf
-- the above truncate SQL will not be blocked on QD, because sess 1's insert do not hold locks for leafs on QD
-- so the above truncate will be dispatched to segments
-- on the segments that we do not set break point of sess1's QE, this truncation can execute, and hold locks on the
-- leaf relation; however, on the other segments, this truncation will be blocked by the insert statement on leaf locks
-- it will hang now
```

Finally, quit the gdb of session 1's QE, then that QE will continue and try to hold locks on leaf, and will be blocked by the truncate.

Thus deadlock happens:
* on one segment, sess1 wait for sess 2 (insert wait for the truncate)
* on the other segments, sess 2 wait for sess 1(truncate wait for insert)

See Issue https://github.com/greenplum-db/gpdb/issues/13652 for details.

This commit fixes the issue by locking all leaf partitions on QD during parse-analyze stage when GDD is disabled. If GDD is enabled, we can break global deadlock.

Fixes #ISSUE_Number

### Type of Change
- [ ] Documentation update

### Test Plan

### Impact

**Performance:**

**User-facing changes:**

**Dependencies:**

### Checklist

### CI Skip Instructions

---


my-ship-it (approved):
LGTM

PR #956, opened 2025-02-22
Pass relation Oid to smgr_AORelOpenSegFile.

reshke:
smgr_* Functions, including smgr_AORelOpenSegFile, were added to CDBD in order to support compute/storage separation solutions, optimized for append load (in few words, for https://github.com/yezzey-gp/yezzey to work). Yezzey uses current segment file relation OIDs in its internal logic. It is not necessary to pass OID each time we call smgr_AORelOpenSegFile, because extension can perform syscahe lookup. 

A ratio for this change is to save few CPU cycles where possible. 

### Type of Change
- [ ] Documentation update

### Test Plan

### Impact

**Performance:**

**User-facing changes:**

**Dependencies:**

### Checklist

### CI Skip Instructions

---


src/backend/storage/smgr/smgr.c
```diff
@@ -104,13 +104,18 @@ static f_smgr smgrsw[SMGR_MAX_ID + 1] = {
 	}
 };
 
+static File	AORelOpenSegFile(__attribute__((unused))Oid reloid, const char *filePath, int fileFlags)
```
my-ship-it, line 107: Why are we doing this, but oid is not used here? Is it because we replace the function pointer in the extension?

src/backend/cdb/cdbappendonlyxlog.c
```diff
@@ -158,7 +158,7 @@ ao_truncate_replay(XLogReaderState *record)
 	pfree(dbPath);
 	dbPath = NULL;
 
-	file = smgr->smgr_ao->smgr_AORelOpenSegFile(path, O_RDWR | PG_BINARY);
+	file = smgr->smgr_ao->smgr_AORelOpenSegFile(InvalidOid, path, O_RDWR | PG_BINARY);
```
my-ship-it, line 161: Like storage manager, should we pass RelFileNodeBackend and segment_filenum instead of oid and path here?

reshke: Your suggestion looks sane, lets pass RelFileNodeBackend/segment_filenum in places where thay are known. 
I suggest to add more params to smgr_AORelOpenSegFile, to keep interface simple (do not use different kinds of callback for one similar purpose)

src/backend/storage/smgr/smgr.c
reshke: Because we cannot assing PathNameOpenFile to smgr_AORelOpenSegFile now, signature is changed.
Screenshot 2025-02-25 at 12 23 11

reshke:
@my-ship-it hi! How do you like this version?

my-ship-it (approved):

PR #1027, opened 2025-03-31
Adjust the loading order of preload

MisterRaindrop:
Prioritize loading the process_shared_preload_libraries.h
header file, and then load the shared_preload_libraries from postgresql.conf.

Fixes #ISSUE_Number

### Type of Change
- [ ] Documentation update

### Test Plan

### Impact

**Performance:**

**User-facing changes:**

**Dependencies:**

### Checklist

### CI Skip Instructions

---

### DIFF
--- a/src/backend/utils/init/miscinit.c
+++ b/src/backend/utils/init/miscinit.c
@@ -1764,31 +1764,13 @@ static char*
 expand_shared_preload_libraries_string()
 {
 	List	   *elemlist = NIL;
+	List	   *pg_elemlist = NIL;
 	List	   *deduplicate_elemlist = NIL;
 	ListCell   *l;
 	char	   *rawstring;
 	char	   *libraries = shared_preload_libraries_string;
 
-	/* Need a modifiable copy of string */
-	rawstring = pstrdup(libraries);
-	if (libraries != NULL && libraries[0] != '\0')
-	{
-		/* Parse string into list of filename paths */
-		if (!SplitDirectoriesString(rawstring, ',', &elemlist))
-		{
-			/* syntax error in list */
-			list_free_deep(elemlist);
-			pfree(rawstring);
-			ereport(LOG,
-					(errcode(ERRCODE_SYNTAX_ERROR),
-					errmsg("invalid list syntax in parameter \"%s\"",
-							libraries)));
-			return NULL;
-		}
-
-	}
-
-	/* load expand libraries */
+	/* load process_shared_preload_libraries.h libraries */
 	int shared_preload_libraries_num = sizeof(process_shared_preload_libraries_array) / sizeof(char *);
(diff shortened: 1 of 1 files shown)


my-ship-it:
Hi @MisterRaindrop, thanks for the PR.
Why we change the order here. Could you please share the context.

MisterRaindrop:
@my-ship-it 
Normally, we load libraries (libs) in two places: one is in the header files, and the other is in the postgresql.conf file. In the past, our approach was to append the libraries that needed to be loaded from the header files to the shared_preload_libraries parameter in postgresql.conf. Since the libraries loaded from the header files are compiled into the kernel, we want them to have the highest priority. If some libraries need to be loaded immediately, it could cause conflicts. Therefore, we place the loading of libraries from the header files first, ensuring they are loaded during the compilation process before the libraries specified in the postgresql.conf configuration are loaded.

my-ship-it:
> @my-ship-it OK Normally, we load libraries (libs) in two places: one is in the header files, and the other is in the postgresql.conf file. In the past, our approach was to append the libraries that needed to be loaded from the header files to the shared_preload_libraries parameter in postgresql.conf.
> 
> Since the libraries loaded from the header files are compiled into the kernel, we want them to have the highest priority, If some libraries need to be loaded immediately, it could cause conflicts.
> 
> Therefore, we place the loading of libraries from the header files first, ensuring they are loaded during the compilation process before the libraries specified in the postgresql.conf configuration are loaded.

Can you give an example of what kind of problems we will encounter please? The loading order matters looks strange, let's see if there are other ways to solve it

yjhjstz:
How about give a hook to self adjust preload library order ?

MisterRaindrop:
Initially, we added the header file to load the lib because we wanted to load some specific libs. These libs remain unchanged regardless of user modifications.

If we switch to using hooks, these specific libs might fail to load, which could lead to significant issues.

MisterRaindrop:
> > Example:
> > Ensuring lib1 is Loaded First is Crucial
> > For instance, we have added lib1 in the header file, and in postgresql.conf, we have added lib2, lib3, and lib4.
> > Following the original approach, the loading order would be:
> > ```
> > lib2, lib3, lib4, lib1
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > If we require lib1 to be loaded first, this cannot be achieved because the order is lib2, lib3, lib4, lib1. In theory, the header file should have the highest loading priority.
> > After modification, the loading order will be: header file + postgresql.conf
> > ```
> > lib1, lib2, lib3, lib4
> > ```
> 
> So we require loading the modules closely related to the kernel in the Header File first, and then loading the modules in the configuration file?

Yes

yjhjstz (approved):

my-ship-it (approved):

PR #1052, opened 2025-04-17
ORCA: fallback the locking clause

### LINKED ISSUE #1031: [Bug] ORCA generates wrong plan for select-for-update statement.
### Apache Cloudberry version

_No response_

### What happened

postgres=# set optimizer to on;
SET
postgres=# explain select * from test for update;
                                  QUERY PLAN
------------------------------------------------------------------------------
 Gather Motion 2:1  (slice1; segments: 2)  (cost=0.00..431.00 rows=1 width=4)
   ->  Seq Scan on test  (cost=0.00..431.00 rows=1 width=4)
 Optimizer: GPORCA
(3 rows)

postgres=# set optimizer to off;
SET
postgres=# explain select * from test for update;
                                     QUERY PLAN
------------------------------------------------------------------------------------
 Gather Motion 2:1  (slice1; segments: 2)  (cost=0.00..2457.50 rows=96300 width=10)
   ->  LockRows  (cost=0.00..1013.00 rows=48150 width=10)
         ->  Seq Scan on test  (cost=0.00..531.50 rows=48150 width=10)
(4 rows)

when set optimizer to on, plan for select-for-update statement is not correct because LockRows is lacked.

_No response_

### How to reproduce

set optimizer to on;

explain select * from test for update;

### Operating System

centos

### Anything else

_No response_

### Code of Conduct

jiaqizho:
Fixes #1031

### What does this PR do?
If the current GDD is closed and there is a `forupdate` statement, ORCA and planner does not need to generate a `LockRows` operator, expect the table is foreign table. However, if the current GDD is enabled, ORCA will generate a plan without the `LockRows` operator, which is a wrong plan.

Supporting locking clauses in ORCA is a very difficult thing. Because ORCA does not fill `root->glob->paramExecTypes` in every operator, even if ORCA generates the `LockRows` operator, it may not working in the executor. For details, please refer to the EPQ(evalPlanQual) mechanism.

Also I don't think ORCA should consider supporting locking clause. ORCA is an optimizer that is more biased towards OLAP scenarios. Support the EPQ mechanism will make ORCA more complex and more difficult to expand.

### Type of Change
- [ ] Documentation update

### Test Plan

### Impact

**Performance:**

**User-facing changes:**

**Dependencies:**

### Checklist

### CI Skip Instructions

---

### DIFF
--- a/src/backend/gpopt/translate/CTranslatorDXLToPlStmt.cpp
+++ b/src/backend/gpopt/translate/CTranslatorDXLToPlStmt.cpp
@@ -4723,14 +4723,7 @@ CTranslatorDXLToPlStmt::TranslateDXLDml(
 		isSplit = true; // AO tables are always use split updates
 	}
 	
-	if (md_rel->IsPartitioned())
-	{
-		dml->forceTupleRouting = true;
-	}
-	else
-	{
-		dml->forceTupleRouting = false;
-	}
+	dml->forceTupleRouting = md_rel->IsPartitioned();
 
 	if (IMDRelation::EreldistrMasterOnly != md_rel->GetRelDistribution())
 	{
--- a/src/backend/gpopt/translate/CTranslatorQueryToDXL.cpp
+++ b/src/backend/gpopt/translate/CTranslatorQueryToDXL.cpp
@@ -627,6 +627,12 @@ CTranslatorQueryToDXL::TranslateSelectQueryToDXL()
 		}
 	}
 
+	// Locking clause is not supported.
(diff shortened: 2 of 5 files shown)


my-ship-it (approved):

HuSen8891 (approved):

jiaqizho:
`checkCanOptSelectLockingClause` won't let ORCA create locking clause.

In ORCA case, i guess that CBDB will lock the table when `enableLockOptimization` is on.

So closed current PR.

PR #1054, opened 2025-04-18
Replace std::shared_ptr<File> by std::unique_ptr<File>

gfphoenix78:
Shared smart pointer will introduce additional cost, and it also make the ownership complicated.
Use unique pointer to make the ownership explicitly visible.

### Test Plan

### Impact

**Performance:**

**User-facing changes:**

**Dependencies:**

### Checklist

### CI Skip Instructions

---

### DIFF
--- a/contrib/pax_storage/src/cpp/storage/orc/orc_test.cc
+++ b/contrib/pax_storage/src/cpp/storage/orc/orc_test.cc
@@ -144,7 +144,7 @@ TEST_F(OrcTest, WriteTuple) {
   ASSERT_NE(nullptr, local_fs);
 
   auto file_ptr = local_fs->Open(file_name_, fs::kWriteMode);
-  EXPECT_NE(nullptr, file_ptr);
+  EXPECT_NE(nullptr, file_ptr.get());
 
   OrcWriter::WriterOptions writer_options;
   writer_options.rel_tuple_desc = tuple_slot->tts_tupleDescriptor;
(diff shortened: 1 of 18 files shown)


contrib/pax_storage/src/cpp/storage/orc/orc_test.cc
```diff
@@ -192,16 +192,16 @@ TEST_F(OrcTest, WriteReadStripes) {
   auto local_fs = Singleton<LocalFileSystem>::GetInstance();
   ASSERT_NE(nullptr, local_fs);
 
-  std::shared_ptr<File> file_ptr = local_fs->Open(file_name_, fs::kWriteMode);
-  EXPECT_NE(nullptr, file_ptr);
+  std::unique_ptr<File> file_ptr = local_fs->Open(file_name_, fs::kWriteMode);
```
Smyatkin-Maxim, line 195: nit: you can use auto everywhere for consistency, I guess

Smyatkin-Maxim (approved):
lgtm

Smyatkin-Maxim:
A side question about PAX not related to this pr: as far as I can see PAX relies on C++ memory allocators. It doesn't use PG memory contexts, right? This may become problematic for production cases with high memory usage when overcommit is off.

gfphoenix78:
> A side question about PAX not related to this pr: as far as I can see PAX relies on C++ memory allocators. It doesn't use PG memory contexts, right? This may become problematic for production cases with high memory usage when overcommit is off.

Yes, the c++ code doesn't use PG memory context. Because we expect to scan multiple pax data file concurrently. High memory usage may throw an exception that will be caught by the wrapper code. When the wrapper catches an exception, it will raise a PG error, the c++ objects will be released in the clean stage.

PR #1079, opened 2025-04-29
Replace Pylint with Ruff for Python linting

edespino:
- Remove Pylint installation, configuration, and invocation from the Makefile.
- Remove the GPL-licensed tarball: gpMgmt/bin/pythonSrc/ext/pylint-0.21.0.tar.gz
- Remove the old Pylint configuration file (.rcfile) since it is no longer relevant.
- Introduce Ruff installation via the official installation script.
- Dynamically install Ruff instead of storing tarballs or source files.
- Update checkcode target to run Ruff instead of Pylint.
- Clean up references to Pylint tarballs in the clean/distclean target.
- Update LICENSE and add license/LICENSE-ruff.txt
- Add new pyproject.toml file to configure Ruff:

  [tool.ruff] line-length = 88

  [tool.ruff.lint] select = ["E", "F", "W"]

Note: Ruff installation requires internet access during the first run.

Note: The removal of the Pylint .rcfile requires eventual replacement with
      corresponding Ruff configuration enhancements if more complex rules
      are needed in the future.

This change only introduces Ruff into the workflow. It does not attempt to fix or address any issues that Ruff may report.


LICENSE
```diff
@@ -260,6 +260,9 @@ The PostgreSQL software includes:
       gpMgmt/bin/pythonSrc/ext/parse-1.8.2.tar.gz
       see licenses/LICENSE-parse.txt
 
+      gpMgmt/bin/pythonSrc/ext/ruff-0.11.5.tar.gz
+      see licenses/LICENSE-ruff.txt
+
```
tuhaihe, line 265: If we install `ruff` via the internet instead of bundling its tarball in our source code, no need to add the license info, just handle it like other dependencies, right?

https://infra.apache.org/licensing-howto.html#bundled-vs-non-bundled

edespino: @tuhaihe You are correct. Let me remove the license file updates from this PR.

gpMgmt/bin/pyproject.toml
```diff
@@ -0,0 +1,5 @@
+[tool.ruff]
+line-length = 88
```
tuhaihe, line 2:
```suggestion
# Maximum number of characters on a single line.
line-length = 120

# Indentation width (number of spaces).
indent-width = 4

[format]
# indent with spaces, rather than tabs.
indent-style = "space"
```

tuhaihe: I added some settings that are equal to the original `.rcfile` file. If keep your file unchanged, LGTM too. FYI.

tuhaihe (approved):

gpMgmt/bin/pyproject.toml
edespino: I'd like to leave it as is. We can leave these changes to anyone who picks up `ruff` support.

tuhaihe: Ok, NP.

my-ship-it (approved):

PR #1104, opened 2025-05-15
ao/co: fix an issue with COPY zero-column CO table

lss602726449:
The issue is that if the CO table once had column (and data), but later all columns are dropped, then COPY this table would hang:

```
create table co(a int) using ao_column;
insert into co select * from generate_series(1,10);
alter table co drop column a;

copy co to stdout; -- this hangs forever
```

The is because aocs_getnext is not handling this case well: it would open the segment files but skip the actually scanning (because no projected columns), write no tupleslot values, and finally return TRUE which doesn't make any sense as it did not read any tuple. So a loop to get next tuple would loop forever. If we return FALSE here, we would stop the loop early.

Note that, however, this does not make COPY's behavior for CO to be the same as heap or AO, where we would still copy out the same number of rows as in the table, just no data in the row. In order to get it right, we can make a targeted change to COPY TO so it leave the decision to the CO AM layer regarding which column to project, which is also more consistent with SELECT *.

The change to aocs_getnext() would serve as just defensive checks.

Also fixed an issue with aoco_beginscan_extractcolumns() where we shouldn't write proj[0] if natts==0, otherwise would overwrite invalid memory:
```
postgres=# create table co() using ao_column;
CREATE TABLE
postgres=# select * from co;
WARNING:  detected write past chunk end in ExecutorState 0x55f234ea87c8  (seg1 slice1 127.0.1.1:7003 pid=10726)
WARNING:  detected write past chunk end in ExecutorState 0x555b0a692708  (seg2 slice1 127.0.1.1:7004 pid=10727)
WARNING:  detected write past chunk end in ExecutorState 0x562e118bf768  (seg0 slice1 127.0.1.1:7002 pid=10725)
--
(0 rows)
```

Reviewed-by: Kevin Yeap <kyeap@vmware.com>

Fixes #ISSUE_Number

### Type of Change
- [ ] Documentation update

### Test Plan

### Impact

**Performance:**

**User-facing changes:**

**Dependencies:**

### Checklist

### CI Skip Instructions

---

### DIFF
--- a/src/test/regress/output/aocs.source
+++ b/src/test/regress/output/aocs.source
@@ -1403,3 +1403,22 @@ SELECT * FROM shorten_udt;
  (1,foo)
 (1 row)
 
+-- Test zero-column table
+create table t1()using ao_column;
+create index on t1 using btree ((1));
+insert into t1 default values;
+select * from t1;
+--
+(0 rows)
+
+delete from t1;
+select * from t1;
+--
+(0 rows)
+
+insert into t1 default values;
+select * from t1;
+--
+(0 rows)
+
+drop table t1;
--- a/src/backend/access/aocs/aocsam.c
+++ b/src/backend/access/aocs/aocsam.c
@@ -709,6 +709,13 @@ ReadNext:
 		/* If necessary, open next seg */
 		if (scan->cur_seg < 0 || err < 0)
 		{
+			/*
+			 * Bail out early if we do not have any column in the projection.
+			 * Placing here in order to have less impact on the hot path. 
+			 */
+			if (scan->columnScanInfo.num_proj_atts == 0)
+				return false;
+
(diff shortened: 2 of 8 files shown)


src/test/regress/output/aocs.source
```diff
@@ -1403,3 +1403,22 @@ SELECT * FROM shorten_udt;
  (1,foo)
 (1 row)
 
+-- Test zero-column table
+create table t1()using ao_column;
+create index on t1 using btree ((1));
+insert into t1 default values;
+select * from t1;
+--
+(0 rows)
```
gfphoenix78, line 1412: The test case is wrong.

gfphoenix78:
The PR still has some basic problems. I'll close this PR. Fill free to reopen it when the PR has fixed the test cases.

PR #1135, opened 2025-06-01
Support new `VERSION`/`BUILD_NUMBER` files for non-Git versioning fallback

edespino:
This change enhances the `getversion` utility to support Apache Cloudberry packaging workflows where Git metadata is not available. This is especially important for source tarballs and downstream builds that operate in Git-free environments.

**Key Improvements:**

* If the repository is not a Git repo, the script will now:

  * **(a)** Attempt to parse `VERSION` if it follows the legacy format:
    `<version> build <build_number>`.
  * **(b)** Fallback to reading the version string from a `VERSION` file and, if available, read the build number from a `BUILD_NUMBER` file.

* The script's header has been updated to document these fallback paths and clarify behavior for different build environments.

**Rationale:**

* Apache source tarballs, per ASF guidelines, cannot contain `.git` directories. This enhancement ensures `getversion` continues to produce semver-compatible version strings (e.g., `2.0.0-incubating-rc1 build 1`) even when Git is unavailable.
* The update ensures the same utility can be used both in CI/CD pipelines and user environments derived from official source distributions.

**Backward Compatibility:**

* The legacy `<version> build <build_number>` format remains supported.
* The Git-based logic is untouched and continues to operate as before.


gfphoenix78:
Nice work! 👍 I have a question: when is the VERSION file updated?

tuhaihe (approved):

edespino:
@gfphoenix78 

> Nice work! 👍 I have a question: when is the VERSION file updated?

Thanks! 👍

Great question. The VERSION file is used primarily in non-git repository scenarios, allowing users to override the version string without editing the generated `configure` script.

However, the Apache Cloudberry release process does not generate a `VERSION` file. Instead, we update `configure.ac` with the release version (e.g., `2.0.0-incubating-rc1`), regenerate `configure`, and commit these changes just before tagging the release.

We will be creating a `BUILD_NUMBER` file—defaulting to `1` — as part of the release process. This file is included in the source tarball to support reproducible builds and consistent versioning outside Git.

gfphoenix78 (approved):
LGTM

Issue #1101, opened 2025-05-14
[Bug] unstable case


avamingli:
### What happened

https://github.com/apache/cloudberry/actions/runs/15015096147/job/42192042707

```sql
0s
Run # Search for regression.diffs recursively
Found regression.diffs at: ./src/test/regress/regression.diffs
diff -I HINT: -I CONTEXT: -I GP_IGNORE: -U3 /__w/cloudberry/cloudberry/src/test/regress/expected/gp_explain.out /__w/cloudberry/cloudberry/src/test/regress/results/gp_explain.out
--- /__w/cloudberry/cloudberry/src/test/regress/expected/gp_explain.out	2025-05-[1](https://github.com/apache/cloudberry/actions/runs/15015096147/job/42192042707#step:18:1)4 01:06:12.981846295 -0700
+++ /__w/cloudberry/cloudberry/src/test/regress/results/gp_explain.out	2025-05-14 01:06:12.997846264 -0700
@@ -672,7 +671,7 @@
                      ->  WorkTable Scan on y (never executed)
                      ->  Materialize (never executed)
                            ->  Gather Motion 3:1  (slice1; segments: 3) (never executed)
-                                 ->  Seq Scan on recursive_table_ic (actual rows=4061 loops=1)
+                                 ->  Seq Scan on recursive_table_ic (actual rows=6090 loops=1)
 GP_IGNORE:(13 rows)
 
 DROP TABLE recursive_table_ic;
```
Found in PR https://github.com/apache/cloudberry/pull/1098, succeed if a retry.

It seems a catalog statistic issue, I haven't seem it before.

### How to reproduce

CI

### Operating System

CI

### Are you willing to submit PR?

- [ ] Yes, I am willing to submit a PR!

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://github.com/apache/cloudberry/blob/main/CODE_OF_CONDUCT.md).

avamingli:
Should be fixed by #1140

PR #1140, opened 2025-06-04
Fix unstable scan rows caused by limit squelch.

songdongxiaoa2:
Squelch in limit node may cause scan finish early, row count in scan may vary.

### What does this PR do?
Change gp_explain test case, ignore rows count to avoid case failure caused by unstable results.

### Type of Change

### Breaking Changes
no

### Test Plan

### Impact

**Performance:**

no

**User-facing changes:**

no

**Dependencies:**

no

---


avamingli:
@songdongxiaoa2   Good catch! 
This may fix #1101 Thanks.

avamingli (approved):
LGTM.

Issue #1097, opened 2025-05-14
[Bug] PAX: building error in Rocky8 because of unsuitable Protobuf version


tuhaihe:
### Apache Cloudberry version

main

### What happened

When trying to build the Cloudberry with enabling PAX, one error returned:

```
 $ make -j$(nproc) -C ~/cloudberry | tee ~/cloudberry/make-$(date "+%Y.%m.%d-%H.%M.%S").log
 
 
 -- Found Protobuf: /usr/lib64/libprotobuf.so;-lpthread (found version "3.5.0")
 CMake Error at /usr/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:230 (message):
   Could NOT find Protobuf: Found unsuitable version "3.5.0", but required is
   at least "3.6.1" (found /usr/lib64/libprotobuf.so;-lpthread)
 Call Stack (most recent call first):
   /usr/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:598 (_FPHSA_FAILURE_MESSAGE)
   /usr/share/cmake/Modules/FindProtobuf.cmake:650 (FIND_PACKAGE_HANDLE_STANDARD_ARGS)
   FindDependencies.cmake:3 (find_package)
   CMakeLists.txt:83 (include)


 -- Configuring incomplete, errors occurred!
 make[1]: *** [Makefile:52: build] Error 1
 make[1]: Leaving directory '/home/gpadmin/cloudberry/contrib/pax_storage'
 make: *** [GNUmakefile:32: all] Error 2
 make: Leaving directory '/home/gpadmin/cloudberry'
```

For the default protobuf version in Rockylinux 8:

```
[gpadmin@cdw ~]$ rpm -q  protobuf
protobuf-3.5.0-15.el8.x86_64
```

### What you think should happen instead

Can remove the protobuf version requirement or use the default protobuf version in Rocky Linux 8 as the minimal requirement.

### How to reproduce

```
[...]

PR #1151, opened 2025-06-10
Add configure-time check for protobuf when --enable-pax is specified

### LINKED ISSUE #1148: [Bug] --enable_pax fails to detect missing protobuf development libraries
### Apache Cloudberry version

When building Apache Cloudberry with --enable_pax, the configure step does not check for required protobuf development environment, leading to an opaque build failure downstream.

Expected:
configure should fail early and clearly if protobuf headers are missing.

### What happened

Build failure

_No response_

### How to reproduce

Execute configure with --enable-pax in an environment which does not have protobuf dev environment

### Operating System

Linux

### Anything else

_No response_

### Code of Conduct

yjhjstz:
When building with --enable-pax, the build now checks for the presence of protobuf (version >= 3.5.0) using pkg-config. If protobuf is not found, configure will fail with an appropriate error message. This ensures that missing dependencies are caught early in the build process.

Fixes #1148 

### Type of Change
- [ ] Documentation update

### Test Plan

### Impact

**Performance:**

**User-facing changes:**

**Dependencies:**

### Checklist

### CI Skip Instructions

---

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -31,6 +31,7 @@ dnl your responsibility whether the result works or not.])])
 AC_COPYRIGHT([Copyright (c) 1996-2021, PostgreSQL Global Development Group])
 AC_CONFIG_SRCDIR([src/backend/access/common/heaptuple.c])
 AC_CONFIG_AUX_DIR(config)
+m4_include([config/ax_compare_version.m4])
 AC_PREFIX_DEFAULT(/usr/local/cbdb)
 AC_SUBST(configure_args, [$ac_configure_args])
 AC_DEFINE_UNQUOTED(CONFIGURE_ARGS, ["$ac_configure_args"], [Saved arguments from configure])
@@ -932,7 +933,40 @@ PGAC_ARG_BOOL(enable, pax, no,
               [enable PAX support],
               [AC_DEFINE(USE_PAX_STORAGE, 1,
                          [Define to 1 to support PAX])])
-AC_MSG_RESULT([checking whether to build with PAX support ... $enable_pax])
+if test "$enable_pax" = yes; then
+  # Check protobuf >= 3.5.0
+  PKG_CHECK_MODULES([PROTOBUF], [protobuf >= 3.5.0],
+    [AC_DEFINE([HAVE_PROTOBUF], [1], [Define if protobuf is available])],
+    [AC_MSG_ERROR([protobuf >= 3.5.0 is required for PAX support])]
+  )
+
+  # Check libzstd >= 1.4.0
+  PKG_CHECK_MODULES([ZSTD], [libzstd >= 1.4.0],
+    [AC_DEFINE([HAVE_ZSTD], [1], [Define if zstd is available])],
+    [AC_MSG_ERROR([libzstd >= 1.4.0 is required for PAX support])]
+  )
+
+  # Check cmake >= 3.11.0 using AX_COMPARE_VERSION
+  AC_PATH_PROG([CMAKE], [cmake], [no])
+  if test "$CMAKE" = "no"; then
+    AC_MSG_ERROR([cmake is required for PAX support but was not found in PATH])
+  fi
+
+  AC_MSG_CHECKING([CMake version])
+  CMAKE_VERSION=`$CMAKE --version | head -n1 | sed 's/[^0-9]*\([0-9][0-9.]*\).*/\1/'`
+  AC_MSG_RESULT([$CMAKE_VERSION])
+
+  AX_COMPARE_VERSION([$CMAKE_VERSION], [ge], [3.11.0],
+    [],
+    [AC_MSG_ERROR([cmake >= 3.11.0 is required for PAX support])]
+  )
+
+  # All checks passed
+  AC_MSG_RESULT([checking whether to build with PAX support ... yes])
+else
+  # Explicitly disabled
+  AC_MSG_RESULT([checking whether to build with PAX support ... no])
+fi
 AC_SUBST(enable_pax)
 
 #
--- a/config/ax_compare_version.m4
+++ b/config/ax_compare_version.m4
@@ -1,0 +1,177 @@
+# ===========================================================================
+#    https://www.gnu.org/software/autoconf-archive/ax_compare_version.html
+# ===========================================================================
+#
+# SYNOPSIS
+#
+#   AX_COMPARE_VERSION(VERSION_A, OP, VERSION_B, [ACTION-IF-TRUE], [ACTION-IF-FALSE])
+#
+# DESCRIPTION
+#
+#   This macro compares two version strings. Due to the various number of
+#   minor-version numbers that can exist, and the fact that string
+#   comparisons are not compatible with numeric comparisons, this is not
+#   necessarily trivial to do in a autoconf script. This macro makes doing
+#   these comparisons easy.
+#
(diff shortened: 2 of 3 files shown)


configure.ac
tuhaihe, line 1: We may also need to check cmake >= 3.11, and gcc/g++ >=11, libzstd >= 1.4.0, other than protobuf. They are also required for the PAX build.

yjhjstz: `checking for libzstd >= 1.4.0... yes ` already have.

yjhjstz: pax only need c++14, `checking whether g++ supports C++14 features with -std=c++14... yes` already have.

yjhjstz: cmake >=3.11 I will do in next pr, step by step.

tuhaihe: > `checking for libzstd >= 1.4.0... yes ` already have.

Yes, but it's for `--with-zstd` check. I think there is a chance of running `configure` with `--without-zstd`. Therefore, we need to ensure a double-check for pax.

tuhaihe: > pax only need c++14, `checking whether g++ supports C++14 features with -std=c++14... yes` already have.

Can we add the g++ version check here? You know, the default g++ version is `8.5.0 20210514` under Rocky Linux 8. 

Through my test, it shows `checking whether g++ supports C++14 features with -std=c++14... yes` though g++ 8.5.0 is installed, but will face failure using g++ 8.5.0. We need to upgrade it to 11+. 

Related issue - https://github.com/apache/cloudberry/issues/1097#issuecomment-2883361723.

edespino (dismissed):
The check should not be verbose when the --enable-pax checks for the dependencies (example: protobuf) pass. It should only be verbose when the dependencies are not met.

The following:

```
checking whether to build with PAX support ... yes
configure: Checking for protobuf >= 3.5.0 for PAX support...
checking for protobuf >= 3.5.0... yes
configure: protobuf found
```

Should only have:
```
checking whether to build with PAX support ... yes
```

configure.ac
yjhjstz: > > pax only need c++14, `checking whether g++ supports C++14 features with -std=c++14... yes` already have.
> 
> Can we add the g++ version check here? You know, the default g++ version is `8.5.0 20210514` under Rocky Linux 8.
> 
> Through my test, it shows `checking whether g++ supports C++14 features with -std=c++14... yes` though g++ 8.5.0 is installed, but will face failure using g++ 8.5.0. We need to upgrade it to 11+.
> 
> Related issue - [#1097 (comment)](https://github.com/apache/cloudberry/issues/1097#issuecomment-2883361723).

but I test use gcc-9 , it works ok.

yjhjstz:
> The check should not be verbose when the --enable-pax checks for the dependencies (example: protobuf) pass. It should only be verbose when the dependencies are not met.
> 
> The following:
> 
> ```
> checking whether to build with PAX support ... yes
> configure: Checking for protobuf >= 3.5.0 for PAX support...
> checking for protobuf >= 3.5.0... yes
> configure: protobuf found
> ```
> 
> Should only have:
> 
> ```
> checking whether to build with PAX support ... yes
> ```

simplify，done.

configure.ac
yjhjstz: > > `checking for libzstd >= 1.4.0... yes ` already have.
> 
> Yes, but it's for `--with-zstd` check. I think there is a chance of running `configure` with `--without-zstd`. Therefore, we need to ensure a double-check for pax.

libzstd check done.

tuhaihe: > but I test use gcc-9 , it works ok.


With a lower GCC version, there may be numerous warnings during compilation and installation. Have you experienced this? I will test it later.

yjhjstz: > > but I test use gcc-9 , it works ok.
> 
> With a lower GCC version, there may be numerous warnings during compilation and installation. Have you experienced this? I will test it later.

no experienced. But limit gcc version >= 11  is not good idea. We suggest test c++ features like `-std=c++14`.

tuhaihe: > > > but I test use gcc-9 , it works ok.
> > 
> > 
> > With a lower GCC version, there may be numerous warnings during compilation and installation. Have you experienced this? I will test it later.
> 
> no experienced. But limit gcc version >= 11 is not good idea. We suggest test c++ features like `-std=c++14`.

Cool. I tested with g++ 9.2.1 and the build was successful without too many warnings. With g++ 8.5.0, it will still fail to build the PAX-related modules (googlebench).

So for the g++ version check, we can ask gcc 9+ version installed at least. How does it make sense?

configure.ac
```diff
@@ -932,7 +933,40 @@ PGAC_ARG_BOOL(enable, pax, no,
...
+  fi
+
+  AC_MSG_CHECKING([CMake version])
+  CMAKE_VERSION=`$CMAKE --version | head -n1 | sed 's/[^0-9]*\([0-9][0-9.]*\).*/\1/'`
+  AC_MSG_RESULT([$CMAKE_VERSION])
+
+  AX_COMPARE_VERSION([$CMAKE_VERSION], [ge], [3.11.0],
+    [],
+    [AC_MSG_ERROR([cmake >= 3.11.0 is required for PAX support])]
+  )
```
tuhaihe, line 962: I just raised the idea that we can only verify whether protobuf, cmake, or libzstd are installed. The default versions of those libraries under Rocky8 all meet the minimum version requirement, so we can simplify the version check for them here.

Only the g++ version requires at least version 9+ or it will fail to build the PAX-related modules. @yjhjstz

configure.ac
yjhjstz: Better not do so, I would like to  change BUILD_GBENCH to off default in next pr.

configure.ac
tuhaihe: ~~Another idea is that if we can specify the googlbench submodule version that can be built with the g++ 8 version. Then we can also skip the g++ version check.~~

yjhjstz: Better not do so, I would like to change BUILD_GBENCH to off default in next pr.

configure.ac
tuhaihe: > Better not do so, I would like to change BUILD_GBENCH to off default in next pr.

Good. Thanks!

configure.ac
tuhaihe: > Better not do so, I would like to change BUILD_GBENCH to off default in next pr.

Good. thanks!

tuhaihe:
Hey @yjhjstz, Just curious if there's any chance we could implement the function ourselves, instead of introducing the new file `config/ax_compare_version.m4`?

yjhjstz:
> Hey @yjhjstz, Just curious if there's any chance we could implement the function ourselves, instead of introducing the new file `config/ax_compare_version.m4`?

you can see config/*.m4,  for example `config/ax_cxx_compile_stdcxx.m4`,  from `https://www.gnu.org/software/autoconf-archive/ax_cxx_compile_stdcxx.html`, it's common practice.

tuhaihe:
Updates: https://github.com/apache/cloudberry/pull/1165 - this PR will support building Cloudberry with PAX directly under g++ 8 under Rocky Linux 8. So, no need to compare the g++ version for PAX.

my-ship-it (approved):

yjhjstz (commented):
@edespino Could you please review the updated changes and approve if it looks good?

PR #1158, opened 2025-06-13
Update OIDs registered in the CBDB kernel to unused_oids

jiaqizho:
For the PAX extension, because some AM callback logic is hardcoded in the AO/AOCS table, we have to register 7047 as AM OID in CBDB, so that CBDB can checkout the table am is is PAX.

But `unused_oids` does not know this behavior, so the current commit adds an `excluded_oids` array, which is used to record OIDs that are only registered in the kernel but not registered in the pg_catalog.

Fixes #ISSUE_Number

### Type of Change
- [ ] Documentation update

### Test Plan

### Impact

**Performance:**

**User-facing changes:**

**Dependencies:**

### Checklist

### CI Skip Instructions

---

### DIFF
--- a/src/include/catalog/unused_oids
+++ b/src/include/catalog/unused_oids
@@ -28,9 +28,21 @@ chdir $FindBin::RealBin or die "could not cd to $FindBin::RealBin: $!\n";
 use lib "$FindBin::RealBin/../../backend/catalog/";
 use Catalog;
 
+# Some of OIDs register in CBDB kernel
+my @excluded_oids = (
+    # ex.
+    # 4, 5           # excluded single OID
+    # 10000..10050   # excluded range OIDs
+
+    7047             # PAX AM ID
+);
+
 my @input_files = (glob("pg_*.h"), glob("gp_*.h"));
 
 my $oids = Catalog::FindAllOidsFromHeaders(@input_files);
+
+# Push the excluded OIDs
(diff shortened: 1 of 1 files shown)


jiaqizho:
@gongxun0928

src/include/catalog/unused_oids
```diff
@@ -28,10 +28,22 @@ chdir $FindBin::RealBin or die "could not cd to $FindBin::RealBin: $!\n";
 use lib "$FindBin::RealBin/../../backend/catalog/";
 use Catalog;
 
+# Some of OIDs register in CBDB kernel
+my @excluded_oids = (
+    # ex.
+    # 4, 5           # excluded single OID
+    # 10000..10050   # excluded range OIDs
+
+    7047             # PAX AM ID
```
gfphoenix78, line 37: It's highly recommend to extract the oid value from header file `#define PAX_TABLE_AM_OID 7047` in contrib/pax_storage/src/cpp/comm/pax_rel.h.

gfphoenix78: The file `src/include/catalog/duplicate_oids` also hard-coded the oid of pax table am.

yjhjstz: we can impl function like `FindAllOidsFromHeaders`

PR #1163, opened 2025-06-16
Reject zero-column table creation with AO column storage

### LINKED ISSUE #1055: [Bug] AO column table can't process zero-column table
### Apache Cloudberry version

main branch, with cassert

### What happened

```sql
set default_table_access_method = ao_column;
create table t1();
create index on t1 using btree ((1)); -- assert fail

insert into t1 default values;
select * from t1;

delete from t1;
select * from t1;

insert into t1 default values;
select * from t1;
```

The above query will crash.

The above query should has the same behavior as heap tables.

### How to reproduce

```sql
set default_table_access_method = ao_column;
create table t1();
create index on t1 using btree ((1)); -- assert fail

insert into t1 default values;
select * from t1;

delete from t1;
select * from t1;

insert into t1 default values;
select * from t1;
```

### Operating System

no spec

### Anything else

_No response_

### Code of Conduct

yjhjstz:
Creating a table with AO columnar storage and no columns is not supported. This patch adds an explicit check in DefineRelation() to throw an error when such a case is encountered. This ensures users receive a clear error message rather than encountering undefined behavior.

Also adds a regression test under misc_jiras to verify the behavior.

Fixes #1055

### Type of Change
- [ ] Documentation update

### Test Plan

### Impact

**Performance:**

**User-facing changes:**

**Dependencies:**

### Checklist

### CI Skip Instructions

---

### DIFF
--- a/src/backend/commands/tablecmds.c
+++ b/src/backend/commands/tablecmds.c
@@ -1038,6 +1038,15 @@ DefineRelation(CreateStmt *stmt, char relkind, Oid ownerId,
 	/* Greenplum specific code */
 	if (list_length(schema) == 0)
 	{
+		/*
+		 * Issue #1055: AO column storage not supported for zero-column tables,
+		 * report error.
+		 */
+		if (accessMethodId == AO_COLUMN_TABLE_AM_OID)
+			ereport(ERROR,
+				(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+				 errmsg("AO column storage is not supported for zero-column tables")));
+
 		elogif(Gp_role == GP_ROLE_DISPATCH, LOG,
 			   "creating a table with no columns.");
 
@@ -6849,6 +6858,13 @@ ATRewriteTables(AlterTableStmt *parsetree, List **wqueue, LOCKMODE lockmode,
 			else
 				NewAccessMethod = OldHeap->rd_rel->relam;
 
+			if (OldHeap->rd_rel->relnatts == 0 && NewAccessMethod == AO_COLUMN_TABLE_AM_OID)
+			{
+				ereport(ERROR,
+						(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+						 errmsg("cannot rewrite empty table \"%s\" as an AO columnar table",
+								RelationGetRelationName(OldHeap))));
+			}
 			/*
 			 * Select persistence of transient table (same as original unless
 			 * user requested a change)
--- a/src/test/regress/sql/misc_jiras.sql
+++ b/src/test/regress/sql/misc_jiras.sql
@@ -70,3 +70,12 @@ from gp_dist_random('gp_distribution_policy') where localoid = 't2_17271'::regcl
 drop table t_17271;
 drop table t1_17271;
 drop table t2_17271;
+
+-- Github Issue 1055
+-- AO column storage not supported for zero-column tables
+create table t_1055() using ao_column;
+
+-- This should fail because the table is empty and AO column storage is not supported for zero-column
+create table t1_1055() using ao_row;
+alter table t1_1055 set access method ao_column;
+
--- a/src/test/regress/expected/misc_jiras.out
+++ b/src/test/regress/expected/misc_jiras.out
@@ -113,3 +113,13 @@ from gp_dist_random('gp_distribution_policy') where localoid = 't2_17271'::regcl
 drop table t_17271;
 drop table t1_17271;
 drop table t2_17271;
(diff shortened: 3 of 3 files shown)


my-ship-it (approved):

gfphoenix78:
I have two concerns about the issue:
1. What's the behavior if we drop the aocs table to zero column?
2. Is it possible to convert other type of table with zero column to aocs table?

yjhjstz:
> 1. table to zero column

can you provide sql ?

avamingli:
> > What's the behavior if we drop the aocs table to zero column?
> > Is it possible to convert other type of table with zero column to aocs table?
> 
> @gfphoenix78 can you provide sql ?

create a table with one column, then drop the column.

Same concern, create a table with zero column is valid in Postgres/GPDB.

avamingli:
> > What's the behavior if we drop the aocs table to zero column?
> > Is it possible to convert other type of table with zero column to aocs table?
> 
> @gfphoenix78 can you provide sql ?

create a table with one column, then drop the column.

Same concern, create a table with zero column is valid in Postgres/GPDB.

> > > What's the behavior if we drop the aocs table to zero column?
> > > Is it possible to convert other type of table with zero column to aocs table?
> > 
> > 
> > @gfphoenix78 can you provide sql ?
> 
> create a table with one column, then drop the column.
> 
> Same concern, create a table with zero column is valid in Postgres/GPDB.

Designed to be valid.
https://www.postgresql.org/docs/14/sql-createtable.html


> Zero-Column Tables
PostgreSQL allows a table of no columns to be created (for example, CREATE TABLE foo();). This is an extension from the SQL standard, which does not allow zero-column tables. Zero-column tables are not in themselves very useful, but disallowing them creates odd special cases for ALTER TABLE DROP COLUMN, so it seems cleaner to ignore this spec restriction.

yjhjstz:
> Is it possible to convert other type of table with zero column to aocs table?

[fec6f10](https://github.com/apache/cloudberry/pull/1163/commits/fec6f104a64fe7fad6ad0743118ca05df0ac66bd) try to reject , if agreed, then I will add related testcase.

gfphoenix78:
> > > What's the behavior if we drop the aocs table to zero column?
> > > Is it possible to convert other type of table with zero column to aocs table?
> > 
> > 
> > @gfphoenix78 can you provide sql ?
> 
> create a table with one column, then drop the column.
> 
> Same concern, create a table with zero column is valid in Postgres/GPDB.

Yes. The normal table with columns may drop columns one by one, until to have no column.

This PR only disallow aocs table to create a zero-column table, but we may still have zero-column table in other table storage. `ALTER TABLE ... SET ACCESS METHOD` may convert other table storage to aocs, which is also invalid.

src/backend/commands/tablecmds.c
```diff
@@ -1047,6 +1047,15 @@ DefineRelation(CreateStmt *stmt, char relkind, Oid ownerId,
...
 			policy = createRandomPartitionedPolicy(policy->numsegments);
+
+		/*
+		 * Issue #1055: AO column storage not supported for zero-column tables,
+		 * report error.
+		 */
+		if (accessMethodId == AO_COLUMN_TABLE_AM_OID)
+			ereport(ERROR,
+				(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+				 errmsg("AO column storage is not supported for zero-column tables")));
```
gfphoenix78: The error message should appear before the line 1041 to suppress the message "creating a table with no columns.".

src/test/regress/sql/misc_jiras.sql
```diff
@@ -70,3 +70,8 @@ from gp_dist_random('gp_distribution_policy') where localoid = 't2_17271'::regcl
 drop table t_17271;
 drop table t1_17271;
 drop table t2_17271;
+
+-- Github Issue 1055
+-- AO column storage not supported for zero-column tables
+set default_table_access_method = ao_column;
```
gfphoenix78: Please add more test cases here:
1. `alter table ... drop column ...` to drop columns one by one until to zero-column.
2. `alter table ... set access method` from other type of table storage to aocs.

yjhjstz: > drop columns one by one until to zero-column

wild to reject drop column to zero ?

src/backend/commands/tablecmds.c
yjhjstz: done

yjhjstz:
close due to `wild to reject drop column to zero `

PR #1173, opened 2025-06-19
Parallel DISTINCT plan of multi-stage.

avamingli:
Postgres UPSTREAM does not support parallel DISTINCT processing since DISTINCT across multiple workers cannot be guaranteed. In MPP databases, however, we can utilize Motion to redistribute tuples across multiple workers within a parallel query.

For a DISTINCT query like:
```sql
select distinct a from t_distinct_0;
```
we can create a parallel plan based on the underlying node's Parallel Scan on the table. The tuples are distributed randomly after the Parallel Scan, even when the distribution key matches the target expression.

The pre-distinct node uses Streaming HashAggregate or HashAggregate to deduplicate some tuples in parallel, which are then redistributed according to the DISTINCT expressions. Finally, a second-stage process handles the DISTINCT operation.
```sql
                         QUERY PLAN
------------------------------------------------------------
 Gather Motion 6:1  (slice1; segments: 6)
   ->  HashAggregate
         Group Key: a
         ->  Redistribute Motion 6:6  (slice2; segments: 6)
               Hash Key: a
               Hash Module: 3
               ->  Streaming HashAggregate
                     Group Key: a
                     ->  Parallel Seq Scan on t_distinct_0
 Optimizer: Postgres query optimizer
(10 rows)
```
Parallel Group Aggregation is also supported:
```sql
explain(costs off)
select distinct a, b from t_distinct_0;
                        QUERY PLAN
-----------------------------------------------------------
 GroupAggregate
   Group Key: a, b
   ->  Gather Motion 6:1  (slice1; segments: 6)
         Merge Key: a, b
         ->  GroupAggregate
               Group Key: a, b
               ->  Sort
                     Sort Key: a, b
                     ->  Parallel Seq Scan on t_distinct_0
 Optimizer: Postgres query optimizer
(10 rows)
```

## performance

For DISTINCT, the performance in parallel processing is closely tied to the tuples among each worker process, particularly in the case of streaming hash aggregates. 
Overall, the test cases indicate that a parallel execution plan with two workers achieves nearly half the runtime of a non-parallel plan.

The parallel plans(hashagg, streaming-hashagg, hashagg-groupagg) have 2 workers.

DISTINCT | 1 | 2 | 3 | avg
-- | -- | -- | -- | --
non-parallel | 8207.833 | 8416.01 | 8163.197 | 8262.346667
hashagg | 4550.331 | 4558.036 | 4219.485 | 4442.617333
[...]

PR #1178, opened 2025-06-20
perf: Improve hash agg performance in large-scale data situations

gongxun0928:
During the TPC-DS tests, we observed an issue where CloudberryDB (CBDB) performed worse than Greenplum (GPDB) when the query plan generated a multi-stage aggregation (e.g., TPC-DS query 04). The phenomenon showed that the deduplication effect of CBDB’s **Streaming Partial HashAggregate** was significantly worse. As a result, the **Finalize HashAggregate** operator in CBDB processed significantly more data compared to GPDB under the same dataset.

```
Example plan from CBDB:
Gather Motion 32:1  (slice1; segments: 32)  (cost=0.00..19988.81 rows=1800000 width=76)
      ->  Finalize HashAggregate  (cost=0.00..19663.99 rows=56250 width=81)
            Group Key: customer_gp.c_customer_id, customer_gp.c_first_name,
                  customer_gp.c_last_name, customer_gp.c_preferred_cust_flag,
                  customer_gp.c_birth_country, customer_gp.c_login,
                  customer_gp.c_email_address, date_dim_gp.d_year
            ->  Redistribute Motion 32:32  (slice2; segments: 32)  (cost=0.00..19603.35 rows=56250 width=81)
                  Hash Key: customer_gp.c_customer_id, customer_gp.c_first_name, customer_gp.c_last_name,
                        customer_gp.c_preferred_cust_flag, customer_gp.c_birth_country, customer_gp.c_login,
                        customer_gp.c_email_address, date_dim_gp.d_year
                  ->  Streaming Partial HashAggregate  (cost=0.00..19589.09 rows=56250 width=81)
                        Group Key: customer_gp.c_customer_id, customer_gp.c_first_name, customer_gp.c_last_name,
                              customer_gp.c_preferred_cust_flag, customer_gp.c_birth_country, customer_gp.c_login,
                              customer_gp.c_email_address, date_dim_gp.d_year
                        ->  Hash Join  (cost=0.00..12346.24 rows=6935137 width=95)
                              Hash Cond: (store_sales_gp.ss_customer_sk = customer_gp.c_customer_sk)
                              ...
                              ...
```
Upon further investigation, we found that the **`NumericAggState`** structure in CloudberryDB contained two additional fields compared to Greenplum:
```c
int64 pInfcount; /* count of +Inf values */
int64 nInfcount; /* count of -Inf values */
```
These fields were introduced in PostgreSQL 14 to support +/- Inf values for numeric types. Consequently, the size of the NumericAggState structure increased from 128 bytes to 144 bytes.

In the Streaming Partial HashAggregate, a NumericAggState structure is created for each grouping key to track statistics for numeric types. This results in CBDB allocating 16 more bytes per grouping key compared to GPDB. This additional memory allocation contributes to the observed performance difference.

To address this issue, we need to adjust the aggstate->hash_mem_limit Inspired by PostgreSQL’s handling of hash_mem_limit, we introduce a scaling factor called hash_mem_multiplier. Following the changes made in PostgreSQL 15, we should set the default value of hash_mem_multiplier  to 2.0 to ensure better memory utilization during hash-based aggregation operations. In actual TPC-DS tests, we found that under most of the same  statement_mem conditions, when hash_mem_multiplier=1.5 is set, the  performance of greenplum and cloudberrydb is similar.

https://www.postgresql.org/message-id/flat/CAH2-Wzndc_ROk6CY-bC6p9O53q974Y0Ey4WX8jcPbuTZYM4Q3A%40mail.gmail.com

Fixes #ISSUE_Number

### Type of Change
- [ ] Documentation update

### Test Plan

### Impact

**Performance:**

**User-facing changes:**

**Dependencies:**

### Checklist

### CI Skip Instructions

---

### DIFF
--- a/src/backend/executor/nodeAgg.c
+++ b/src/backend/executor/nodeAgg.c
@@ -268,6 +268,7 @@
 #include "utils/dynahash.h"
 #include "utils/expandeddatum.h"
 #include "utils/faultinjector.h"
+#include "utils/guc.h"
 #include "utils/logtape.h"
 #include "utils/lsyscache.h"
 #include "utils/memutils.h"
@@ -1870,6 +1871,8 @@ hash_agg_set_limits(AggState *aggstate, double hashentrysize, double input_group
 			strict_memlimit = operator_mem;
 	}
 
+	strict_memlimit = strict_memlimit * hash_mem_multiplier;
+
 	/* if not expected to spill, use all of work_mem */
 	if (input_groups * hashentrysize < strict_memlimit * 1024L)
 	{
@@ -2092,6 +2095,26 @@ hash_agg_update_metrics(AggState *aggstate, bool from_tape, int npartitions)
 			}
 		}
 	}
+
+#ifdef USE_ASSERT_CHECKING
+	if (unlikely(DEBUG1 >= log_min_messages) && aggstate->streaming)
+	{
+		for (int setno = 0; setno < aggstate->num_hashes; setno++)
+		{
+			AggStatePerHash perhash = &aggstate->perhash[setno];
+			if (perhash->chain_count > 0)
+			{
+				elog(DEBUG1, "hash_agg_update_metrics streaming: setno=%d, chain_length_total=%.1f, chain_length_max=%d, bucket_used=%lu, bucket_total=%lu, num_expansions=%u",
+					 setno,
+					 (double)perhash->chain_length_total / perhash->chain_count,
+					 perhash->chain_length_max,
+					 perhash->bucket_used,
+					 perhash->bucket_total,
+					 perhash->num_expansions);
+			}
+		}
+	}
+#endif
 }
 
 /*
@@ -2147,6 +2170,8 @@ hash_choose_num_partitions(AggState *aggstate, double input_groups, double hashe
 		if (operator_mem < strict_memlimit)
 			strict_memlimit = operator_mem;
 	}
+
+	strict_memlimit = strict_memlimit * hash_mem_multiplier;
 
 	/*
 	 * Avoid creating so many partitions that the memory requirements of the
@@ -2276,7 +2301,7 @@ lookup_hash_entries(AggState *aggstate)
 				initialize_hash_entry(aggstate, hashtable, entry);
 			pergroup[setno] = entry->additional;
 		}
-		else
+		else if (!aggstate->streaming)
 		{
 			HashAggSpill *spill = &aggstate->hash_spills[setno];
 			TupleTableSlot *slot = aggstate->tmpcontext->ecxt_outertuple;
--- a/src/test/singlenode_regress/expected/groupingsets.out
+++ b/src/test/singlenode_regress/expected/groupingsets.out
@@ -2069,14 +2069,13 @@ select g100, g10, sum(g::numeric), count(*), max(g::text)
 from gs_data_1 group by cube (g1000, g100,g10);
              QUERY PLAN              
 -------------------------------------
- GroupAggregate
+ MixedAggregate
+   Hash Key: g100
+   Hash Key: g100, g10
    Group Key: g1000, g100, g10
    Group Key: g1000, g100
    Group Key: g1000
    Group Key: ()
-   Sort Key: g100, g10
-     Group Key: g100, g10
-     Group Key: g100
    Sort Key: g10, g1000
      Group Key: g10, g1000
      Group Key: g10
@@ -2084,7 +2083,7 @@ from gs_data_1 group by cube (g1000, g100,g10);
          Sort Key: g1000, g100, g10
          ->  Seq Scan on gs_data_1
  Optimizer: Postgres query optimizer
-(15 rows)
+(14 rows)
 
 create table gs_hash_1 as
 select g100, g10, sum(g::numeric), count(*), max(g::text)
--- a/contrib/pax_storage/src/test/regress/expected/groupingsets.out
+++ b/contrib/pax_storage/src/test/regress/expected/groupingsets.out
@@ -1784,6 +1784,7 @@ select array(select row(v.a,s1.*) from (select two,four, count(*) from onek grou
 (2 rows)
 
 -- test the knapsack
+set hash_mem_multiplier = 1;
 set enable_indexscan = false;
 set work_mem = '64kB';
 explain (costs off)
@@ -1863,6 +1864,7 @@ explain (costs off)
  Optimizer: Postgres query optimizer
 (15 rows)
 
+reset hash_mem_multiplier;
 -- check collation-sensitive matching between grouping expressions
 -- (similar to a check for aggregates, but there are additional code
 -- paths for GROUPING, so check again here)
--- a/contrib/pax_storage/src/test/regress/expected/groupingsets_optimizer.out
+++ b/contrib/pax_storage/src/test/regress/expected/groupingsets_optimizer.out
@@ -1847,6 +1847,7 @@ select array(select row(v.a,s1.*) from (select two,four, count(*) from onek grou
 (2 rows)
 
 -- test the knapsack
+set hash_mem_multiplier = 1;
 set enable_indexscan = false;
 set work_mem = '64kB';
 explain (costs off)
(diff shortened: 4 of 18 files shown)


avamingli:
> The phenomenon showed that the deduplication effect of CBDB’s **Streaming Partial HashAggregate** was significantly worse. As a result, the **Finalize HashAggregate** operator in CBDB processed significantly more data compared to GPDB under the same dataset

Thank you for your insights!

The Streaming HashAggregate does not spill; it sends out tuples once the hash table is full. Its performance is closely linked to the distinct property of the data from subnodes within the in-memory hash range. However, evaluating this distinct property can be challenging.

While HashAggregate may require more space and disk I/O, it can deduplicate more tuples. Conversely, Streaming HashAggregate might incur higher network I/O, deduplicating fewer tuples but potentially speeding up slice execution upon the motion node. Ultimately, the effectiveness of either approach depends on factors like network, disk performance, and, crucially, the distinct property of subnode data, especially given the in-memory limit of the hash table.

I encountered a similar observation in #1173, where we discussed adding a GUC to control Streaming HashAggregate. This allows users to tailor the approach to their specific environments, which could be beneficial, especially in ORCA.

Anyway, your work looks great!

src/backend/executor/nodeAgg.c
```diff
@@ -2092,6 +2095,25 @@ hash_agg_update_metrics(AggState *aggstate, bool from_tape, int npartitions)
...
+				 setno, 
+				 (double)perhash->chain_length_total / perhash->chain_count,
+				 perhash->chain_length_max,
+				 perhash->bucket_used,
+					perhash->bucket_total,
+					perhash->num_expansions);
+			}
+		}
+	}
+
```
avamingli: 1. codes indent
2. Is it advisable to log messages in the executor? Could this introduce overhead during execution?

src/backend/executor/nodeAgg.c
```diff
@@ -2276,7 +2300,7 @@ lookup_hash_entries(AggState *aggstate)
 				initialize_hash_entry(aggstate, hashtable, entry);
 			pergroup[setno] = entry->additional;
 		}
-		else
+		else if (!aggstate->streaming)
```
avamingli, line 2304: Not sure, could we get here if using streaming  hashagg?

jianlirong:
Actually, I was thinking, in CBDB's Streaming Partial HashAggregate, is it really necessary for the following two fields to be int64? Maybe int32 is enough.

```
int64 pInfcount; /* count of +Inf values */
int64 nInfcount; /* count of -Inf values */
```

gongxun0928:
> Actually, I was thinking, in CBDB's Streaming Partial HashAggregate, is it really necessary for the following two fields to be int64? Maybe int32 is enough.
> 
> ```
> int64 pInfcount; /* count of +Inf values */
> int64 nInfcount; /* count of -Inf values */
> ```

Except for some special cases, it should be very rare for the sum of +Inf/-Inf values in an aggregate to exceed 2^32-1. If we do not consider compatibility with such special case, we could change the type from int64 to int32. Combined with a field reordering like the example below, the struct size can be reduced from 144 bytes to 128 bytes. This would keep the memory usage for numeric aggregation in line with Greenplum, while also improving hash aggregation efficiency.

origin: 144bytes
```
typedef struct NumericAggState
{
	bool		calcSumX2;		/* if true, calculate sumX2 */
	MemoryContext agg_context;	/* context we're calculating in */
	int64		N;				/* count of processed numbers */
	NumericSumAccum sumX;		/* sum of processed numbers */
	NumericSumAccum sumX2;		/* sum of squares of processed numbers */
	int			maxScale;		/* maximum scale seen so far */
	int64		maxScaleCount;	/* number of values seen with maximum scale */
	/* These counts are *not* included in N!  Use NA_TOTAL_COUNT() as needed */
	int64		NaNcount;		/* count of NaN values */
	int64		pInfcount;		/* count of +Inf values */
	int64		nInfcount;		/* count of -Inf values */
} NumericAggState;
```

reordering: 128bytes
```
typedef struct NumericAggState
{
	bool		calcSumX2;		/* if true, calculate sumX2 */
	int32		maxScale;		/* maximum scale seen so far */
	int64		maxScaleCount;	/* number of values seen with maximum scale */
	MemoryContext agg_context;	/* context we're calculating in */
	int64		N;				/* count of processed numbers */
	NumericSumAccum sumX;		/* sum of processed numbers */
	NumericSumAccum sumX2;		/* sum of squares of processed numbers */
	int64		NaNcount;		/* count of NaN values */
	int32		pInfcount;		/* count of +Inf values */
	int32		nInfcount;		/* count of -Inf values */
} NumericAggState;
```
I'll try to give a performance comparison of this change by next Monday. Then we'll decide how to proceed.

jianlirong:
What I want to express is that for Streaming Partial HashAggregate, the data structure NumericAggState for the intermediate state in the first stage aggregation doesn't need to be the same as the state data structure for the final stage aggregation. For the final stage aggregation's state data structure, it should indeed maintain consistency with PG, and the two fields mentioned above could indeed exceed the maximum value of int32. However, for Streaming Partial HashAggregate, we can do some special handling: when the above two fields reach the maximum value of int32, we can stream out the results of the Partial HashAggregate, and then reconstruct a new Partial HashAggregate result starting from zero. Considering that int32 should be very large, even in extreme cases where the number of tuples is very high, the overall execution performance won't actually decline, because in such cases, the cost that NumericAggState brings to motion is almost negligible compared to scanning and performing aggregation operations on more than 2^32 - 1 tuples.

jianlirong:
In other words, for the intermediate state data structure NumericAggState in the first stage aggregation, all the int64 fields inside can actually be changed to int32, not just only the above two.

gongxun0928:
> In other words, for the intermediate state data structure NumericAggState in the first stage aggregation, all the int64 fields inside can actually be changed to int32, not just only the above two.

This approach is challenging. For instance, both the Finalize HashAggregate and Partial HashAggregate phases rely on the same workflow—advance_aggregates → xxx → do_numeric_accum—to perform numeric accumulation and update the aggregate state. If we were to use different NumericAggState structures for different phases, we would need to refactor a large number of numeric-related functions to support a more compact state data structure.
Additionally, in finalize_aggregates, we would need to select different serialization functions depending on the type of aggsplit. 
All of these changes would result in significant differences between our codebase and upstream PostgreSQL, making future version upgrades more difficult to manage.

gongxun0928:
I took a look at the latest PostgreSQL code (PostgreSQL 18 beta). Although pInfcount and nInfcount in NumericAggState are defined as int64, they are actually used as boolean flags in practice. I'm considering starting a discussion upstream to propose changing these two fields to bool in order to reduce the memory footprint of the NumericAggState struct. If there is consensus, I’d like to drive this change upstream.

update：
The only scenario where we actually need to increment or decrement the count of infinities is during sliding window aggregation, where we need to reuse previous statistics. However, even in this case, the values of pInfcount and nInfcount will never exceed 2^32-1, so using a 32-bit integer should be sufficient.


finally:
In order to maintain compatibility with upstream, I finally chose to adjust the value of the hash_mem_multiplier variable

src/backend/executor/nodeAgg.c
gongxun0928: yeah. All Hash Agg Node have a calling path like this: 
ExecAgg->agg_fill_hash_table->lookup_hash_entries

src/test/singlenode_regress/expected/groupingsets.out
```diff
@@ -2069,22 +2069,21 @@ select g100, g10, sum(g::numeric), count(*), max(g::text)
 from gs_data_1 group by cube (g1000, g100,g10);
              QUERY PLAN              
 -------------------------------------
- GroupAggregate
```
jiaqizho, line 2072: i still not get, why `hash_mem_multiplier` effect the planner? i don't see the relative logical in planner which used  `hash_mem_multiplier`.

gongxun0928: used by get_hash_memory_limit(); 
In consider_groupingsets_paths, when the value of get_hash_memory_limit() increases, an AGG_MIXED processing path may be generated
https://github.com/apache/cloudberry/blob/4b9657f357c24a18023e47c23a2be029e4139106/src/backend/optimizer/plan/planner.c#L4651-L4663

jiaqizho (approved):
LGTM

avamingli (approved):

gfphoenix78 (approved):

PR #1192, opened 2025-06-26
ORCA: allow different strategy control the redistribute key below agg…

jiaqizho:
…regate

Fixes #ISSUE_Number

### Type of Change
- [ ] Documentation update

### Test Plan

### Impact

**Performance:**

**User-facing changes:**

**Dependencies:**

### Checklist

### CI Skip Instructions

---

### DIFF
--- a/src/backend/gporca/libgpopt/src/operators/CPhysicalAgg.cpp
+++ b/src/backend/gporca/libgpopt/src/operators/CPhysicalAgg.cpp
@@ -8,7 +8,6 @@
 //	@doc:
 //		Implementation of basic aggregate operator
 //---------------------------------------------------------------------------
-
 #include "gpopt/operators/CPhysicalAgg.h"
 
 #include "gpos/base.h"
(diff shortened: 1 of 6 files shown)


src/backend/utils/misc/guc_gp.c
```diff
@@ -4488,6 +4489,17 @@ struct config_int ConfigureNamesInt_gp[] =
...
 	},
 
+	{
+		{"optimizer_agg_pds_strategy", PGC_USERSET, DEVELOPER_OPTIONS,
+			gettext_noop("Set the strategy of agg required distribution."),
+			NULL,
+			GUC_NOT_IN_SAMPLE
+		},
+		&optimizer_agg_pds_strategy,
+		OPTIMIZER_AGG_PDS_ALL_KEY, OPTIMIZER_AGG_PDS_ALL_KEY, OPTIMIZER_AGG_PDS_MINIMAL_LEN_KEY,
```
leborchuk: Sorry, but if I understand correctly, the goal we are pursuing is to exclude long strings from hash calculations. For example, if we have 10 columns and 5th column is a long string, it would be better not to hash it, and hash all the remaining columns in order to achieve uniform distribution of data across segments. 

From this logic, perhaps we could add a GUC with the maximum data type size for hash to exclude all long strings. If all columns contain long strings, then let's select the one with the minimum size.

jiaqizho: good idea, i will add this logic in GUC `optimizer_agg_pds_strategy ` as one of strategy

src/backend/gporca/libgpopt/src/operators/CPhysicalAgg.cpp
```diff
@@ -293,11 +292,65 @@ CDistributionSpec *
 CPhysicalAgg::PdsMaximalHashed(CMemoryPool *mp, CColRefArray *colref_array)
 {
 	GPOS_ASSERT(nullptr != colref_array);
+	CColRefArray *pcraResHashs = nullptr;
+
+	if (GPOS_FTRACE(EopttraceAggRRSFirstKey)) {
+		pcraResHashs = GPOS_NEW(mp) CColRefArray(mp);
+		if (colref_array->Size() > 0) {
```
leborchuk, line 301: Here we could get the empty pcraResHashs if colref_array size is 0. Maybe check `GPOS_ASSERT_IMP(colref_array->Size() > 0);` at the beginning and not to check anymore?

src/backend/gporca/libgpopt/src/operators/CPhysicalAgg.cpp
```diff
@@ -293,11 +292,65 @@ CDistributionSpec *
...
+			const gpmd::IMDType *pmdtyp = pcr->RetrieveType();
+			LINT typlen = pmdtyp->IsFixedLength() ? pmdtyp->Length() : (gpos::lint_max - 1);
+
+			if (typlen < minimal_typlen) {
+				minimal_typlen = typlen;
+				minimal_typlen_ul = ul;
+			}
+		}
+
+		if (minimal_typlen_ul != -1) {
```
leborchuk, line 322: Hmmm, what if we do not check for minimal_typlen_ul != -1  here because it's impossible. And if  minimal_typlen_ul == -1 we will catch assert on line 346 but will not know why the pcraResHashs->Size()  is zero. Instead we could write GPOS_ASSERT(minimal_typlen_ul != -1);

src/backend/gporca/libgpopt/src/operators/CPhysicalAgg.cpp
jiaqizho: `colref_array->Size()` can be 0. I don't understand why we need to check if colref_array->Size() LT 0 ？

src/backend/gporca/libgpopt/src/operators/CPhysicalAgg.cpp
jiaqizho: `minimal_typlen_ul` will be `-1` if current `colref_array` is empty.

src/backend/gporca/libgpopt/src/operators/CPhysicalAgg.cpp
leborchuk: Uuupss, sorry, it's my mistake

GPOS_ASSERT_IMP fires only if enable_cassert enabled in configuration, which is false by default and in production environment. 

I'm used to the fact that in the rest of my projects assert also works in production builds.

src/backend/gporca/libgpopt/src/operators/CPhysicalAgg.cpp
leborchuk: Yep, the same as for the previos comment, not relevant, could be ignored

gongxun0928 (approved):

PR #1224, opened 2025-07-11
Fix check for database version inside pg_upgrade.

reshke:
Due to recent project renaming, gpupgrade now failing to upgarde greenplum 6 to cloudberry. The issue is with check-database-version logic. Fix is kinda straingforwand, the only issue is with `dbstring2` variable name, as i cannot came up with anything better.


avamingli (approved):
LGTM.

PR #1257, opened 2025-07-24
Fix large oid

avamingli:
Fixes #ISSUE_Number

### Type of Change
- [ ] Documentation update

### Test Plan

### Impact

**Performance:**

**User-facing changes:**

**Dependencies:**

### Checklist

### CI Skip Instructions

---


src/backend/executor/nodeModifyTable.c
```diff
@@ -99,8 +101,43 @@ static TupleTableSlot *ExecPrepareTupleRouting(ModifyTableState *mtstate,
...
+typedef struct ModifiedLeafRelidsData
+{
+	ModifiedLeafRelidsKey	key;
+} ModifiedLeafRelidsData;
+
+static uint32
+modified_leaf_hash(const void *key, Size keysize)
+{
+	Assert(keysize == sizeof(ModifiedLeafRelidsKey));
+	return DatumGetUInt32(hash_any((const unsigned char*) key,
```
gfphoenix78, line 120: It assumes there is no any padding bits in the key, or the padding bits are always the same. Please declare a static assert to ensure there is no padding bytes.

src/backend/executor/nodeModifyTable.c
```diff
@@ -99,8 +101,43 @@ static TupleTableSlot *ExecPrepareTupleRouting(ModifyTableState *mtstate,
...
+									keysize));
+}
+
+static int
+modified_leaf_compare(const void *key1, const void *key2, Size keysize)
+{
+	Assert(keysize == sizeof(ModifiedLeafRelidsKey));
+	ModifiedLeafRelidsKey *k1 = (ModifiedLeafRelidsKey*) key1;
+	ModifiedLeafRelidsKey *k2 = (ModifiedLeafRelidsKey*) key2;
```
gfphoenix78: Better to keep the `const` qualifier.

src/backend/executor/nodeModifyTable.c
```diff
@@ -3764,16 +3819,51 @@ static void
...
+				inserted = lappend_oid(inserted, r->key.relid);
+				break;
+			case CMD_UPDATE:
+				updated = lappend_oid(updated, r->key.relid);
+				break;
+			case CMD_DELETE:
+				deleted = lappend_oid(deleted, r->key.relid);
+				break;
+			default:
+				break;
```
gfphoenix78, line 3850: `Assert(0)`?

src/backend/executor/nodeModifyTable.c
avamingli: Postgres codes also use Assert, but either is OK,

avamingli: It reported an error when I use your approach, StaticAssertDecl:
> expression must have a constant value

I think that's why Postgres don't use a static assert in those.
The keysize is  passed as a param here, static assert doesn't apply.

src/backend/executor/nodeModifyTable.c
avamingli: Good point, done.

src/backend/executor/nodeModifyTable.c
avamingli: done.

src/backend/executor/nodeModifyTable.c
gfphoenix78: keysize is not used in static assert. static assert like

```c
StaticAssertDecl(sizeof(Oid) + sizeof(CmdType) == sizeof(ModifiedLeafRelidsKey));
```

to make sure there is no padding in the struct, or the hash/compare functions should take care of the it.

avamingli: Hmm… that seems unnecessary—no standard hash functions work that way. I don’t plan to spend more time on this; the current code is good enough as is. Most of these don’t even include assertions anyway.

avamingli: > to make sure there is no padding in the struct, or the hash/compare functions should take care of the it.

And, padding is a fundamental behavior to ensure proper memory alignment. You cannot guarantee its absence in a standard-compliant way,  nor an assertion could do.

gfphoenix78: > Hmm… that seems unnecessary—no standard hash functions work that way. I don’t plan to spend more time on this; the current code is good enough as is. Most of these don’t even include assertions anyway.

An alternative way is to call memset of the key, see https://github.com/postgres/postgres/blob/db6461b1c9aae122b90bb52430f06efb306b371a/src/backend/utils/cache/relfilenumbermap.c#L153

I'm sure the hash value is wrong if there is any possible uninitialized padding bit in the key. The current key structure looks has no padding, but we'd better have the compiler for this assumption.

avamingli: > An alternative way is to call memset of the key, see

Make sense, you'r right, fixed. MemSet(0) is necessary here, added it to each caller.

PR #1283, opened 2025-08-01
Replace pip3 download with curl for Python deps

tuhaihe:
Replace pip3 download commands with curl for downloading psutil,
PyYAML and PyGreSQL packages in gpMgmt/bin/Makefile.

This change unifies the download mechanism to use curl consistently
for all Python dependencies, avoiding potential pip3 download
compatibility issues across different environments.

Changes:
- Remove pip3 download commands for psutil, PyYAML and PyGreSQL
- Use curl to download packages directly from PyPI
- Maintain existing wheel and cython installation for PyYAML build
- Keep the same package versions and download locations

Fixes #ISSUE_Number

### Type of Change
- [ ] Documentation update

### Test Plan

### Impact

**Performance:**

**User-facing changes:**

**Dependencies:**

### Checklist

### CI Skip Instructions

---

### DIFF
--- a/configure
+++ b/configure
@@ -742,6 +742,9 @@ with_krb_srvnam
 krb_srvtab
 with_gssapi
 with_pythonsrc_ext
+PIP3
+CURL
+PYTHON3
 with_python
 with_perl
 with_tcl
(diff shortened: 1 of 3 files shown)


tuhaihe:
~We will have a failure during the build process if enabling the configure option `--with-pythonsrc-ext` without pre-installing the `python3-pip` package in Ubuntu. `python3-pip` will be installed as a dependency by Python in Rocky Linux, but will not be installed in Ubuntu Docker Image.~

~The following is how to fix it:~

- ~Pre-install `python3-pip` as a required dependency;~
- ~Check if the python3-pip is installed; if not, will install this package to resolve the failure.~

tuhaihe:
Updates: will add the curl, pip3 check in the configure.ac. 

Why?
- curl and pip3 are not default-installed in the Ubuntu 20.04/22.04 Docker image. So for a better user experience, we need to check these dependencies before `make install` the Cloudberry. When users run the configure with the option `--with-pythonsrc-ext`, it will check the dependencies earlier.

Why use curl instead of `pip3 download`?
- `pip3 download` has some compatibility issues between different Python versions (3.6 under Rocky8, 3.9 under Rocky9), especially for downloading pyyaml - pyaml 5.7 is a little old and has different behavior across Ubuntu and Rockylinux. So for a simple way, change the way of downloading these Python packages.

tuhaihe:
Already tested successfully on Ubuntu 20.04/22.04 + Rocky Linux 8/9.

tuhaihe:
> [!NOTE]
> This commit will also be cherry-picked into the branch `REL_2_STABLE`.

configure
```diff
@@ -9892,6 +9894,98 @@ fi
...
+  fi
+done
+  done
+IFS=$as_save_IFS
+
+  ;;
+esac
+fi
+PIP3=$ac_cv_path_PIP3
+if test -n "$PIP3"; then
```
jiaqizho, line 10019: also check python3 ? and do we required the version of python3?

tuhaihe: I found that when installing pip3, it also installs Python3 as a required dependency in both Ubuntu and Rocky Linux. (However, when installing only Python3, pip is not a required dependency in Ubuntu (but is also installed in Rocky Linux).)

If pip3 is installed, Python3 will also be installed in both Ubuntu and Rocky Linux. So, the check for pip3 is enough here?

tuhaihe: - Python 3.6.8 is the default version, Rocky Linux 8
- Python 3.8.10 is the default version, Ubuntu 20.4

For now, no specific Python3 version should be required, IMO.

tuhaihe: > also check python3 ? and do we required the version of python3?

Already add the check for python3.

Issue #1301, opened 2025-08-08
[Bug] Unnecessary Motion Nodes Added for Partitioned Tables Join with other tables


avamingli:
### Apache Cloudberry version

main

### What happened

While debug TPC-DS with partitioned tables, I found that when executing queries on partitioned tables, the planner sometimes generates redundant Motion nodes even when distribution keys are logically compatible. 

The issue leads to suboptimal query plans with avoidable data redistribution overhead.

### How to reproduce

```sql
CREATE TABLE t1 (id varchar(32), date date) DISTRIBUTED BY (id)
PARTITION BY RANGE (date)
(START (date '2016-01-01') INCLUSIVE END (date '2016-01-04') EXCLUSIVE EVERY (INTERVAL '1 day'));
CREATE TABLE t2 (id varchar(32)) DISTRIBUTED BY (id);
analyze t1;
analyze t2;
\d+ t1;
                                      Partitioned table "public.t1"
 Column |         Type          | Collation | Nullable | Default | Storage  | Stats target | Description
--------+-----------------------+-----------+----------+---------+----------+--------------+-------------
 id     | character varying(32) |           |          |         | extended |              |
 date   | date                  |           |          |         | plain    |              |
Partition key: RANGE (date)
Partitions: t1_1_prt_1 FOR VALUES FROM ('01-01-2016') TO ('01-02-2016'),
            t1_1_prt_2 FOR VALUES FROM ('01-02-2016') TO ('01-03-2016'),
            t1_1_prt_3 FOR VALUES FROM ('01-03-2016') TO ('01-04-2016')
Distributed by: (id)

\d+ t2;
                                            Table "public.t2"
 Column |         Type          | Collation | Nullable | Default | Storage  | Stats target | Description
--------+-----------------------+-----------+----------+---------+----------+--------------+-------------
 id     | character varying(32) |           |          |         | extended |              |
Distributed by: (id)

EXPLAIN(COSTS OFF) SELECT COUNT(*) FROM t1 JOIN t2 USING(id);
                           QUERY PLAN
----------------------------------------------------------------
 Aggregate
   ->  Gather Motion 3:1  (slice1; segments: 3)
         ->  Hash Join
               Hash Cond: ((t1.id)::text = (t2.id)::text)
               ->  Dynamic Seq Scan on t1
                     Number of partitions to scan: 3 (out of 3)
               ->  Hash
                     ->  Seq Scan on t2
 Optimizer: GPORCA
(9 rows)

[...]

PR #1305, opened 2025-08-11
Add option to avoid additional generated EquivalenceCalss for RelabelType in cdb_pull_up_eclass.

avamingli:
Fix issue https://github.com/apache/cloudberry/issues/1301

For partition tables, cdb_pull_up_eclass will call get_eclass_for_sort_expr to find an existing equivalence class it is a member of. It will fail and build a new single-member EquivalenceClass for it when there is RelabelType. That will cause cdbpath_match_preds_to_both_distkeys return false even the locus are compatible because the RestrictInfo's left_ec/right_ec and rel's distribution key's dk_eclasses are two different pointers.
And unnecessary Motions will be added in that case.
```sql
CREATE TABLE t1 (id varchar(32), date date) DISTRIBUTED BY (id) PARTITION BY RANGE (date)
(START (date '2016-01-01') INCLUSIVE END (date '2016-01-02') EXCLUSIVE EVERY (INTERVAL '1 day')); CREATE TABLE t2 (id varchar(32)) DISTRIBUTED BY (id); EXPLAIN(COSTS OFF) SELECT COUNT(*) FROM t1 JOIN t2 USING(id);
                            QUERY PLAN
------------------------------------------------------------------
 Aggregate
   ->  Gather Motion 3:1  (slice1; segments: 3)
         ->  Hash Join
               Hash Cond: ((t1.id)::text = (t2.id)::text)
                     Hash Key: t1.id
                     ->  Seq Scan on t1_1_prt_1 t1
               ->  Hash
                     ->  Seq Scan on t2
(10 rows)
```
cdbpullup_findEclassInTargetList() will ignore the presence or absence of a RelabelType node atop either expr in determining whether an EC member expr matches a targetlist expr. But get_eclass_for_sort_expr() didn't handle additional RelabelType and a new EC may be generated. We should not disturb get_eclass_for_sort_expr() as it's used at many places and UPSTREAM didn't add RelabelType to EC for that case.
If find a sub_distkeyexpr from cdbpullup_findEclassInTargetList and we know that there is RelabelType stripped inside, we don't need to try to fetch or generate new EC in get_eclass_for_sort_expr(). So, add options for cdb_pull_up_eclass to ignore RelabelType when needed.

Fixes #ISSUE_Number

### Type of Change
- [ ] Documentation update

### Test Plan

### Impact

**Performance:**

**User-facing changes:**

**Dependencies:**

### Checklist

### CI Skip Instructions

---


my-ship-it (approved):
LGTM

PR #1284, opened 2025-08-01
Fix row estimation for parallel subquery paths.

avamingli:
In CBDB, row estimation is determined by the relation's rows and cluster segments.
However, when there is a parallel subquery scan path, each worker will process fewer rows (divided by parallel_workers).

```sql
set enable_parallel = off;
explain SELECT e.name
FROM employees e
WHERE e.salary > (
    SELECT AVG(salary)
    FROM employees
    WHERE department_id = e.department_id);
                                                         QUERY PLAN                                                         
----------------------------------------------------------------------------------------------------------------------------
 Gather Motion 3:1  (slice1; segments: 3)  (cost=163.42..307.76 rows=3767 width=218)
   ->  Hash Join  (cost=163.42..257.54 rows=1256 width=218)
         Hash Cond: (e.department_id = "Expr_SUBQUERY".csq_c0)
         Join Filter: (e.salary > "Expr_SUBQUERY".csq_c1)
         ->  Seq Scan on employees e  (cost=0.00..71.67 rows=3767 width=254)
         ->  Hash  (cost=150.92..150.92 rows=1000 width=36)
               ->  Broadcast Motion 3:3  (slice2; segments: 3)  (cost=130.09..150.92 rows=1000 width=36)
                     ->  Subquery Scan on "Expr_SUBQUERY"  (cost=130.09..137.59 rows=333 width=36)
                           ->  Finalize HashAggregate  (cost=130.09..134.26 rows=333 width=36)
                                 Group Key: employees.department_id
                                 ->  Redistribute Motion 3:3  (slice3; segments: 3)  (cost=90.50..122.67 rows=990 width=36)
                                       Hash Key: employees.department_id
                                       ->  Partial HashAggregate  (cost=90.50..102.87 rows=990 width=36)
                                             Group Key: employees.department_id
                                             ->  Seq Scan on employees  (cost=0.00..71.67 rows=3767 width=36)
 Optimizer: Postgres query optimizer
(16 rows)
```
Subquery Scan on "Expr_SUBQUERY"  (cost=130.09..137.59 **rows=333** width=36)
While, a parallel Subquery Scan has the same rows though cost is less than that.

```sql
set enable_parallel = on;
set min_parallel_table_scan_size = 0;
explain SELECT e.name
FROM employees e
WHERE e.salary > (
    SELECT AVG(salary)
    FROM employees
    WHERE department_id = e.department_id);
                                                        QUERY PLAN                                                         
---------------------------------------------------------------------------------------------------------------------------
[...]

PR #1311, opened 2025-08-13
Give UNION ALL more opportunities for parallel plans in MPP.

avamingli:
Previously in CBDB, we had to disable parallel execution for UNION ALL queries containing Motion nodes due to a subtle but critical correctness issue.

The problem occurred when Parallel Append workers marked subnodes as completed, causing other workers to skip them. While normally harmless, this became critical in MPP databases where Motion nodes are ubiquitous. This limitation forced us to disable parallel plans for most UNION ALL queries involving distributed tables, missing significant optimization opportunities.

As a result, we fell back to serial execution:
```sql
explain(costs off) select b, count(*) from t1 group by b union all select b, count(*) from t2 group by b;
                            QUERY PLAN
------------------------------------------------------------------
   ->  Append
         ->  HashAggregate
               Group Key: t1.b
                     Hash Key: t1.b
                     ->  Seq Scan on t1
         ->  HashAggregate
               Group Key: t2.b
               ->  Redistribute Motion 3:3  (slice3; segments: 3)
                     Hash Key: t2.b
                     ->  Seq Scan on t2
(13 rows)
```
The commit makes plan parallel by first attempting a parallel-aware Append when it's safe to do so, but crucially, we now have a robust fallback path: when Motion hazards are detected, we switch to using a parallel-oblivious Append. This works because while Parallel Append might skip slices containing Motions, regular Append doesn't have this problem - it will reliably execute all subnodes regardless of whether they contain Motion nodes or not. Moreover, since CBDB's Motion nodes are designed to handle tuples individually, we don't need to worry about coordination between workers when processing these Motion nodes.

This approach unlocks powerful new optimization opportunities, as shown in this example where we can now execute the query with different levels of parallelism for each subplan (2 workers for t1 and 3 workers for t2):
```sql
explain(costs off) select b, count(*) from t1 group by b union all select b, count(*) from t2 group by b;
                            QUERY PLAN
------------------------------------------------------------------
 Gather Motion 9:1  (slice1; segments: 9)
   ->  Append
         ->  HashAggregate
               Group Key: t1.b
               ->  Redistribute Motion 6:9  (slice2; segments: 6)
                     Hash Key: t1.b
                     Hash Module: 3
                     ->  Parallel Seq Scan on t1
         ->  HashAggregate
               Group Key: t2.b
               ->  Redistribute Motion 9:9  (slice3; segments: 9)
                     Hash Key: t2.b
                     Hash Module: 3
                     ->  Parallel Seq Scan on t2
(15 rows)
```
This change represents a significant improvement in CBDB's query optimizer, allowing UNION ALL queries to benefit from parallel execution even when they contain Motion nodes, while maintaining correctness and supporting flexible parallelism configurations across different parts of the query. The optimization is particularly valuable for complex queries like TPC-DS tests where UNION ALL operations are common.

Fixes #ISSUE_Number

### Type of Change
- [ ] Documentation update

### Test Plan

### Impact

**Performance:**

**User-facing changes:**

**Dependencies:**

### Checklist

### CI Skip Instructions

---

### DIFF
--- a/src/backend/optimizer/path/allpaths.c
+++ b/src/backend/optimizer/path/allpaths.c
@@ -1839,8 +1839,38 @@ add_paths_to_append_rel(PlannerInfo *root, RelOptInfo *rel,
 			 * estimate.
 			 */
 			partial_rows = appendpath->path.rows;
-			/* Add the path if subpath has not Motion.*/
-			if (appendpath->path.parallel_safe && appendpath->path.motionHazard == false)
+
+			if (enable_parallel_append)
+			{
+				/* Add the path if subpath didn't encounter motion hazard.*/
+				if (appendpath->path.parallel_safe && (appendpath->path.motionHazard == false))
+					add_partial_path(rel, (Path *)appendpath);
+				else
+				{
+					/*
+					 * CBDB_PARALLEL:
+					 * When a parallel-aware Append is dropped due to motion hazard,
+					 * we attempt a second pass using parallel-oblivious Append.
+					 *
+					 * This approach is feasible in CBDB because:
+					 * 1. All Motions in a parallel plan handle tuples individually
+					 * 2. Parallel Append might miss executing slices containing Motions,
+					 *    whereas regular Append does not have this problem
+					 *
+					 * This behavior is conceptually similar to UPSTREAM's Append node
+					 * with partial paths implementation.
+					 */
+					appendpath = create_append_path(root, rel, NIL, partial_subpaths,
+													NIL, NULL, parallel_workers,
+													false /*enable_parallel_append*/,
+													-1);
+					partial_rows = appendpath->path.rows;
+
+					if (appendpath->path.parallel_safe)
+						add_partial_path(rel, (Path *)appendpath);
+				}
+			}
+			else if (appendpath->path.parallel_safe)
 				add_partial_path(rel, (Path *)appendpath);
 		}
 	}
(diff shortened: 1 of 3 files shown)


avamingli:
Case failed, based on https://github.com/apache/cloudberry/pull/1284 to be merged first.

src/backend/optimizer/path/allpaths.c
```diff
@@ -1839,8 +1839,38 @@ add_paths_to_append_rel(PlannerInfo *root, RelOptInfo *rel,
...
+					 *    whereas regular Append does not have this problem
+					 *
+					 * This behavior is conceptually similar to UPSTREAM's Append node
+					 * with partial paths implementation.
+					 */
+					appendpath = create_append_path(root, rel, NIL, partial_subpaths,
+													NIL, NULL, parallel_workers,
+													false /*enable_parallel_append*/,
+													-1);
+					partial_rows = appendpath->path.rows;
```
yjhjstz, line 1867: why update partial_rows again ?

avamingli: We try a parallel-oblivious approach for parallel append  when parallel-aware append is unsafe to use. 
The previous partial_rows calculations were designed for a parallel-aware append path. If the parallel-aware append path is deemed safe, we can directly add the path without adjustments. 
Otherwise, we need to correct the partial_rows values accordingly.

PR #1324, opened 2025-08-22
feat: support runtime filter push down to Table AM

gongxun0928:
when seq scan begins, check whether the scanflags of table am is set to determine whether the runtime filter is pushed down.

When the runtime filter is pushed down to pax am, pax am converts the min/max scankey in the runtime filter into PFTNode and performs min/max filtering.

```
CREATE TABLE t1(c1 int, c2 int, c3 int, c4 int, c5 int) using pax with(minmax_columns='c1,c2');
insert into t1 select i,i,i,i,i from generate_series(1,10000000) i;
analyze t1;

CREATE TABLE t2(c1 int, c2 int, c3 int, c4 int, c5 int) with (appendonly=true, orientation=column) distributed REPLICATED;
INSERT INTO t2 VALUES (1,1,1,1,1), (2,2,2,2,2), (3,3,3,3,3), (4,4,4,4,4);
INSERT INTO t2 select * FROM t2;
INSERT INTO t2 select * FROM t2;
INSERT INTO t2 select * FROM t2;

set gp_enable_runtime_filter_pushdown to off;
EXPLAIN (ANALYZE, COSTS OFF, SUMMARY OFF, TIMING OFF)
SELECT t1.c3 FROM t1, t2 WHERE t1.c2 = t2.c2;
                                    QUERY PLAN
----------------------------------------------------------------------------------
 Gather Motion 1:1  (slice1; segments: 1) (actual rows=32 loops=1)
   ->  Hash Join (actual rows=32 loops=1)
         Hash Cond: (t1.c2 = t2.c2)
         Extra Text: Hash chain length 8.0 avg, 8 max, using 4 of 524288 buckets.
         ->  Seq Scan on t1 (actual rows=10000000 loops=1)
         ->  Hash (actual rows=32 loops=1)
               Buckets: 524288  Batches: 1  Memory Usage: 4098kB
               ->  Seq Scan on t2 (actual rows=32 loops=1)
 Optimizer: GPORCA
(9 rows)

Time: 1576.548 ms (00:01.577)

gpadmin=# set gp_enable_runtime_filter_pushdown to on;
SET
Time: 0.362 ms
gpadmin=# EXPLAIN (ANALYZE, COSTS OFF, SUMMARY OFF, TIMING OFF)
SELECT t1.c3 FROM t1, t2 WHERE t1.c2 = t2.c2;
                                    QUERY PLAN
----------------------------------------------------------------------------------
 Gather Motion 1:1  (slice1; segments: 1) (actual rows=32 loops=1)
   ->  Hash Join (actual rows=32 loops=1)
         Hash Cond: (t1.c2 = t2.c2)
         Extra Text: Hash chain length 8.0 avg, 8 max, using 4 of 524288 buckets.
         ->  Seq Scan on t1 (actual rows=131072 loops=1)
         ->  Hash (actual rows=32 loops=1)
               Buckets: 524288  Batches: 1  Memory Usage: 4098kB
               ->  Seq Scan on t2 (actual rows=32 loops=1)
 Optimizer: GPORCA
(9 rows)

Time: 38.471 ms
```

Fixes #ISSUE_Number

### Type of Change
- [ ] Documentation update

### Test Plan

### Impact

**Performance:**

**User-facing changes:**

**Dependencies:**

### Checklist

### CI Skip Instructions

---

### DIFF
--- a/contrib/pax_storage/src/cpp/access/pax_access_handle.cc
+++ b/contrib/pax_storage/src/cpp/access/pax_access_handle.cc
@@ -453,6 +453,7 @@ uint32 PaxAccessMethod::ScanFlags(Relation relation) {
   flags |= SCAN_FORCE_BIG_WRITE_LOCK;
 #endif
 
+  flags |= SCAN_SUPPORT_RUNTIME_FILTER;
   return flags;
 }
 
--- a/contrib/pax_storage/src/cpp/access/pax_scanner.cc
+++ b/contrib/pax_storage/src/cpp/access/pax_scanner.cc
@@ -218,7 +218,7 @@ bool PaxScanDesc::BitmapNextTuple(struct TBMIterateResult *tbmres,
 }
 
 TableScanDesc PaxScanDesc::BeginScan(Relation relation, Snapshot snapshot,
-                                     int nkeys, struct ScanKeyData * /*key*/,
+                                     int nkeys, struct ScanKeyData *key,
                                      ParallelTableScanDesc pscan, uint32 flags,
                                      std::shared_ptr<PaxFilter> &&pax_filter,
                                      bool build_bitmap) {
@@ -326,8 +326,8 @@ void PaxScanDesc::EndScan() {
 }
 
 TableScanDesc PaxScanDesc::BeginScanExtractColumns(
-    Relation rel, Snapshot snapshot, int /*nkeys*/,
-    struct ScanKeyData * /*key*/, ParallelTableScanDesc parallel_scan,
+    Relation rel, Snapshot snapshot, int nkeys,
+    struct ScanKeyData *key, ParallelTableScanDesc parallel_scan,
     struct PlanState *ps, uint32 flags) {
   std::shared_ptr<PaxFilter> filter;
   List *targetlist = ps->plan->targetlist;
@@ -361,7 +361,7 @@ TableScanDesc PaxScanDesc::BeginScanExtractColumns(
   filter->SetColumnProjection(std::move(col_bits));
 
   if (pax_enable_sparse_filter) {
-    filter->InitSparseFilter(rel, qual);
+    filter->InitSparseFilter(rel, qual, key, nkeys);
 
     // FIXME: enable predicate pushdown can filter rows immediately without
     // assigning all columns. But it may mess the filter orders for multiple
@@ -375,7 +375,7 @@ TableScanDesc PaxScanDesc::BeginScanExtractColumns(
       filter->InitRowFilter(rel, ps, filter->GetColumnProjection());
     }
   }
-  return BeginScan(rel, snapshot, 0, nullptr, parallel_scan, flags,
+  return BeginScan(rel, snapshot, nkeys, key, parallel_scan, flags,
                    std::move(filter), build_bitmap);
 }
 
--- a/contrib/pax_storage/src/cpp/storage/filter/pax_sparse_pg_path.cc
+++ b/contrib/pax_storage/src/cpp/storage/filter/pax_sparse_pg_path.cc
@@ -36,7 +36,7 @@
 
 namespace pax {
 
-void PaxSparseFilter::Initialize(List *quals) {
+void PaxSparseFilter::Initialize(List *quals, ScanKey key, int nkeys) {
   ListCell *qual_cell;
   std::vector<std::shared_ptr<PFTNode>> fl_nodes; /* first level nodes */
   std::string origin_tree_str;
@@ -44,7 +44,7 @@ void PaxSparseFilter::Initialize(List *quals) {
   // no inited
   Assert(!filter_tree_);
 
-  if (!quals) {
+  if (!quals && nkeys == 0) {
     return;
   }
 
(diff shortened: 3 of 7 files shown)


contrib/pax_storage/src/cpp/access/pax_access_handle.cc
```diff
@@ -453,6 +453,7 @@ uint32 PaxAccessMethod::ScanFlags(Relation relation) {
   flags |= SCAN_FORCE_BIG_WRITE_LOCK;
 #endif
 
+  flags |= SCAN_SUPPORT_RUNTIME_FILTER;
```
yjhjstz, line 456:
```sql
minmax_columns = cbdb::GetMinMaxColumnIndexes(relation);
if (!minmax_columns.empty()) {
flags |= SCAN_SUPPORT_RUNTIME_FILTER;
}
```
need minmax_columns ?

gongxun0928: Min/Max filters depend on column statistics, but push-down bloom filters can support row-level filtering on the pax table without relying on column statistics. So regardless of whether there is min/max statistics for the column, we can push down the bloom filter

Row-level filtering via bloom filters will be implemented later.

Furthermore, we should consider enabling statistics updates for the PAX table in a bgworker.  Currently, minmax_columns needs to be specified when creating a table, and it will slow down the performance of data insertion.

yjhjstz: that means enable SCAN_SUPPORT_RUNTIME_FILTER default , will impact tpcds, tpch ?

gongxun0928: When gp_enable_runtime_filter_pushdown is enabled by default, it applies to all queries on PAX-formatted tables, which is expected to affect TPC-DS/TPC-H performance. 

IN and min-max filters rely on column-level statistics, whereas bloom filters perform row-level filtering. Even without IN or min-max filtering, simply pushing the bloom filter from SeqNext() down to the TableAM level can reduce unnecessary memory copies. For example, if a row has already been filtered by a bloom filter expression, there is no need to construct and return the corresponding TableTupleSlot for that row.

It leads to a minor performance improvement

contrib/pax_storage/src/cpp/storage/filter/pax_sparse_pg_path.cc
```diff
@@ -57,6 +57,23 @@ void PaxSparseFilter::Initialize(List *quals) {
...
+  // walk scan key and only support min/max filter now
+  for (int i = 0; i < nkeys; i++) {
+    // TODO: support bloom filter in PaxFilter
+    // but now just skip it, SeqNext() will check bloom filter in PassByBloomFilter()
+    if (key[i].sk_flags & SK_BLOOM_FILTER) {
+      continue;
+    }
+
+    if (key[i].sk_strategy != BTGreaterEqualStrategyNumber &&
+        key[i].sk_strategy != BTLessEqualStrategyNumber) {
```
gfphoenix78, line 60: Can't process column filters with `sk_strategy = BTEqualStrategyNumber`?

gongxun0928: currently the runtime filter will only push down the following three strategies. 
https://github.com/apache/cloudberry/blob/768801c3db840250da7d69d1ece13a41a7c35033/src/backend/executor/nodeHash.c#L4169-L4192

contrib/pax_storage/src/cpp/storage/filter/pax_sparse_pg_path.cc
```diff
@@ -57,6 +57,23 @@ void PaxSparseFilter::Initialize(List *quals) {
     fl_nodes.emplace_back(std::move(fl_node));
   }
 
+  // walk scan key and only support min/max filter now
+  for (int i = 0; i < nkeys; i++) {
```
leborchuk, line 52: We add ProcessScanKey nodes after the ExprWalker nodes. Does this mean ExprWalker are executed before ProcessScanKey? If so, maybe it would be cheaper to execute the ProcessScanKey first?

gongxun0928: That makes sense!

src/backend/executor/nodeSeqscan.c
```diff
@@ -80,13 +80,18 @@ SeqNext(SeqScanState *node)
 		 * node->filter_in_seqscan is false means scankey need to be pushed to
 		 * AM.
 		 */
-		if (gp_enable_runtime_filter_pushdown && !node->filter_in_seqscan)
+		if (gp_enable_runtime_filter_pushdown && node->filters &&
```
yjhjstz: lost `!node->filter_in_seqscan` condition ?

leborchuk (approved):

contrib/pax_storage/src/cpp/access/pax_scanner.cc
```diff
@@ -326,8 +326,8 @@ void PaxScanDesc::EndScan() {
 }
 
 TableScanDesc PaxScanDesc::BeginScanExtractColumns(
-    Relation rel, Snapshot snapshot, int /*nkeys*/,
-    struct ScanKeyData * /*key*/, ParallelTableScanDesc parallel_scan,
+    Relation rel, Snapshot snapshot, int nkeys,
+    struct ScanKeyData *key, ParallelTableScanDesc parallel_scan,
```
jiaqizho, line 330: `struct ScanKeyData *` -> `ScanKey` ?

my-ship-it (approved):

contrib/pax_storage/src/cpp/access/pax_scanner.cc
gongxun0928: It is better to keep the function signature consistent with it in the table access method
```
	/*
	 * GPDB: Extract columns for scan from targetlist and quals. This is mainly
	 * for AOCS tables.
	 */
	TableScanDesc	(*scan_begin_extractcolumns) (Relation rel,
												  Snapshot snapshot,
												  int nkeys, struct ScanKeyData *key,
												  ParallelTableScanDesc parallel_scan,
												  struct PlanState *ps,
												  uint32 flags);
```

PR #1363, opened 2025-09-19
Movable DataBase Locales for Cloudberry

leborchuk:
We inherited this issue from PostgreSQL.

PostgreSQL uses glibc to sort strings. In version glibc=2.28, collations broke down badly (in general, there are no guarantees when updating glibc). Changing collations breaks indexes. Similarly, a cluster with different collations also behaves unpredictably.

What and when something has changed in glibc can be found on https://github.com/ardentperf/glibc-unicode-sorting Also there is special postgresql-wiki https://wiki.postgresql.org/wiki/Locale_data_changes And you tube video https://www.youtube.com/watch?v=0E6O-V8Jato

In short, the issue can be seen through the use of bash:

( echo "1-1"; echo "11" ) | LC_COLLATE=en_US.UTF-8 sort

gives the different results in ubuntu 18.04 and 22.04.

There is no way to solve the problem other than by not changing the symbol order. We freeze symbol order and use it instead of glibc.

Here the solution https://github.com/postgredients/mdb-locales.

In this PR I have added PostgreSQL patch that replaces all glibc locale-related calls with a calls to an external libary. It activates using new configure parameter --with-mdblocales, which is off by default.

Using custom locales needs libmdblocales1 package and mdb-locales package with symbol table.

Build needs libmdblocales-dev package with headers.

Fixing the symbol order is necessary for OS upgrade. For example Ubuntu 22.04 EOL is April 2027, Rocky 8 Active Support ended May 2024, and Security support ends in 2029. 

We use Movable DataBase Locales in Greenplum 6 and all our PostgreSQL installations (starting with PostgreSQL 12). This patch is adopted patch version from our internal PostgreSQL 14 fork.


jianlirong:
Thank you very much for bringing up this issue and providing the
corresponding patch. Yes, we've run into this problem as well - the same
database version behaving inconsistently when deployed on different OS
platforms. This is definitely an issue that requires fixing.

Lirong


Leonid ***@***.***> 于2025年9月19日周五 20:26写道：

src/test/regress/sql/misc.sql
```diff
@@ -0,0 +1,271 @@
...
+
+DELETE FROM onek;
+
+COPY onek FROM '/home/xifos/git/cloudberry-gpdb/src/test/regress/results/onek.data';
+
+SELECT unique1 FROM onek WHERE unique1 < 2 ORDER BY unique1;
+
+DELETE FROM onek2;
+
+COPY onek2 FROM '/home/xifos/git/cloudberry-gpdb/src/test/regress/results/onek.data';
```
yjhjstz: path use like `COPY onek FROM '@abs_builddir@/results/onek.data';`

leborchuk:
Fixed all the issues, failed tests still waiting for the working CI

tuhaihe:
Hi @leborchuk, we can use `squash and merge`. Before that, we need to tidy up the commit message title and body. Is it sufficient to use the information from the first commit?

leborchuk:
> Hi @leborchuk, we can use `squash and merge`. Before that, we need to tidy up the commit message title and body. Is it sufficient to use the information from the first commit?

Yes, it's the original feature developer, It would be rude to remove him from the authors. I will squash the PR's on my laptop and update PR

PR #1383, opened 2025-10-09
Fetch yyjson at build time instead of vendoring it

tuhaihe:
Move yyjson from top-level dependency directory to `contrib/pax_storage/src/cpp/contrib/yyjson` since it's only used by the PAX storage component. Update all references in CMake build files, .gitmodules, and documentation to reflect the new location. This change improves the project structure by keeping PAX-specific dependencies together and removing the unnecessary top-level dependency directory.

The move includes:
- Moving yyjson directory to contrib/pax_storage/src/cpp/contrib/yyjson
- Updating CMakeLists.txt and pax.cmake to reference the new location
- Updating .gitmodules to reflect the new submodule path
- Updating README.md documentation to show the new location

Fixes #ISSUE_Number

### Type of Change
- [ ] Documentation update

### Test Plan

### Impact

**Performance:**

**User-facing changes:**

**Dependencies:**

### Checklist

### CI Skip Instructions

---


edespino:
Hi @tuhaihe !

I reviewed your PR and wanted to share an **alternative implementation** that eliminates the need for git submodules entirely. This approach uses CMake's `FetchContent` module, which I've seen work well in other Apache projects like MADlib.

## Alternative: CMake FetchContent

I've implemented this on branch [`fetchcontent-yyjson`](https://github.com/edespino/cloudberry/tree/fetchcontent-yyjson) for your consideration.

**Key change:** Replace the git submodule with automatic dependency fetching during CMake configuration.

### Implementation

**`contrib/pax_storage/CMakeLists.txt`**:
```cmake
if(USE_MANIFEST_API AND NOT USE_PAX_CATALOG)
    include(FetchContent)

    # Try system package first
    find_package(yyjson QUIET)

    if(NOT yyjson_FOUND)
        message(STATUS "yyjson not found in system, fetching from GitHub...")
        FetchContent_Declare(
            yyjson
            GIT_REPOSITORY https://github.com/ibireme/yyjson.git
            GIT_TAG 0.12.0
            GIT_SHALLOW TRUE
        )

        set(SAVED_BUILD_SHARED_LIBS ${BUILD_SHARED_LIBS})
        set(BUILD_SHARED_LIBS ON)

        FetchContent_MakeAvailable(yyjson)

        set(BUILD_SHARED_LIBS ${SAVED_BUILD_SHARED_LIBS})
    else()
        message(STATUS "Using system yyjson package")
    endif()
endif()
```

**`contrib/pax_storage/src/cpp/cmake/pax.cmake`**:
```cmake
# Update include path to use FetchContent variable
set(pax_target_include ${pax_target_include} ${yyjson_SOURCE_DIR}/src)
```

**Remove from `.gitmodules`** - No submodule entry needed.

### Benefits

1. **Simpler workflow** - No `git submodule update --init --recursive` required
2. **Automatic handling** - CMake downloads yyjson when needed during configuration
3. **Version explicit** - `GIT_TAG 0.12.0` (latest release) is clearer than a commit hash
   - Submodules point to opaque commit SHAs - without checking the yyjson repo, you can't tell if it's a release, a random commit, or how old it is
   - FetchContent uses semantic versions - immediately clear you're using the latest stable release
4. **System package support** - Respects system-installed yyjson if available
5. **Pattern consistency** - Aligns with how protobuf and zstd are already handled via `find_package()`

### Important Note

During my review, I noticed that **yyjson is only used when both `USE_MANIFEST_API=ON` and `USE_PAX_CATALOG=OFF`**, which is not the default configuration. This means:
- Default builds (`./configure --enable-pax`) don't use yyjson at all
- Your reorganization and this alternative have zero impact on normal builds
- The dependency is only relevant for the JSON manifest implementation mode

### Testing

To verify this works when yyjson IS needed:

```bash
cd contrib/pax_storage/build
cmake .. -DUSE_MANIFEST_API=ON -DUSE_PAX_CATALOG=OFF
# Should see: "yyjson not found in system, fetching from GitHub..."
make
```

### Recommendation

Your PR's organizational improvement is valuable regardless. This FetchContent approach is simply an alternative that:
- Reduces the dependency management burden (no submodule commands)
- Follows patterns used in other Apache projects
- Makes the build more self-contained

Both approaches work - this is just offered as food for thought. Happy to discuss!

**Branch for reference:** [`fetchcontent-yyjson`](https://github.com/edespino/cloudberry/tree/fetchcontent-yyjson)

yjhjstz (approved):

tuhaihe:
Hi @edespino, thanks for your great idea. There are other submodules needed for PAX building, such as tabulate. I prefer to keep the same behavior across yyjson and the other submodules so that users can download all required submodules with `git submodule update --init --recursive` once, rather than downloading `yyjson` separately during the build.

tuhaihe:
More:

I searched and found that `yyjson` is available via the EPEL repository on Rocky Linux 9, but it is not available in Ubuntu 22.04 (though it can be found in Ubuntu 25.04). Given that it is not a commonly provided package across popular Linux distros, we can continue using the submodule approach for now.

tuhaihe:
Would like to have more voices from the core PAX developers on this. cc @gfphoenix78 @jiaqizho @gongxun0928 

At least, it's not good to place the `yyjson` separately at the top of the dir.

edespino:
## Subject: Release Engineering View – Using EPEL vs. Building from Source

When building on Rocky Linux 9, EPEL can be convenient for missing dependencies, but it introduces several trade-offs worth noting:

### Key Issues

* **Reproducibility:** EPEL is rolling; rebuilds can silently pick up new versions. Exact, bit-reproducible releases become hard to guarantee.
* **Supply-chain control:** Packages are maintained outside ASF governance. Building from source keeps provenance and signatures within our audit trail.
* **Licensing clarity:** ASF policy expects verifiable source for all distributed code. Relying on EPEL binaries delegates license vetting to Fedora.
* **Runtime portability:** Binaries linked against EPEL libraries may fail on systems where EPEL isn’t enabled or where ABI flags differ.
* **Longevity:** EPEL mirrors move forward; old package versions aren’t preserved, complicating long-term rebuilds.

### Recommended Approach

* Build third-party libraries **from verified upstream source** whenever they appear in our release artifacts.
* Use EPEL only for transient **developer tools**, not for runtime dependencies.
* If EPEL is required in CI, **snapshot or mirror** it to fix versions.
* Record all external sources and hashes in `DEPENDENCIES` or release notes.

### Bottom Line

EPEL is fine for development convenience, but from a release-engineering and ASF-compliance standpoint, **source builds give stronger reproducibility, auditability, and long-term stability**.

edespino:
> Hi @edespino, thanks for your great idea. There are other submodules needed for PAX building, such as tabulate. I prefer to keep the same behavior across yyjson and the other submodules so that users can download all required submodules with `git submodule update --init --recursive` once, rather than downloading `yyjson` separately during the build.

One thing I'm curious about is why the selected submodule commit shas were used. It is not important for the relocation work. But it wasn't obvious at all why the yyjson submodule commit sha was used.

jiaqizho:
> Would like to have more voices from the core PAX developers on this. cc @gfphoenix78 @jiaqizho @gongxun0928
> 
> At least, it's not good to place the `yyjson` separately at the top of the dir.

The reason why `dependency/yyjson` is separated and stored instead of being placed under PAX is because not only PAX depends on `yyjson`. Does the core of cloud also rely on `yyjson`?

@gfphoenix78  do you still have any context about this module?

tuhaihe:
> > Would like to have more voices from the core PAX developers on this. cc @gfphoenix78 @jiaqizho @gongxun0928
> > At least, it's not good to place the `yyjson` separately at the top of the dir.
> 
> The reason why `dependency/yyjson` is separated and stored instead of being placed under PAX is because not only PAX depends on `yyjson`. Does the core of cloud also rely on `yyjson`?
> 

Thanks @jiaqizho for your reply. I remember our private conversation on this topic. ❤️

Now, from the open-source side, I believe we can move the `yyjson` to the correct place. If some downstream vendors refer to it as a dependency, the downstream should have a patch for it or clone it separately.

tuhaihe:
I have repushed this PR with a different approach: instead of relocating the
yyjson submodule, it removes the submodule and fetches yyjson at configure
time, as @edespino suggested above. Thanks Ed — your `fetchcontent-yyjson`
branch is what this is based on.

The reason for the change of direction is the 2.2.0 release.

## What prompted this

During the 2.2.0-rc1 IPMC vote, Justin Mclean raised several LICENSE
completeness issues, one of them about yyjson: he noted that
`dependency/yyjson/doc/doxygen/html/jquery.js` bundles jQuery, Sizzle,
jQuery UI and jQuery UI Touch Punch, plus doxygen-awesome-css, while our
LICENSE entry points only at yyjson's own MIT license.

That is accurate, and there is slightly more of it than reported. The
generated doxygen output under `dependency/yyjson/doc/` carries five groups
of third-party code that the root LICENSE does not account for:

| Component | License |
| --- | --- |
| jQuery 3.6.0 (Sizzle is inlined into it) | MIT, OpenJS Foundation |
| jQuery UI 1.12.1 | MIT, jQuery Foundation |
| jQuery UI Touch Punch 0.2.3 | dual MIT / GPLv2 |
| doxygen-awesome-css | MIT, jothepro |
| Doxygen's own helper scripts (`dynsections.js`, `navtree.js`, `resize.js`, `menu.js`) | MIT, Dimitri van Heesch |

The last group was not in Justin's list. It is worth noting that these are
MIT and not GPL: Doxygen itself is GPLv2, but it licenses the JavaScript it
emits under MIT via an explicit `@licstart` block, so there is no Category X
problem here — only a completeness one.

We could declare all five in the root LICENSE. But before doing that it is
worth asking whether we need to carry yyjson in the source tree at all.

## yyjson is not built by anything we ship

This also answers the question @jiaqizho raised earlier in this thread:

> The reason why `dependency/yyjson` is separated and stored instead of being
> placed under PAX is because not only PAX depends on `yyjson`. Does the core
> of cloud also rely on `yyjson`?

It does not. Searching `src/`, `contrib/`, `gpcontrib/`, `gpMgmt/` and
`gpAux/`, exactly one file references yyjson:

```
contrib/pax_storage/src/cpp/manifest/manifest.c:42:#include "yyjson.h"
```

There is no second consumer, in PAX or anywhere else.

Beyond that, even PAX does not compile it in any configuration we build.
`contrib/pax_storage/CMakeLists.txt` only pulls yyjson in under
`if(USE_MANIFEST_API AND NOT USE_PAX_CATALOG)`. `USE_MANIFEST_API` defaults
to `OFF` and `USE_PAX_CATALOG` defaults to `ON`, and nothing in the tree
turns them the other way — not CI, not the devops build scripts, not
configure. They can only be set by hand through the environment.

Checked against an actual release-configuration build on Rocky 9:

- no yyjson objects or `libyyjson*` anywhere in the build tree
- no `libyyjson*` in the installed `$GPHOME`
- `pax.so` and `libpaxformat.so` contain **0** yyjson symbols
- `manifest.c` produces no object file at all
- the build log confirms `build pax, USE_MANIFEST_API=OFF USE_PAX_CATALOG=ON`

So we ship a submodule that no released binary contains.

## What it costs to keep it

- `dependency/yyjson` is 9.3 MB across 804 files
- of which 3.0 MB and 201 files are generated doxygen output
- `pom.xml` carries **249** Apache RAT exclusions for it, **200** of them for
  the docs alone
- plus the five undeclared third-party components above

## The change

Fetching on demand removes the components rather than documenting them,
which seems clearly better for a dependency that no shipped build uses.

This also matches what we already do elsewhere in the tree: psutil, PyYAML
and PyGreSQL are downloaded during the build by `gpMgmt/bin/Makefile`, with
the comment "This replaces the previously bundled tar.gz files to comply with
Apache Release policy". yyjson is the same kind of problem with the same kind
of answer.

The version is now a release tag rather than a submodule pointer. Ed asked
earlier why the previous commit SHA was chosen — as far as I can tell there
was no particular reason: `61c03f6` is "Try fix github action", not a release
at all, on a tree whose `CMakeLists.txt` still said `VERSION 0.10.0`.

Removed along with the submodule: the `LICENSE` entry,
`licenses/LICENSE-yyjson.txt`, and all 249 RAT exclusions — 285 deletions
against 29 insertions, on top of the 9.3 MB the submodule itself brought in.

### One deviation from Ed's draft

Ed's branch tries a system yyjson through `find_package(yyjson QUIET)` before
falling back to fetching. I left that path out, because as written it cannot
work:

- on the `find_package` path, `yyjson_SOURCE_DIR` is never set, so the include
  path in `pax.cmake` would expand to `/src`
- an installed yyjson exports the target as `yyjson::yyjson`, but
  `pax.cmake` links the bare name `yyjson`

None of our build images can reach that branch either (yyjson is not in
Rocky's default repositories, and not in Ubuntu 22.04), so it would never be
exercised. Since the previous behaviour was always to build yyjson from
source, always fetching keeps this change behaviour-preserving. Happy to add
the system-package path back properly — via a `yyjson::yyjson` alias and the
target's `INTERFACE_INCLUDE_DIRECTORIES` — if people would rather have it.

## Verification

On Rocky 9:

- **Default configuration** — the PAX build succeeds with `http_proxy` and
  `https_proxy` unset, i.e. with no network access at all. This is the
  configuration every release build uses, and it is unaffected.
- **`USE_MANIFEST_API=ON USE_PAX_CATALOG=OFF`** — CMake reports
  `Fetching yyjson 0.12.0 for the manifest API`, `_deps/yyjson-src` checks out
  at tag `0.12.0`, `libyyjson.so.0.12.0` is built, `flags.make` puts
  `-I.../_deps/yyjson-src/src` on the compile line that `manifest.c` uses, and
  `link.txt` carries `libyyjson.so.0.12.0`.

## A separate finding for the PAX maintainers

While testing the second configuration I found that it does not build on
`main` at all, with or without this PR:

```
micro_partition_iterator_manifest.cc:309:30: error: 'struct RelationData' has
no member named 'rd_node'; did you mean 'rd_id'?
```

That is the PostgreSQL 16 rename of `rd_node` to `rd_locator`, which this file
never picked up. An unpatched tree with the submodule still in place fails at
the same two lines with the same error, so this is pre-existing and unrelated
to the change here.

Fixing it is more than those two lines — `pax_access_method_internal.cc` and
`catalog/manifest_api.h` still use `RelFileNode` in their `USE_MANIFEST_API`
branches as well. I will open that separately rather than mix it into this PR,
since it applies only to the PostgreSQL 16 based `main` branch. Flagging it
here mainly as a data point: the manifest API path has been unbuildable since
the PostgreSQL 16 merge, which is consistent with nothing enabling it.

cc @jiaqizho @leborchuk 

## Backpatch

2.2.0 is cut from `REL_2_STABLE`, not `main`, so this needs to land there too
to help the release. I checked the patch against `REL_2_STABLE`: everything
applies cleanly except one hunk in `LICENSE`, because #1987 already corrected
`dependencies/yyjson` to `dependency/yyjson` there. Resolving it just means
deleting those two lines either way.

my-ship-it (approved):
LGTM. It appears to only change how yyjson is fetched.

PR #1384, opened 2025-10-10
Optimize null bitmap handling for better append performance

gongxun0928:
1. Pre-allocate null bitmap capacity to pax_max_tuples_per_group to avoid capacity checks
2. initialize null bitmap with all bits set to 1 (0xff) by default, assuming non-null values are more frequent

Fixes #ISSUE_Number

### Type of Change
- [ ] Documentation update

### Test Plan

### Impact

**Performance:**

**User-facing changes:**

**Dependencies:**

### Checklist

### CI Skip Instructions

---

### DIFF
--- a/contrib/pax_storage/src/cpp/comm/bitmap.h
+++ b/contrib/pax_storage/src/cpp/comm/bitmap.h
@@ -134,12 +134,28 @@ struct BitmapRaw final {
     static_assert(BM_WORD_BITS == (1 << BM_WORD_SHIFTS));
     return (index >> BM_WORD_SHIFTS) < size;
   }
-  inline bool Empty() const {
+
+  inline bool Empty(uint32 end_index) const {
     if (!bitmap) return true;
-    for (size_t i = 0; i < size; i++)
-      if (bitmap[i]) return false;
+
+    uint32 end_word = BM_INDEX_WORD_OFF(end_index);
+    uint32 end_bit_offset = BM_INDEX_BIT_OFF(end_index);
+
+    for (uint32 i = 0; i < end_word && i < size; i++) {
+      if (bitmap[i] != 0) return false;
+    }
+
+    // Check partial word at end
+    if (end_word < size && end_bit_offset > 0) {
+      T mask = (T(1) << end_bit_offset) - 1;
+      if (bitmap[end_word] & mask) return false;
+    }
+
     return true;
   }
+
+  inline bool Empty() const { return Empty(size * sizeof(T) * 8ULL); }
+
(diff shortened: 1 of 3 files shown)


contrib/pax_storage/src/cpp/comm/bitmap.h
```diff
@@ -134,12 +134,31 @@ struct BitmapRaw final {
...
+    
+    for (uint32 i = 0; i < end_word && i < size; i++) {
+      if (bitmap[i] != 0) return false;
+    }
+    
+    // Check partial word at end
+    if (end_word < size && end_bit_offset > 0) {
+      T mask = (T(1) << end_bit_offset) - 1;
+      if (bitmap[end_word] & mask) return false;
+    }
```
gfphoenix78, line 152: If the function is in the hot path, we may need more optimization, like compare 64bits one time, SIMD.

contrib/pax_storage/src/cpp/comm/bitmap.h
```diff
@@ -247,24 +267,27 @@ class BitmapTpl final {
   }
 
   inline bool Empty() const { return raw_.Empty(); }
+  inline bool Empty(uint32 end_index) const {
```
gfphoenix78: Is end_index inclusive or exclusive? Please add comment to describe it.

contrib/pax_storage/src/cpp/comm/bitmap.h
```diff
@@ -134,12 +134,31 @@ struct BitmapRaw final {
...
+    if (end_word < size && end_bit_offset > 0) {
+      T mask = (T(1) << end_bit_offset) - 1;
+      if (bitmap[end_word] & mask) return false;
+    }
+    
     return true;
   }
+
+  inline bool Empty() const {
+    return Empty(size);
```
gfphoenix78: size is the number of sizeof(T), not the maximum index of the bitmap.

contrib/pax_storage/src/cpp/comm/bitmap.h
gongxun0928: it's not a hot path, and directly using BM_INDEX_WORD_OFF is efficient and straightforward. It depends on the length of type T. If T is uint64_t, it checks in batches based on a 64-bit width. If T is uint8_t, it checks in byte. 

The reason for not using uint64_t to process a uint8_t* bitmap is that the bitmap's address may not be 8-byte aligned, which would require copying the memory into a variable of type uint64_t for comparison.

PR #1397, opened 2025-10-16
feature: asynchronously update file statistics during vacuum

gongxun0928:
Previously, statistics (min-max, sum, count, etc.) were computed synchronously during data insertion, causing significant slowdowns due to heavy computational overhead.

This change introduces an asynchronous approach to maintain statistics:
- Add a GUC parameter to control statistics collection during writes (disabled by default)
- Skip statistics computation during INSERT to ensure fast writes
- Update statistics asynchronously during VACUUM on PAX tables by scanning file metadata
- Re-read files and refresh statistics only when metadata indicates they are stale
- A vacuum was implemented. First, it queries the data block file marked for deletion, scans for valid tuples within it, and rewrites them to a new data file.

In this PR, you'll notice changes in many test results related to VACUUM. The main reason is that after data has been updated or deleted, when VACUUM is performed, files marked for deletion will be rewritten—potentially consolidating multiple such files in the process.

It's important to note that this PR does not include the implementation of VACUUM FULL.

```
create table t1(c1 int, c2 int, c3 int, c4 int, c5 int, c6 int) using pax with(minmax_columns='c1,c2,c3,c4,c5,c6');
set pax.enable_sync_collect_stats to on; -- collect stats synchronously
insert into t1 select i,i,i,i,i,i from generate_series(1,1000000) i;
INSERT 0 1000000
Time: 2733.731 ms (00:02.734)

create table t2(c1 int, c2 int, c3 int, c4 int, c5 int, c6 int) using pax;
insert into t2 select i,i,i,i,i,i from generate_series(1,1000000) i;
INSERT 0 1000000
Time: 1816.836 ms (00:01.817)

```

Fixes #ISSUE_Number

### Type of Change
- [ ] Documentation update

### Test Plan

### Impact

**Performance:**

**User-facing changes:**

**Dependencies:**

### Checklist

### CI Skip Instructions

---


contrib/pax_storage/src/cpp/access/pax_inserter.cc
```diff
@@ -45,6 +45,7 @@ CPaxInserter::CPaxInserter(Relation rel)
 
   writer_->SetWriteSummaryCallback(&cbdb::InsertOrUpdateMicroPartitionEntry)
       ->SetFileSplitStrategy(std::make_unique<PaxDefaultSplitStrategy>())
+      ->SetEnableStats(pax::pax_enable_sync_collect_stats)
```
gfphoenix78, line 48: Will this behavior be controlled in cloud? How to set this value in functions used by cloud?

contrib/pax_storage/src/cpp/catalog/pax_aux_table.cc
```diff
@@ -278,6 +287,11 @@ void UpdateVisimap(Oid aux_relid, int block_id, const char *visimap_filename) {
         NameGetDatum(&pt_visimap_name);
   }
 
+  // visimap only update in delete operation, and stats is always valid
+  repls[ANUM_PG_PAX_BLOCK_TABLES_PTISSTATSVALID - 1] = true;
+  nulls[ANUM_PG_PAX_BLOCK_TABLES_PTISSTATSVALID - 1] = false;
+  values[ANUM_PG_PAX_BLOCK_TABLES_PTISSTATSVALID - 1] = BoolGetDatum(true);
```
gfphoenix78: Why is `isstatsvalid` changed if the visimap file is updated? If it's not changed, repls[..] should be false and leave nulls[..] and values[..] unset.

contrib/pax_storage/src/cpp/access/pax_access_handle.cc
```diff
@@ -411,6 +416,107 @@ void CCPaxAccessMethod::FinishBulkInsert(Relation relation, int options) {
...
+  {
+    std::vector<int> minmax_cols = cbdb::GetMinMaxColumnIndexes(rel);
+    std::vector<int> bf_cols = cbdb::GetBloomFilterColumnIndexes(rel);
+
+    if (minmax_cols.empty() && bf_cols.empty()) {
+      return;
+    }
+
+    Oid aux_oid = cbdb::GetPaxAuxRelid(RelationGetRelid(rel));
+    Relation aux_rel = cbdb::TableOpen(aux_oid, RowExclusiveLock);
```
gfphoenix78, line 552: Does it work with manifest meta? The aux heap table is not exist when meta data file is used.

contrib/pax_storage/src/cpp/storage/micro_partition_stats.cc
```diff
@@ -656,10 +648,17 @@ void MicroPartitionStats::AddRow(TupleTableSlot *slot) {
...
+    if (slot->tts_isnull[i]) {
+      if (update_null_stats) {
+        AddNullColumn(i);
+      }
+    } else {
+      if (update_null_stats) {
+        column_mem_stats_[i].null_count_stats_.all_null = false;
+        ++column_mem_stats_[i].null_count_stats_.not_null_count;
+      }
       AddNonNullColumn(i, slot->tts_values[i]);
```
gfphoenix78, line 660: How is the combination of update-stats and update-null valid?

contrib/pax_storage/src/cpp/access/pax_access_handle.cc
gfphoenix78: 1. Use auxiliary table directly. (USE_MANIFEST_API=OFF USE_PAX_CATALOG=ON)
2. Use manifest API to access metadata info. (USE_MANIFEST_API=ON USE_PAX_CATALOG=ON/OFF)

contrib/pax_storage/src/cpp/access/pax_inserter.cc
gongxun0928: If you want to control whether to collect statistics synchronously in cloud am , specify WriterOptions.enable_stats=true when creating MicroPartitionWriter.

contrib/pax_storage/src/cpp/access/pax_access_handle.cc
gongxun0928: done

PR #1424, opened 2025-10-31
Fix: tde encrypt buffer context don't use share memory

kongfanshen-0801:
Now we Fix: tde encrypt buffer context don't use share memory

We did not calculate the size of shared memory for BufEncCtx and BufDecCtx in the CreateSharedMemoryAndMemories function, which is a potential issue. There is actually no need to use shared memory here, as the parent process does not read the values of BufEncCtx and BufDecCtx. When we use AES algorithm the BufEncCtx and BufDecCtx also does not use shared memory. In order to maintain a consistent code style,we use malloc. cann't use palloc to apply for memory here, as TopmemoryContext is NULL during initialization and the memoryContext has not been setted yet.

Fixes #ISSUE_Number

### Type of Change
- [ ] Documentation update

### Test Plan

### Impact

**Performance:**

**User-facing changes:**

**Dependencies:**

### Checklist

### CI Skip Instructions

---

### DIFF
--- a/src/backend/crypto/bufenc.c
+++ b/src/backend/crypto/bufenc.c
@@ -36,6 +36,8 @@ static unsigned char buf_encryption_iv[BUFENC_IV_SIZE];
 
 void *BufEncCtx = NULL;
 void *BufDecCtx = NULL;
+static sm4_ctx sm4_enc_ctx;
+static sm4_ctx sm4_dec_ctx;
 
 static void set_buffer_encryption_iv(Page page, BlockNumber blkno);
 
@@ -47,16 +49,28 @@ InitializeBufferEncryption(void)
 	if (!FileEncryptionEnabled)
 		return;
 
+	/* 
+	 * To avoid memory leaks, when the postmaster resets the cluster, 
+	 * need to release the memory which was alloced at last time. 
+	 */
+	if (BufEncCtx && BufEncCtx != &sm4_enc_ctx)
+	{
+		pg_cipher_ctx_free(BufEncCtx);
+		BufEncCtx = NULL;
+	}
+	if (BufDecCtx && BufDecCtx != &sm4_dec_ctx)
+	{
+		pg_cipher_ctx_free(BufDecCtx);
+		BufDecCtx = NULL;
+	}
+	
 	key = KmgrGetKey(KMGR_KEY_ID_REL);
 
 	if (CheckIsSM4Method())
 	{
-		bool found;
-		BufEncCtx = ShmemInitStruct("sm4 encryption method encrypt ctx",
-												sizeof(sm4_ctx), &found);
+		BufEncCtx = &sm4_enc_ctx;
+		BufDecCtx = &sm4_dec_ctx;
 
-		BufDecCtx = ShmemInitStruct("sm4 encryption method decrypt ctx",
-												sizeof(sm4_ctx), &found);
 		sm4_ofb_setkey_enc((sm4_ctx *)BufEncCtx, (unsigned char *)key->key);
 		sm4_ofb_setkey_dec((sm4_ctx *)BufDecCtx, (unsigned char *)key->key);
 	}


src/backend/crypto/bufenc.c
```diff
@@ -51,12 +51,9 @@ InitializeBufferEncryption(void)
 
 	if (CheckIsSM4Method())
 	{
-		bool found;
-		BufEncCtx = ShmemInitStruct("sm4 encryption method encrypt ctx",
-												sizeof(sm4_ctx), &found);
+		BufEncCtx = malloc(sizeof(sm4_ctx));
```
my-ship-it: When should we free allocated memory?

kongfanshen-0801: can not free the memory. when run initdb command, the database run in single-user mode. We still need the BufEncCtx and BufDecCtx in postmaster process.

make create-demo-cluster  fail, because initdb raise error as follow:
performing post-bootstrap initialization ... LOG:  gp_role forced to 'utility' in single-user mode
2025-10-31 17:27:02.463981 CST,,,p27587,th407206016,,,,0,,,seg-10000,,,,sx1,"FATAL","XX001","invalid page in block 0 of relation global/1260",,,,,,,,"ReadBuffer_common","bufmgr.c",1138,
initdb: error: error while reading cdb_init.d directory: Broken pipe
initdb: removing data directory "/home/gpadmin/cloudberrydb/gpAux/gpdemo/datadirs/qddir/demoDataDir-1

Maybe the key is wrong in BufDecCtx  if we free the memory, then get wrong data after decrypt. At last get invalid page.

my-ship-it: So, is it possible to use static variable instead of malloced memory?

kongfanshen-0801: It is feasible to use static variables here. But multiple variables are needed.
The type of the BufEncCtx/BufDecCtx variable is void *, and different memory sizes can be applied depending on the encryption algorithm used.
If it is sm4, then the variable type is sm4_ctx.
If it is the AES algorithm, then the variable type is PgCipherCtx.
PgCipherCtx is not a fixed type and needs to be determined based on the compilation macro parameters. The code is as follows:
#ifdef USE_OPENSSL
typedef EVP_CIPHER_CTX PgCipherCtx;
#else
typedef void PgCipherCtx;
#endif
So here we use the void * type and apply different memory sizes according to different algorithms.

gfphoenix78: Why not use a static memory block? This function will be called in postmaster process. When the postmaster resets the cluster, it will allocate an additional memory buffer, so memory leaks.

gfphoenix78: The following `BufEncCtx = (void *)pg_cipher_ctx_create(...` may also allocate the memory block twice if the postmaster resets the cluster.

kongfanshen-0801: Cann't use the static memory block。 when enable AES algorithm,  we use open_ssl function EVP_CIPHER_CTX_new to alloc memory for BufEncCtx/BufDecCtx. The code of alloc memory in third party libraries.
When the postmaster resets the cluster, free memory if BufEncCtx/BufDecCtx is not null. This can avoid memory leaks.

src/backend/crypto/bufenc.c
```diff
@@ -47,16 +47,28 @@ InitializeBufferEncryption(void)
...
+	if (BufEncCtx)
+	{
+		free(BufEncCtx);
+		BufEncCtx = NULL;
+	}
+	if (BufDecCtx)
+	{
+		free(BufDecCtx);
+		BufDecCtx = NULL;
+	}
```
gfphoenix78: `free` doesn't match `pg_cipher_ctx_create()`. It's allocated by `EVP_CIPHER_CTX_new()`. We shouldn't assume it uses malloc/calloc internally.

gfphoenix78:
```c
static sm4_ctx sm4_enc_ctx;
static sm4_ctx sm4_dec_ctx;
...
if (BufEncCtx && BufEncCtx != &sm4_enc_ctx) {
    pg_cipher_ctx_free(BufEncCtx);
    BufEncCtx = NULL;
}
if (BufDecCtx && BufDecCtx != &sm4_dec_ctx) {
    pg_cipher_ctx_free(BufDecCtx);
    BufDecCtx = NULL;
}
...
if (CheckIsSM4Method())
{
  BufEncCtx = &sm4_enc_ctx;
  BufDecCtx = &sm4_dec_ctx;
}
...
```

kongfanshen-0801: Good point — I’ve updated the code accordingly.

my-ship-it (approved):

PR #1426, opened 2025-11-07
Update logfilter.py

jangjangji:
- Add buffer.seek(0) after truncate to reset file pointer position
- Add isinstance and len check before accessing item[16] to prevent IndexError
- Improve robustness when processing malformed CSV log entries

Fixes #ISSUE_Number

### What does this PR do?
This PR fixes two issues in the `CsvFlatten` class in `gplogfilter`:
1. **Buffer pointer reset:** Added `buffer.seek(0)` after `truncate(0)` to properly reset the StringIO file pointer position.  
2. **Index bounds checking:** Added validation before accessing `item[16]` to prevent `IndexError` on malformed CSV rows.

---

### Type of Change
- [ ] Documentation update

---

### Test Plan
- [x] Manual testing with production CSV log files  
- [x] Verified `gplogfilter` correctly filters CSV logs with time range and pattern matching  

---

### Impact
**Performance:**  
No measurable impact expected.

**User-facing changes:**  
None (internal logic improvement only).

**Dependencies:**  
No changes.

---

### Checklist

---

### Additional Context
N/A

---

### CI Skip Instructions

---

### DIFF
--- a/gpMgmt/bin/gppylib/logfilter.py
+++ b/gpMgmt/bin/gppylib/logfilter.py
@@ -278,12 +278,16 @@ class CsvFlatten(object):
         item = next(self.source)
         #we need to make a minor format change to the log level field so that
         # our single regex will match both.
-        item[16] = item[16] + ": "
+        # Check if item has enough columns before accessing index 16
+        if isinstance(item, list) and len(item) > 16:
+            item[16] = item[16] + ": "
 
         self.buffer.truncate(0)
+        self.buffer.seek(0)
         self.writer.writerow(item)
 
         return self.buffer.getvalue()
+
 
 #------------------------------- Spying --------------------------------
 


leborchuk (approved):
Looks good, ship it.

jangjangji:
> Looks good, ship it.

Thanks for reviewing!

gpMgmt/bin/gppylib/logfilter.py
```diff
@@ -278,13 +278,17 @@ def __next__(self):
         item = next(self.source)
         #we need to make a minor format change to the log level field so that
         # our single regex will match both.
-        item[16] = item[16] + ": "
+        # Check if item has enough columns before accessing index 16
+        if isinstance(item, list) and len(item) > 16:
+            item[16] = item[16] + ": "
```
gfphoenix78, line 283: When is the condition false? Is it correct to ignore the else branch?

jangjangji: **### When is the condition false?**

  The condition is false when len(item) <= 16, which occurs when the CSV reader incorrectly parses lines containing multi-line strings (e.g., SQL      
  statements with newlines).

  While our CSV files follow the standard 23-column PostgreSQL csvlog format, some log entries contain multi-line quoted strings that cause the CSV    
   reader to return incomplete rows with fewer than 17 columns.



###   **Is it correct to ignore the else branch?**

  Yes. The item[16] + ": " transformation is optional formatting to help regex pattern matching - not a required operation.

  Without this check, a single malformed line causes IndexError, which terminates the entire iterator silently (treated as StopIteration),
  resulting in 0 output despite reading thousands of lines.

  By skipping the transformation for malformed lines:
  - Valid lines (23 columns) → processed normally with the transformation
  - Malformed lines (< 17 columns) → passed through unchanged
  - Downstream filters naturally handle/discard non-matching entries

  This allows processing to continue instead of failing silently on the first malformed line.

jangjangji: @gfphoenix78  Sorry for the late reply. I've just added my response above. Please take a look when you have time.

PR #1434, opened 2025-11-12
Fix: Python build dependencies installation

### LINKED ISSUE #1433: [Bug] `pip3 install --user wheel "cython<3.0.0"` error on Ubuntu 24.04
### Apache Cloudberry version

main

### What happened

When running `make install -C ~/cloudberry`, errors returned:

```
# Download PyYAML using curl (only if not exists)
Downloading PyYAML-5.4.1.tar.gz...
# install dependencies of gpdemo
for file in gpdemo-defaults.sh lalshell generate_certs.sh demo_cluster.sh probe_config.sh README; do \
	/usr/bin/install -c -m 755 ../../gpAux/gpdemo/$file '/usr/local/cloudberry-db/bin/lib/gpdemo/'$file ; \
done
# Symlink gpcheckcat from bin to bin/lib to maintain backward compatibility
if [ ! -L /usr/local/cloudberry-db/bin/lib/gpcheckcat  ]; then \
	cd /usr/local/cloudberry-db/bin/lib/ && ln -s ../gpcheckcat gpcheckcat; \
fi
/usr/bin/install -c -m 644 gpload.bat '/usr/local/cloudberry-db/bin/gpload.bat'
# Download PyGreSQL using curl (only if not exists)
Downloading PyGreSQL-5.2.tar.gz...
# Install wheel and cython for PyYAML building
pip3 install --user wheel "cython<3.0.0"
error: externally-managed-environment

× This environment is externally managed
╰─> To install Python packages system-wide, try apt install
    python3-xyz, where xyz is the package you are trying to
    install.

    If you wish to install a non-Debian-packaged Python package,
    create a virtual environment using python3 -m venv path/to/venv.
    Then use path/to/venv/bin/python and path/to/venv/bin/pip. Make
    sure you have python3-full installed.

    If you wish to install a non-Debian packaged Python application,
    it may be easiest to use pipx install xyz, which will manage a
    virtual environment for you. Make sure you have pipx installed.

    See /usr/share/doc/python3.12/README.venv for more information.

note: If you believe this is a mistake, please contact your Python installation or OS distribution provider. You can override this, at the risk of breaking your Python installation or OS, by passing --break-system-packages.
hint: See PEP 668 for the detailed specification.
make[2]: *** [Makefile:100: download-python-deps] Error 1
make[2]: Leaving directory '/home/gpadmin/cloudberry/gpMgmt/bin'
make[1]: *** [Makefile:7: install-bin-recurse] Error 2
make[1]: Leaving directory '/home/gpadmin/cloudberry/gpMgmt'
make: *** [GNUmakefile:80: install] Error 2
make: Leaving directory '/home/gpadmin/cloudberry'
```

_No response_

### How to reproduce

Following the build docs: https://cloudberry.apache.org/docs/next/deployment/quick-build

### Operating System

Ubuntu 24.04.3 LTS

### Anything else

_No response_

### Code of Conduct

tuhaihe:
Improve wheel and cython dependency management in gpMgmt/bin/Makefile to handle Ubuntu 24.04's PEP 668 restrictions while maintaining compatibility with Rocky Linux and older Ubuntu versions.

Changes:
- Split wheel and cython dependency checks into separate commands
- Add fallback to --break-system-packages flag for Ubuntu 24.04+
- Only install dependencies if not already present in the system
- Maintain backward compatibility with existing build environments

This resolves build failures on Ubuntu 24.04 where pip install --user is restricted by default, while preserving the existing behavior on Rocky Linux 8/9 and Ubuntu 20.04/22.04 systems.

Fixes #1433

### Type of Change
- [ ] Documentation update

### Test Plan

### Impact

**Performance:**

**User-facing changes:**

**Dependencies:**

### Checklist

### CI Skip Instructions

---

### DIFF
--- a/gpMgmt/bin/Makefile
+++ b/gpMgmt/bin/Makefile
@@ -111,8 +111,19 @@ download-python-deps:
 	else \
 		echo "PyGreSQL-$(PYGRESQL_VERSION).tar.gz already exists, skipping download"; \
 	fi
-	# Install wheel and cython for PyYAML building
-	pip3 install --user wheel "cython<3.0.0"
+	# Install wheel and cython for PyYAML building (only if not exists)
+	@if python3 -c "import wheel" >/dev/null 2>&1; then \
+		echo "wheel already exists, skipping installation"; \
+	else \
+		echo "Installing wheel..."; \
+		pip3 install --user wheel 2>/dev/null || pip3 install --user --break-system-packages wheel; \
+	fi
+	@if python3 -c "import cython" >/dev/null 2>&1; then \
+		echo "cython already exists, skipping installation"; \
+	else \
+		echo "Installing cython..."; \
+		pip3 install --user "cython<3.0.0" 2>/dev/null || pip3 install --user --break-system-packages "cython<3.0.0"; \
+	fi
 
 #
 # PyGreSQL


tuhaihe:
We also need to update the installation docs to install the required dependencies before building Cloudberry.

gfphoenix78:
Could you explain why we need flag `--break-system-packages`? Where are these py packages installed, system library path for python? Is it possible to install them under `$GPHOME/lib`?

tuhaihe:
> Could you explain why we need flag `--break-system-packages`? Where are these py packages installed, system library path for python? Is it possible to install them under `$GPHOME/lib`?

Hi @gfphoenix78 Thanks for the review!

The `--break-system-packages` flag is needed for Ubuntu 24.04 and newer, which enforce [PEP 668](https://peps.python.org/pep-0668/)
 restrictions. This policy prevents `pip install --user` from modifying system-managed Python environments. Without this flag, pip will refuse to install packages like `wheel` or `cython`, causing build failures.

Regarding the installation location: when using `--user`, the packages are installed under the user’s local site directory (e.g., ~/.local/lib/python3.x/site-packages), not under the system library path. This avoids conflicts with system Python.

Installing them into `$GPHOME/lib` would require customizing `PYTHONPATH` and ensuring consistent Python environments across all build and runtime contexts, which might introduce complexity and compatibility issues. The current approach minimizes intrusion into the system and maintains backward compatibility with other platforms like Rocky Linux and older Ubuntu versions.

In short, `--break-system-packages` is only used as a fallback on Ubuntu 24.04+, while other systems continue using the existing safe paths.

We will also need to update the build docs to guide the users to install the needed packages from the software source. If these packages are detected as installed, then will skip the installation during the build.

gpMgmt/bin/Makefile
```diff
@@ -111,8 +111,19 @@ download-python-deps:
...
 		echo "PyGreSQL-$(PYGRESQL_VERSION).tar.gz already exists, skipping download"; \
 	fi
-	# Install wheel and cython for PyYAML building
-	pip3 install --user wheel "cython<3.0.0"
+	# Install wheel and cython for PyYAML building (only if not exists)
+	@if python3 -c "import wheel" >/dev/null 2>&1; then \
+		echo "wheel already exists, skipping installation"; \
+	else \
+		echo "Installing wheel..."; \
+		pip3 install --user wheel 2>/dev/null || pip3 install --user --break-system-packages wheel; \
```
gfphoenix78, line 119: `--break-system-packages` may hide possible problems.

tuhaihe: Is any possible approach recommended?

tuhaihe: Hi @gfphoenix78, thanks for raising this — the concern is valid.

We plan to update the docs ([[cloudberry-site#337](https://github.com/apache/cloudberry-site/pull/337)](https://github.com/apache/cloudberry-site/pull/337)) to recommend installing the required Python dependencies **before** building, so in most cases this step will be skipped entirely.

In the Makefile, `pip3 install --user --break-system-packages` is intentionally used **only as a fallback**, and only when the standard `--user` installation is blocked on Ubuntu 24.04 due to PEP 668. The dependencies involved here (`wheel`, `cython`) are **build-time only** and are installed into the user site, not the system Python environment.

That said, we agree this is not ideal long-term. A cleaner approach would be to isolate build-time Python dependencies (e.g. via a dedicated venv), which we can consider as a follow-up improvement. This change focuses on unblocking the Ubuntu 24.04 build while keeping backward compatibility with existing environments.

tuhaihe:
Hi @leborchuk could you please help take a look at this PR? Thanks!

leborchuk:
> Hi @leborchuk could you please help take a look at this PR? Thanks!

Ah, a very unusual code. Why would anyone even need to install wheel in the Makefile? For me is Ok to break system packages here, but I should understand why we need install it here, why just not to exit with error

leborchuk (approved):
+1 here for the venv, we cannot create a suitable environment here without virtual env, users linux packages could be very specific.

But right now we're not under venv, so no matter if we broke-system-packages or not, it's better to be broken, else we could get a completely unsuitable environment

tuhaihe:
Thanks for the review and feedback!

I agree that installing Python dependencies implicitly in the build system is not ideal in the long term, especially with PEP 668 and distro packaging policies.  

For this PR, I’d like to keep the current behavior for compatibility reasons (it is consistent with the existing build logic and avoids breaking current users and CI environments). This change mainly addresses the Ubuntu 24.04 build failure while preserving backward compatibility.

As a follow-up, I plan to propose a cleaner approach:
- Remove implicit pip installation from the build process.
- Recommend users to install Python dependencies explicitly via `pip3 install -r python-dependencies.txt` (or via venv / distro packages).
- Update the build documentation accordingly and make the build fail fast if required Python modules are missing.

This would align better with reproducible builds and packaging best practices, but it should be done in a phased manner to avoid disrupting existing workflows.

How about this? @leborchuk @gfphoenix78

leborchuk:
> Thanks for the review and feedback!
> 
> I agree that installing Python dependencies implicitly in the build system is not ideal in the long term, especially with PEP 668 and distro packaging policies.
> 
> For this PR, I’d like to keep the current behavior for compatibility reasons (it is consistent with the existing build logic and avoids breaking current users and CI environments). This change mainly addresses the Ubuntu 24.04 build failure while preserving backward compatibility.
> 
> As a follow-up, I plan to propose a cleaner approach:
> 
> * Remove implicit pip installation from the build process.
> * Recommend users to install Python dependencies explicitly via `pip3 install -r python-dependencies.txt` (or via venv / distro packages).
> * Update the build documentation accordingly and make the build fail fast if required Python modules are missing.
> 
> This would align better with reproducible builds and packaging best practices, but it should be done in a phased manner to avoid disrupting existing workflows.
> 
> How about this? @leborchuk @gfphoenix78

Yes, that's normal. Here we try to solve the issue that code does not work in Ubuntu 24.04. 

Of course, the whole approach should be reconsidered, we should just instruct script user how to prepare environment or get the prepared environment for him. But do not change the environment yourself. It'd be better to change it in another PR. Do not confuse doing a big refactoring with supporting Ubuntu 24.04 altogether.

my-ship-it (approved):

PR #1457, opened 2025-11-27
fix(backend): Skip interconnect initialization for auxiliary background workers

### LINKED ISSUE #1389: Auxiliary process opens unnecessary network sockets
Data exchange between backend processes uses network sockets. The listener socket is created when the backend process is initialized. However, some auxiliary process doesn't require interconnect for data exchange.
Exposing unnecessary network ports is not a good idea, which may be considered security issue in a higher standard.

zhan7236:
## Summary

This PR addresses Issue #1389 by preventing auxiliary background worker processes from opening unnecessary network sockets during interconnect initialization.

## Problem

Auxiliary background workers (ftsprobe, global deadlock detector, dtx recovery, sweeper, pg_cron launcher, ic proxy) were initializing the Motion Layer IPC, which opens network sockets for data exchange. Since these processes don't require interconnect communication, this behavior is unnecessary and can be considered a security concern.

## Solution

Added a check for `amAuxiliaryBgWorker()` in both `cdb_setup()` and `cdb_cleanup()` functions to skip interconnect initialization and cleanup for auxiliary background workers.

### Changes:
- Modified `src/backend/cdb/cdbutil.c`:
  - Added `!amAuxiliaryBgWorker()` condition to `cdb_setup()` to skip `InitMotionLayerIPC()` call
  - Added `!amAuxiliaryBgWorker()` condition to `cdb_cleanup()` to skip `CleanUpMotionLayerIPC()` call
  - Applied to both `#ifdef USE_INTERNAL_FTS` and `#else` code paths

## Testing

- Code review verified correct syntax and logic
- Header file `postmaster/postmaster.h` already included in cdbutil.c
- Function `amAuxiliaryBgWorker()` properly declared in postmaster.h

Fixes #1389

### DIFF
--- a/src/backend/utils/init/postinit.c
+++ b/src/backend/utils/init/postinit.c
@@ -1267,8 +1267,14 @@ InitPostgres(const char *in_dbname, Oid dboid, const char *username,
      * Primary function is to establish connections to the qExecs.
      * This is SKIPPED when the database is in bootstrap mode or 
      * Is not UnderPostmaster.
+     *
+     * Also skip for auxiliary background workers (e.g., ftsprobe, global
+     * deadlock detector, dtx recovery, sweeper, pg_cron launcher). These
+     * processes don't require interconnect for data exchange and opening
+     * unnecessary network sockets may be considered a security issue.
      */
-    if (!bootstrap && IsUnderPostmaster && !IsLoginMonitorWorkerProcess())
+    if (!bootstrap && IsUnderPostmaster && !IsLoginMonitorWorkerProcess() &&
+        !amAuxiliaryBgWorker())
     {
 		cdb_setup();
 		on_proc_exit( cdb_cleanup, 0 );


src/backend/cdb/cdbutil.c
```diff
@@ -1138,7 +1138,14 @@ cdb_setup(void)
 {
 	elog(DEBUG1, "Initializing Cloudberry components...");
 
-	if (Gp_role != GP_ROLE_UTILITY && !IS_SINGLENODE())
+	/*
```
my-ship-it: Wouldn't it be better if we don't call the cdb_setup & cdb_cleanup function?

zhan7236: Makes sense. Updated.

tuhaihe:
Hi @zhan7236 Thank you for your contribution! Your PR has now been merged into the main branch. Congratulations!

PR #1471, opened 2025-12-05
Remove some users of relfilenode.h

reshke:
This is my post-commit review of include/storage merge

as per https://git.postgresql.org/cgit/postgresql.git/commit/?h=b0a55e43299c


reshke:
I did not remove file itself (yet), because of remaining contrib user (PAX). Will do after later, in another pr. Is it OK?

PR #1481, opened 2025-12-08
Fix std::terminate() caused by uncaught exception in ORCA

NJrslv:
### What does this PR do?
This PR fixes a crash, `std::terminate()`, caused by an exception occurring in the catch block of `GPOPTOptimizedPlan()` during `gpopt_context.CloneErrorMsg()`.

When `CloneErrorMsg` fails (typically due to OOM), it throws a `gpdb` error, causing `std::terminate()`.

Postgres has graceful error handling for errors that occur while processing other errors - even during OOM, the `ErrorContext` has reserved space to preserve error messages. 

So rethrow the error to postgres handler.

### Test Plan
Added a new test to the regression `gporca_fault` test.
 
I added a query that errors out during optimization in orca and as the result falls into `catch` block where another exception in `CloneErrorMsg` is thrown via inject fault. 

Previously it resulted in the `std::terminate()` call & process crashing. Now catch the exception and let the postgres handle it with reserved buffer to preserve error message.

### Impact
Instead of 
```sql
server closed the connection unexpectedly
        This probably means the server terminated abnormally
        before or while processing the request.
```
postgres reports an error (in case of OOM we will see it, not std::terminate with exiting processes):
```sql
ERROR:  ...
```

### DIFF
--- a/src/test/regress/expected/gporca_faults_optimizer.out
+++ b/src/test/regress/expected/gporca_faults_optimizer.out
@@ -31,13 +31,13 @@ $$ LANGUAGE plpgsql VOLATILE;
 -- Query that runs the function on an additional non-QD master slice
 -- Include the EXPLAIN to ensure that this happens in the plan.
 EXPLAIN SELECT * FROM func1_nosql_vol(5), foo;
-                                             QUERY PLAN                                              
------------------------------------------------------------------------------------------------------
- Gather Motion 3:1  (slice1; segments: 3)  (cost=10000000000.25..10000001756.28 rows=86100 width=12)
-   ->  Nested Loop  (cost=10000000000.25..10000000608.28 rows=28700 width=12)
+                                           QUERY PLAN                                            
+-------------------------------------------------------------------------------------------------
+ Gather Motion 3:1  (slice1; segments: 3)  (cost=10000000000.25..10000000001.34 rows=3 width=12)
+   ->  Nested Loop  (cost=10000000000.25..10000000001.30 rows=1 width=12)
          ->  Broadcast Motion 1:3  (slice2)  (cost=0.25..0.28 rows=1 width=4)
                ->  Function Scan on func1_nosql_vol  (cost=0.25..0.26 rows=1 width=4)
-         ->  Seq Scan on foo  (cost=0.00..321.00 rows=28700 width=8)
+         ->  Seq Scan on foo  (cost=0.00..1.01 rows=1 width=8)
  Optimizer: Postgres query optimizer
 (6 rows)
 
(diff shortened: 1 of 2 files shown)


reshke:
👍

src/test/regress/expected/gporca_faults_optimizer.out
```diff
@@ -31,13 +31,13 @@ $$ LANGUAGE plpgsql VOLATILE;
 -- Query that runs the function on an additional non-QD master slice
 -- Include the EXPLAIN to ensure that this happens in the plan.
 EXPLAIN SELECT * FROM func1_nosql_vol(5), foo;
-                                           QUERY PLAN                                            
--------------------------------------------------------------------------------------------------
- Gather Motion 3:1  (slice1; segments: 3)  (cost=10000000000.25..10000000001.34 rows=3 width=12)
-   ->  Nested Loop  (cost=10000000000.25..10000000001.30 rows=1 width=12)
+                                             QUERY PLAN                                              
+-----------------------------------------------------------------------------------------------------
+ Gather Motion 3:1  (slice1; segments: 3)  (cost=10000000000.25..10000001756.28 rows=86100 width=12)
```
my-ship-it: Do we know why the number of estimated rows is different?

my-ship-it:
Thanks for the PR, which is a good start 😄

src/test/regress/expected/gporca_faults_optimizer.out
Jaroslavian: I ran this single test to generate `.out`:
```bash
src/test/regress/pg_regress --dbname=regression --load-extension=gp_inject_fault gporca_faults
```

The `foo` table isn’t analyzed in this case, so it produces this cost and # of rows (analyzing `foo` after creation doesn’t change the row count and costs).

I think the `.out` file was generated during a full regress run, where some previous test turned something on that triggered analyze during table creation.

Not a big deal?

my-ship-it: Ok, got it. It is better not to introduce unnecessary modifications.

tuhaihe:
That would be better to have a commit message title and message body when merging this PR. Can follow the [template](https://github.com/apache/cloudberry/blob/main/.gitmessage).

Jaroslavian:
If I am not the one squashing and merging this PR, I have provided the commit message and body below for the merger to copy & paste
```
Fix std::terminate() crash in GPOPTOptimizedPlan()
```

```
Fix a crash occurring when CloneErrorMsg() fails (e.g. from OOM)
and throws an exception inside a catch block. Previously, this 
triggered std::terminate().

Now, rethrow the exception to the Postgres handler. Postgres reserves
memory to handle errors gracefully, ensuring the user receives an
error message instead of the process terminating.

Includes a new regression test with fault injection.
```

my-ship-it (approved):

PR #1484, opened 2025-12-09
Remove PG-14 rebase artifact from dblink and fix for `dblink_get_conn`

reshke:
dblink_connstr_check changed its return type as part of 9b75844. It is error-prone to rebase dblink sources since then. There is no need for second dblink connstr check, even though it produces no harm.

Double-check from gpdb-7:
https://github.com/greenplum-db/gpdb-archive/blob/main/contrib/dblink/dblink.c#L290

Found during PG-16 rebase review.

### DIFF
--- a/src/Makefile.shlib
+++ b/src/Makefile.shlib
@@ -130,7 +130,7 @@ ifeq ($(PORTNAME), darwin)
     DLSUFFIX		= .so
     LINK.shared		= $(COMPILER) -bundle -multiply_defined suppress
   endif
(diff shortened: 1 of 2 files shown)


reshke:
UPD: use modified conn_str in dblink_get_conn, check out 8420c0f commit msg

https://github.com/apache/cloudberry/blob/9b758447dfb65dcb8024f2465c410221ded12841/contrib/dblink/dblink.c#L185-L195

yjhjstz (approved):

PR #1487, opened 2025-12-11
Fix rebase conflicts in pgstattuple

reshke:
`make installcheck` succeeds. 

rebase with respect to https://github.com/postgres/postgres/commit/37b2764593c073ca61c2baebd7d85666e553928b and https://github.com/postgres/postgres/commit/2ed532ee8c474e9767e76e1f3251cc3a0224358c


tuhaihe:
Very cool! Then can add the `make installcheck` to the test matrix later.

reshke:
> Very cool! Then can add the `make installcheck` to the test matrix later.

well, make installcheck now runs ok only for some extensions.
I am planning to fix every contrib extension, but this may take quite a bit of time.

PR #1489, opened 2025-12-11
FIx regress and TAP test for amcheck and enable in CI

reshke:
Fix amcheck regression test output 

Turns out `amcheck` contrib expression test were read since DIRECTORY TABLE feature.

Fix that, also add amcheck to CI

Also, fix TAP test. 003 - disable this test, 2PC tx are not supported for CBDB. for 001 - replace verify_heapam with verify_heapam_internal, for sake of this test expected output.

### DIFF
--- a/.github/workflows/build-cloudberry.yml
+++ b/.github/workflows/build-cloudberry.yml
@@ -286,6 +286,7 @@ jobs:
               },
               {"test":"ic-contrib",
                "make_configs":["contrib/auto_explain:installcheck",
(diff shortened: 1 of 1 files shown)


reshke:
Well, tap tests still fails... 
UPD: fixed

reshke:
Pushed 5728157da7e13bedc87543bf1adbd3e0e225e71a adding amcheck to yet another .yaml file... Not sure if this is correct though

PR #1502, opened 2025-12-22
Fix pg_rewind fail in TAP tests.

reshke:
With --enable-cassert pg_rewind TAP tests
fails in pg_log_generic_v function, 
in Assert(fmt[strlen(fmt) - 1] != '\n');

Trivially fix removing \n

This assertion is introduced cc8d41511721


x4m (approved):

reshke:
This is required for https://github.com/apache/cloudberry/pull/1506 to work

PR #1514, opened 2025-12-30
Run gp_internal_tools tests

andr-sokolov:


.github/workflows/build-deb-cloudberry.yml
```diff
@@ -246,7 +246,8 @@ jobs:
                                "gpcontrib/pxf_fdw:installcheck",
                                "gpcontrib/zstd:installcheck",
                                "gpcontrib/gp_sparse_vector:installcheck",
-                               "gpcontrib/gp_toolkit:installcheck"]
+                               "gpcontrib/gp_toolkit:installcheck",
+                               "gpcontrib/gp_internal_tools:installcheck"]
```
my-ship-it: Any regression diffs?

andr-sokolov: @my-ship-it Do you suggest to remove changes in this file, because the test is run from build-cloudberry.yml too?

tuhaihe: If any test failed, I think we can see the regression diffs in the `ic-gpcontrib` workflow, just like this: 
- https://github.com/apache/cloudberry/actions/runs/21382244420/job/61553702125.
- https://github.com/apache/cloudberry/actions/runs/21302314965/job/61380065385?pr=1514

tuhaihe:
Hi @andr-sokolov could you please resolve conflicts and rebase your PR?

.github/workflows/build-cloudberry.yml
```diff
@@ -312,7 +312,8 @@ jobs:
                                "gpcontrib/zstd:installcheck",
                                "gpcontrib/gp_sparse_vector:installcheck",
                                "gpcontrib/gp_toolkit:installcheck",
-                               "gpcontrib/gp_exttable_fdw:installcheck"]
+                               "gpcontrib/gp_exttable_fdw:installcheck",
```
tuhaihe, line 315: Hi @ostinru do you think there is still a necessity for running `gp_exttable_fdw` here? Would love to hear your thoughts!

ostinru: It seems that `gp_exttable_fdw` is for `gpfdist` and other external tables (not for PXF). So, it should be checked on cloudberry side (not pxf).
And, it is used by users, e.g. for loading data from Spark to GP/CB.

PR #1534, opened 2026-01-15
Use smgr-interface provided create_ao function.

reshke:
Extensible SMGR is one of usefull Cloudberry features that allow to hijack storage execution flow from extension. smgrcreate_ao however does not follow this memo. This commit fixes that.

### DIFF
--- a/src/backend/storage/smgr/smgr.c
+++ b/src/backend/storage/smgr/smgr.c
@@ -497,6 +497,11 @@ smgrcreate_ao(const struct f_smgr_ao *smgr,
 				RelFileNodeBackend rnode,
 				int32 segmentFileNum, bool isRedo)
 {
+	/* If we get there, check that provided smgr structure is sane.
+	 * First off all, outer Appendonly IO utilities should
+	 * pass valid smgr. Also, in-kernel smgr interface implementation
+	 * smgr_create_ao as mdcreate_ao, and extension should define its own. */
+	Assert(smgr != NULL && smgr->smgr_create_ao != NULL);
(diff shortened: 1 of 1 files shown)


src/backend/storage/smgr/smgr.c
```diff
@@ -492,9 +493,11 @@ smgrcreate(SMgrRelation reln, ForkNumber forknum, bool isRedo)
  *		already because we are in a WAL replay sequence.
  */
 void
-smgrcreate_ao(RelFileNodeBackend rnode, int32 segmentFileNum, bool isRedo)
+smgrcreate_ao(const struct f_smgr_ao *smgr,
+				RelFileNodeBackend rnode,
+				int32 segmentFileNum, bool isRedo)
 {
-	mdcreate_ao(rnode, segmentFileNum, isRedo);
+	smgr->smgr_create_ao(rnode, segmentFileNum, isRedo);
```
yjhjstz, line 505: need to check: if (smgr->smgr_create_ao), other f_smgr_ao maybe nullptr.

reshke: well, I can add assert here, because outer functions should guarantee that smgr itself is not null, and extension should guarantee that smgr->smgr_create_ao is not null...

or do you suggest real if statement for extra paranoia?

reshke: actually, I would say that `smgr == null or smgr->smgr_create_ao = null` is condition that should lead to ereport(ERROR)

yjhjstz: OK , Add an assert.

PR #1550, opened 2026-01-28
Cherry-pick pg14.5 commit: Fix incorrect permissions-checking code for extended statistics.

reshke:
Hi! As simple as $subj

I need this for more clean cherry-pick of CVE fix  https://github.com/postgres/postgres/commit/afe38fb6abb8ef364191201f577a5a638ba55509#diff-7548adcbaf4316229f591bae551542cf30cdab923ffa2de20ce23ef9bf787252

Original commit message follows:

=====

Commit a4d75c86b improved the extended-stats logic to allow extended stats to be collected on expressions not just bare Vars.  To apply such stats, we first verify that the user has permissions to read all columns used in the stats.  (If not, the query will likely fail at runtime, but the planner ought not do so.)  That had to get extended to check permissions of columns appearing within such expressions, but the code for that was completely wrong: it applied pull_varattnos to the wrong pointer, leading to "unrecognized node type" failures. Furthermore, although you couldn't get to this because of that bug, it failed to account for the attnum offset applied by pull_varattnos.

This escaped recognition so far because the code in question is not reached when the user has whole-table SELECT privilege (which is the common case), and because only subexpressions not specially handled by statext_is_compatible_clause_internal() are at risk.

I think a large part of the reason for this bug is under-documentation of what statext_is_compatible_clause() is doing and what its arguments are, so do some work on the comments to try to improve that.

Per bug #17570 from Alexander Kozhemyakin.  Patch by Richard Guo; comments and other cosmetic improvements by me.  (Thanks also to Japin Li for diagnosis.)  Back-patch to v14 where the bug came in.

Discussion: https://postgr.es/m/17570-f2f2e0f4bccf0965@postgresql.org

Fixes #ISSUE_Number

### Type of Change
- [ ] Documentation update

### Test Plan

### Impact

**Performance:**

**User-facing changes:**

**Dependencies:**

### Checklist

### CI Skip Instructions

---

### DIFF
--- a/src/test/regress/expected/stats_ext_optimizer.out
+++ b/src/test/regress/expected/stats_ext_optimizer.out
@@ -3231,6 +3231,10 @@ GRANT USAGE ON SCHEMA tststats TO regress_stats_user1;
 SET SESSION AUTHORIZATION regress_stats_user1;
 SELECT * FROM tststats.priv_test_tbl; -- Permission denied
 ERROR:  permission denied for table priv_test_tbl
+-- Check individual columns if we don't have table privilege
+SELECT * FROM tststats.priv_test_tbl
+  WHERE a = 1 and tststats.priv_test_tbl.* > (1, 1) is not null;
+ERROR:  permission denied for table priv_test_tbl
 -- Attempt to gain access using a leaky operator
 CREATE FUNCTION op_leak(int, int) RETURNS bool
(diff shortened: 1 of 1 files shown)


x4m (approved):

yjhjstz:
nice~ ,  squash commits.

reshke:
> nice~ , squash commits.

sure, as always

PR #1551, opened 2026-01-28
Cherry-pick of CVE fix: Fix privilege checks in pg_stats_ext and pg_s…

reshke:
…tats_ext_exprs.

This is 
https://github.com/postgres/postgres/commit/c3425383ba67ae6ecaddc8896025a91faadb430a commit, applied to Cloudberry. There was no issues in apply, only changes are to gporca expected output

original commit message follows

=== 

The catalog view pg_stats_ext fails to consider privileges for expression statistics.  The catalog view pg_stats_ext_exprs fails to consider privileges and row-level security policies.  To fix, restrict the data in these views to table owners or roles that inherit privileges of the table owner.  It may be possible to apply less restrictive privilege checks in some cases, but that is left as a future exercise.  Furthermore, for pg_stats_ext_exprs, do not return data for tables with row-level security enabled, as is already done for pg_stats_ext.

On the back-branches, a fix-CVE-2024-4317.sql script is provided that will install into the "share" directory.  This file can be used to apply the fix to existing clusters.

Bumps catversion on 'master' branch only.

Reported-by: Lukas Fittl
Reviewed-by: Noah Misch, Tomas Vondra, Tom Lane
Security: CVE-2024-4317
Backpatch-through: 14

### DIFF
--- a/.github/workflows/build-cloudberry-rocky8.yml
+++ b/.github/workflows/build-cloudberry-rocky8.yml
@@ -1,0 +1,1909 @@
+# --------------------------------------------------------------------
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements. See the NOTICE file distributed
+# with this work for additional information regarding copyright
+# ownership. The ASF licenses this file to You under the Apache
+# License, Version 2.0 (the "License"); you may not use this file
+# except in compliance with the License. You may obtain a copy of the
+# License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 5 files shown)


avamingli (dismissed):
LGTM. Thanks for cherry-pick.

reshke:
> LGTM. Thanks for cherry-pick.

Thank you for looking into this. 

> OH, I just noticed that we should keep origin commit from Postgres.

Sorry, I didn't get it. Origin commit is postgres/postgres@c342538 ...

reshke:
I still dont get what ought to be done here :(

avamingli:
> I still dont get what ought to be done here :(

Merged. I’ve asked @lss602726449 to review since he has been working on pg_stats_ext recently — thank you for that.
I squashed and merged the changes to avoid a merge commit in the main branch, as we prefer using rebase instead.
Thanks for your contribution!

PR #1562, opened 2026-02-04
Backpatch CVE-fix: Fix security checks in selectivity estimation func…

reshke:
…tions.

origin commit:
https://git.postgresql.org/cgit/postgresql.git/commit/?id=afe38fb6abb8ef364191201f577a5a638ba55509

Applied sql regression changes to singlenode regress & PAX regress too.

====

Commit e2d4ef8de86 (the fix for CVE-2017-7484) added security checks to the selectivity estimation functions to prevent them from running user-supplied operators on data obtained from pg_statistic if the user lacks privileges to select from the underlying table. In cases involving inheritance/partitioning, those checks were originally performed against the child RTE (which for plain inheritance might actually refer to the parent table). Commit 553d2ec2710 then extended that to also check the parent RTE, allowing access if the user had permissions on either the parent or the child. It turns out, however, that doing any checks using the child RTE is incorrect, since securityQuals is set to NULL when creating an RTE for an inheritance child (whether it refers to the parent table or the child table), and therefore such checks do not correctly account for any RLS policies or security barrier views. Therefore, do the security checks using only the parent RTE. This is consistent with how RLS policies are applied, and the executor's ACL checks, both of which use only the parent table's permissions/policies. Similar checks are performed in the extended stats code, so update that in the same way, centralizing all the checks in a new function.

In addition, note that these checks by themselves are insufficient to ensure that the user has access to the table's data because, in a query that goes via a view, they only check that the view owner has permissions on the underlying table, not that the current user has permissions on the view itself. In the selectivity estimation functions, there is no easy way to navigate from underlying tables to views, so add permissions checks for all views mentioned in the query to the planner startup code. If the user lacks permissions on a view, a permissions error will now be reported at planner-startup, and the selectivity estimation functions will not be run.

Checking view permissions at planner-startup in this way is a little ugly, since the same checks will be repeated at executor-startup. Longer-term, it might be better to move all the permissions checks from the executor to the planner so that permissions errors can be reported sooner, instead of creating a plan that won't ever be run. However, such a change seems too far-reaching to be back-patched.

Back-patch to all supported versions. In v13, there is the added complication that UPDATEs and DELETEs on inherited target tables are planned using inheritance_planner(), which plans each inheritance child table separately, so that the selectivity estimation functions do not know that they are dealing with a child table accessed via its parent. Handle that by checking access permissions on the top parent table at planner-startup, in the same way as we do for views. Any securityQuals on the top parent table are moved down to the child tables by inheritance_planner(), so they continue to be checked by the selectivity estimation functions.

Author: Dean Rasheed <dean.a.rasheed@gmail.com>
Reviewed-by: Tom Lane <tgl@sss.pgh.pa.us>
Reviewed-by: Noah Misch <noah@leadboat.com>
Backpatch-through: 13
Security: CVE-2025-8713

### DIFF
--- a/src/test/regress/expected/rowsecurity_optimizer.out
+++ b/src/test/regress/expected/rowsecurity_optimizer.out
@@ -4625,7 +4625,7 @@ RESET SESSION AUTHORIZATION;
 DROP VIEW rls_view;
 DROP TABLE rls_tbl;
 DROP TABLE ref_tbl;
--- Leaky operator test
+-- Leaky operator tests
 CREATE TABLE rls_tbl (a int);
 NOTICE:  Table doesn't have 'DISTRIBUTED BY' clause -- Using column named 'a' as the Greenplum Database data distribution key for this table.
 HINT:  The 'DISTRIBUTED BY' clause determines the distribution of data. Make sure column(s) chosen are the optimal data distribution key to minimize skew.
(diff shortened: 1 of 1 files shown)


reshke:
ahhh will fix tests soon

leborchuk:
Test rowsecurity should be fixed here. 

The original fix https://github.com/postgres/postgres/commit/e2d4ef8de869c57e3bf270a30c12d48c2ce4e00c#diff-60407b5c46aaf9ea36ed1002b1bfaf5e818298c30d3d680c4e9c91f814befd46 committed on May 8, 2017. 14-4 was released on Jun 13, 2022. Do not understand why we missed this fix, but anyway, it should be fixed, thank you very much @reshke for your catch.

reshke:
> Test rowsecurity should be fixed here.
> 
> The original fix [postgres/postgres@e2d4ef8#diff-60407b5c46aaf9ea36ed1002b1bfaf5e818298c30d3d680c4e9c91f814befd46](https://github.com/postgres/postgres/commit/e2d4ef8de869c57e3bf270a30c12d48c2ce4e00c#diff-60407b5c46aaf9ea36ed1002b1bfaf5e818298c30d3d680c4e9c91f814befd46) committed on May 8, 2017. 14-4 was released on Jun 13, 2022. Do not understand why we missed this fix, but anyway, it should be fixed, thank you very much @reshke for your catch.

We did not miss CVE-2017-7484 https://github.com/apache/cloudberry/commit/e2d4ef8de869c57e3bf270a30c12d48c2ce4e00c

tuhaihe:
Hi @reshke would you like to rebase this PR?

PR #1569, opened 2026-02-09
Fix the regex matching bug of the GUC gp_resource_group_cgroup_parent

fairyfar:
Fix the regex matching bug of the GUC gp_resource_group_cgroup_parent

### What does this PR do?

This is a code defect of a original GPDB on resource group enabled.

There is a bug to calculate length of pg_wchar in `gpvars_check_gp_resource_group_cgroup_parent` function.

For example, the value "greenplum database" was originally supposed to be judged as an illegal name, and report error: 

"gp_resource_group_cgroup_parent can only contains alphabet, number and non-leading . _ -".

But it was wrongly judged as legal.

### Type of Change
- [ ] Documentation update

### Test Plan

### Impact

**Performance:**

**User-facing changes:**

**Dependencies:**

### Checklist

### CI Skip Instructions

---


leborchuk:
The same issue is in archived gp7 repo. It looks like we always copy first 4 bytes of data, which could leads not only to wrong name but also to buffer overflow in a case of empty string

tuhaihe:
Hi @fairyfar, thanks for your great work! Done.

PR #1571, opened 2026-02-10
Enhance parallel foreign scan support

MisterRaindrop:
support parallel foreign scan support and add a new mock fdw for testing parallel foreign scans.

Key changes include:
- Implementation of `parallel_foreign_scan_test_fdw` to generate synthetic rows and support parallel scanning.
- Modifications to the optimizer to generate gather paths for foreign tables with parallel capabilities.
- Updates to `execMain.c` to enable parallel mode for gather nodes based on the execution context.
- Addition of test cases to validate the functionality of the new FDW in both coordinator and all-segments modes.

### Type of Change
- [ ] Documentation update

### Test Plan

### Impact

**Performance:**

**User-facing changes:**

**Dependencies:**

### Checklist

### CI Skip Instructions

---

### DIFF
--- a/src/backend/optimizer/util/pathnode.c
+++ b/src/backend/optimizer/util/pathnode.c
@@ -3004,7 +3004,6 @@ GatherPath *
 create_gather_path(PlannerInfo *root, RelOptInfo *rel, Path *subpath,
 				   PathTarget *target, Relids required_outer, double *rows)
 {
-	Assert(false);
 	GatherPath *pathnode = makeNode(GatherPath);
 
 	Assert(subpath->parallel_safe);
@@ -3018,6 +3017,16 @@ create_gather_path(PlannerInfo *root, RelOptInfo *rel, Path *subpath,
 	pathnode->path.parallel_safe = false;
 	pathnode->path.parallel_workers = 0;
 	pathnode->path.pathkeys = NIL;	/* Gather has unordered result */
+
+	/* Inherit locus from subpath — Gather collects within the same segment,
+	 * data distribution across segments doesn't change. */
+	pathnode->path.locus = subpath->locus;
+	pathnode->path.locus.parallel_workers = 0;	/* Gather output is single-stream */
+
+	pathnode->path.motionHazard = subpath->motionHazard;
+	pathnode->path.barrierHazard = subpath->barrierHazard;
+	pathnode->path.rescannable = false;
+	pathnode->path.sameslice_relids = subpath->sameslice_relids;
 
 	pathnode->subpath = subpath;
 	pathnode->num_workers = subpath->parallel_workers;
--- a/src/backend/executor/execMain.c
+++ b/src/backend/executor/execMain.c
@@ -2741,6 +2741,18 @@ ExecutePlan(EState *estate,
 	if (!execute_once || GP_ROLE_DISPATCH == Gp_role)
 		use_parallel_mode = false;
 
+	/*
+	 * If the plan contains Gather/GatherMerge nodes (parallelModeNeeded),
+	 * enable PG-style parallel mode on segments even if CBDB MPP parallel
+	 * mode (useMppParallelMode) didn't set it. This ensures ExecGather
+	 * can launch parallel workers.
+	 */
+	if (!use_parallel_mode &&
+		execute_once &&
+		Gp_role != GP_ROLE_DISPATCH &&
+		estate->es_plannedstmt->parallelModeNeeded)
+		use_parallel_mode = true;
+
 	estate->es_use_parallel_mode = use_parallel_mode;
 	if (use_parallel_mode)
 		EnterParallelMode();
--- a/src/backend/optimizer/path/allpaths.c
+++ b/src/backend/optimizer/path/allpaths.c
@@ -1241,6 +1241,27 @@ set_foreign_pathlist(PlannerInfo *root, RelOptInfo *rel, RangeTblEntry *rte)
 	/* Call the FDW's GetForeignPaths function to generate path(s) */
 	rel->fdwroutine->GetForeignPaths(root, rel, rte->relid);
 	adjust_costs_for_mpp_foreign_scan(rel->pathlist);
+
+	/* Generate Gather paths for FDW parallel partial paths.
+	 * Scoped to foreign tables only — does not affect SeqScan/IndexScan. */
+	if (rel->partial_pathlist != NIL)
+	{
+		Path   *cheapest_partial = (Path *) linitial(rel->partial_pathlist);
+
+		if (cheapest_partial->parallel_safe &&
+			cheapest_partial->parallel_workers > 0)
+		{
+			double	rows = cheapest_partial->rows *
+						   cheapest_partial->parallel_workers;
+
+			add_path(rel,
+					 (Path *) create_gather_path(root, rel,
+												 cheapest_partial,
+												 rel->reltarget,
+												 NULL, &rows),
+					 root);
+		}
+	}
 }
 
 /*
--- a/src/test/fdw/Makefile
+++ b/src/test/fdw/Makefile
@@ -1,4 +1,4 @@
-SUBDIRS = extension
+SUBDIRS = extension parallel_foreign_scan_test
 
 PROGRAM = extended_protocol_commit_test
 
--- a/src/test/fdw/parallel_foreign_scan_test/Makefile
+++ b/src/test/fdw/parallel_foreign_scan_test/Makefile
@@ -1,0 +1,13 @@
+EXTENSION = parallel_foreign_scan_test_fdw
+DATA = parallel_foreign_scan_test_fdw--1.0.sql
+MODULE_big = parallel_foreign_scan_test_fdw
+OBJS = parallel_foreign_scan_test_fdw.o
+REGRESS = parallel_foreign_scan_test
+
+subdir = src/test/fdw/parallel_foreign_scan_test
+top_builddir = ../../../..
+include $(top_builddir)/src/Makefile.global
+NO_PGXS = 1
+include $(top_srcdir)/src/makefiles/pgxs.mk
+
+installcheck: install
--- a/src/test/fdw/parallel_foreign_scan_test/expected/parallel_foreign_scan_test.out
+++ b/src/test/fdw/parallel_foreign_scan_test/expected/parallel_foreign_scan_test.out
@@ -1,0 +1,111 @@
+-- Test parallel foreign scan support in the Cloudberry kernel.
+-- Uses a mock FDW that generates synthetic rows (id, val).
+CREATE EXTENSION parallel_foreign_scan_test_fdw;
+-- ============================================================
+-- PART 1: Coordinator mode (deterministic, tests planner paths)
+-- ============================================================
+CREATE SERVER coord_srv FOREIGN DATA WRAPPER parallel_foreign_scan_test_fdw
+  OPTIONS (mpp_execute 'coordinator');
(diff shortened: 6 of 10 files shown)


src/backend/optimizer/util/pathnode.c
```diff
@@ -3019,6 +3018,16 @@ create_gather_path(PlannerInfo *root, RelOptInfo *rel, Path *subpath,
...
+	/* Inherit locus from subpath — Gather collects within the same segment,
+	 * data distribution across segments doesn't change. */
+	pathnode->path.locus = subpath->locus;
+	pathnode->path.locus.parallel_workers = 0;	/* Gather output is single-stream */
+
+	pathnode->path.motionHazard = subpath->motionHazard;
+	pathnode->path.barrierHazard = subpath->barrierHazard;
+	pathnode->path.rescannable = false;
+	pathnode->path.sameslice_relids = subpath->sameslice_relids;
+
```
avamingli, line 3030: create_gather_path was guarded by Assert(false) for a reason — the locus semantics of Gather in an MPP context were never fully defined or implemented.

The comment is wrong, and the conclusion is wrong. 
Gather does change the data distribution. 
Simplest example: a Hash-distributed table's parallel partial scan has locus HashWorkers (data is hash-distributed across segments, and within each segment it's split across workers). 
Gather collects all workers' data back to the leader process, so the locus should become Hash. You can't just copy the subpath locus and zero out parallel_workers — that's incorrect for HashWorkers , SegmentGeneral, and might be other locus types as well.
Also, this change is global — create_gather_path isn't FDW-specific. Once the Assert is gone, every code path that calls this function is affected. The hardest case is JOINs — mixing Gather with CBDB-style parallelism (Motion/slice) introduces a ton of problems. This PR doesn't seem to have considered any of that.

MisterRaindrop:
> Overall, FDW parallel scan is a direction worth exploring, but this approach is too rough. The core problems are:
> 
> 1. locus transition semantics for Gather in an MPP context haven't been thought through, and the changes are too broad.
> 2. FDW is a black box from the database's perspective.
>    For heap tables we have parallel scan (divide work by pages), for AO/AOCS we have parallel scan (divide work by files) — the work partitioning is well-defined.
>    But for FDWs, the parallel behavior depends entirely on the FDW's own implementation. If an FDW (say file_fdw) sets parallel_safe = true following planner's parallel logic but doesn't actually implement the DSM parallel callbacks (EstimateDSMForeignScan, InitializeDSMForeignScan, InitializeWorkerForeignScan), then multiple workers will each scan the full dataset, producing duplicate rows.

I'm not very familiar with Cloudberry. Still learning.

FDW itself is a black box. Its specific implementation largely depends on how the user implements it. My understanding is that users need to take responsibility for their own implementations. Additionally, I should only enable gather for FDW. In other cases, it should remain false, this will parallel processing advantages of PostgreSQL?


Additionally, I've looked into other aspects of FDW parallelism. Currently, it seems there is no optimal solution.

So, should we aim to implement parallelism that is transparent to users? Or are there better approaches? Could you share some idea?

avamingli:
> > Overall, FDW parallel scan is a direction worth exploring, but this approach is too rough. The core problems are:
> > 
> > 1. locus transition semantics for Gather in an MPP context haven't been thought through, and the changes are too broad.
> > 2. FDW is a black box from the database's perspective.
> >    For heap tables we have parallel scan (divide work by pages), for AO/AOCS we have parallel scan (divide work by files) — the work partitioning is well-defined.
> >    But for FDWs, the parallel behavior depends entirely on the FDW's own implementation. If an FDW (say file_fdw) sets parallel_safe = true following planner's parallel logic but doesn't actually implement the DSM parallel callbacks (EstimateDSMForeignScan, InitializeDSMForeignScan, InitializeWorkerForeignScan), then multiple workers will each scan the full dataset, producing duplicate rows.
> 
> I'm not very familiar with Cloudberry. Still learning.
> 
> FDW itself is a black box. Its specific implementation largely depends on how the user implements it. My understanding is that users need to take responsibility for their own implementations. Additionally, I should only enable gather for FDW. In other cases, it should remain false, this will parallel processing advantages of PostgreSQL?
> 
> Additionally, I've looked into other aspects of FDW parallelism. Currently, it seems there is no optimal solution.
> 
> So, should we aim to implement parallelism that is transparent to users? Or are there better approaches? Could you share some idea?

Neither PostgreSQL nor Cloudberry supports parallel FDW scans, that's a deliberate decision, not an oversight.

On the implementation side: having the kernel generate partial paths for FDW will cause FDWs that don't implement parallel scan callbacks to silently produce wrong results (e.g. duplicate rows). That's a kernel bug, not a user error — we can't shift that responsibility to FDW authors. And mixing Gather with CBDB-style parallelism remains fundamentally broken — the locus handling is wrong, and none of the issues I raised (joins, locus transitions, the overly broad execMain.c change) have been addressed.

More importantly, before discussing how, we need to answer why. What real-world problem does this solve in an MPP system where FDW is already used across segments? And given the risks I mentioned above — broken locus transitions, silent wrong results for existing FDWs, untested join/subquery interactions — even if it can be done, is it worth the complexity? If you want to push this forward, you need to make the case clearly: what's the motivation, and convince us that all the issues raised have sound solutions.

MisterRaindrop:
> > > Overall, FDW parallel scan is a direction worth exploring, but this approach is too rough. The core problems are:
> > > 
> > > 1. locus transition semantics for Gather in an MPP context haven't been thought through, and the changes are too broad.
> > > 2. FDW is a black box from the database's perspective.
> > >    For heap tables we have parallel scan (divide work by pages), for AO/AOCS we have parallel scan (divide work by files) — the work partitioning is well-defined.
> > >    But for FDWs, the parallel behavior depends entirely on the FDW's own implementation. If an FDW (say file_fdw) sets parallel_safe = true following planner's parallel logic but doesn't actually implement the DSM parallel callbacks (EstimateDSMForeignScan, InitializeDSMForeignScan, InitializeWorkerForeignScan), then multiple workers will each scan the full dataset, producing duplicate rows.
> > 
> > 
> > I'm not very familiar with Cloudberry. Still learning.
> > FDW itself is a black box. Its specific implementation largely depends on how the user implements it. My understanding is that users need to take responsibility for their own implementations. Additionally, I should only enable gather for FDW. In other cases, it should remain false, this will parallel processing advantages of PostgreSQL?
> > Additionally, I've looked into other aspects of FDW parallelism. Currently, it seems there is no optimal solution.
> > So, should we aim to implement parallelism that is transparent to users? Or are there better approaches? Could you share some idea?
> 
> Neither PostgreSQL nor Cloudberry supports parallel FDW scans, that's a deliberate decision, not an oversight.
> 
> On the implementation side: having the kernel generate partial paths for FDW will cause FDWs that don't implement parallel scan callbacks to silently produce wrong results (e.g. duplicate rows). That's a kernel bug, not a user error — we can't shift that responsibility to FDW authors. And mixing Gather with CBDB-style parallelism remains fundamentally broken — the locus handling is wrong, and none of the issues I raised (joins, locus transitions, the overly broad execMain.c change) have been addressed.
> 
> More importantly, before discussing how, we need to answer why. What real-world problem does this solve in an MPP system where FDW is already used across segments? And given the risks I mentioned above — broken locus transitions, silent wrong results for existing FDWs, untested join/subquery interactions — even if it can be done, is it worth the complexity? If you want to push this forward, you need to make the case clearly: what's the motivation, and convince us that all the issues raised have sound solutions.

Parallel FDW primarily addresses the issue of slow data loading. This functionality was already implemented in earlier versions of PostgreSQL. Now, I am attempting to integrate this feature into MPP systems. In simple tests, parallelization has indeed delivered a performance improvement of one to two times. Such gains are essential for performance-sensitive business scenarios. Therefore, I am working to introduce this functionality. Alternatively, we could discuss the implementation plan in the issue tracker.

MisterRaindrop:
Thank you for the detailed review comments. Regarding the core issues raised (the security impact of partial paths on FDWs that do not support parallel callbacks, the mixing of Gather and CBDB gang models, locus conversion, and the scope of changes in execMain.c), I agree that these are all issues that need to be addressed seriously.

After reconsideration, I am inclined to withdraw the kernel-side modifications and adopt a pure FDW-layer solution instead. The core idea is:

1. Do not modify the kernel's partial path generation, execMain.c, or locus logic—avoiding all the risks mentioned above.
2. The FDW directly uses CBDB's existing parallel variables (ParallelWorkerNumberOfSlice / TotalParallelWorkerNumberOfSlice) to obtain the current worker number and total count.
3. During execution, the FDW calculates the virtual segment ID based on these two values, modifies the HTTP header sent to PXF, and allows PXF's round-robin sharding mechanism to automatically distribute data evenly among all gang workers.

This solution requires no kernel modifications and will not affect other FDWs.

I would like to confirm: Is this direction reasonable? Are the variables ParallelWorkerNumberOfSlice and TotalParallelWorkerNumberOfSlice stable and reliable under the current CBDB parallel framework? Or do you have a more recommended way for the FDW to perceive gang parallel information?

MisterRaindrop:
Among the existing parallel frameworks in CBDB, after FDW registers a partial path via add_partial_path(), can the planner correctly trigger gang expansion and set ParallelWorkerNumberOfSlice? Or does it require additional kernel adaptation?

avamingli:
> Parallel FDW primarily addresses the issue of slow data loading. This functionality was already implemented in earlier versions of PostgreSQL.

Where exactly? Are you referring to [this commit](https://github.com/postgres/postgres/commit/35746bc348b6bf1f690fe17f4f80cfb68e22f504)?

avamingli:
> This solution requires no kernel modifications and will not affect other FDWs.
> 
> I would like to confirm: Is this direction reasonable?

That sounds more reasonable.

avamingli:
> Are the variables ParallelWorkerNumberOfSlice and TotalParallelWorkerNumberOfSlice stable and reliable under the current CBDB parallel framework?

Yes, they are stable and reliable under the current CBDB parallel framework. But I'm not sure how you plan to use them.

> During execution, the FDW calculates the virtual segment ID based on these two values, modifies the HTTP header sent to PXF, and allows PXF's round-robin sharding mechanism to automatically distribute data evenly among all gang workers.

I'm not entirely sure I follow — isn't this essentially how MPP PXF works today?  except `the virtual segment ID based on these two values` -- not sure, off the hand I think it's not enough, different Slice on same Segment could have same parallel workers.

MisterRaindrop:
> > Are the variables ParallelWorkerNumberOfSlice and TotalParallelWorkerNumberOfSlice stable and reliable under the current CBDB parallel framework?
> 
> Yes, they are stable and reliable under the current CBDB parallel framework. But I'm not sure how you plan to use them.
> 
> > During execution, the FDW calculates the virtual segment ID based on these two values, modifies the HTTP header sent to PXF, and allows PXF's round-robin sharding mechanism to automatically distribute data evenly among all gang workers.
> 
> I'm not entirely sure I follow — isn't this essentially how MPP PXF works today? except `the virtual segment ID based on these two values` -- not sure, off the hand I think it's not enough, different Slice on same Segment could have same parallel workers.

Yes, essentially, it reuses the existing MPP round-robin sharding mechanism of PXF—by modifying the segment ID/count in the HTTP header, PXF can distribute data to N×W gang workers instead of N physical segments. No changes are required on the PXF server side.

Regarding ParallelWorkerNumberOfSlice: From the assignment logic in parallel.c, workers on the same segment are assigned incrementally via DSM entry (0, 1, 2, ...), which should be unique. However, I want to confirm: In the CBDB parallel framework, is this value guaranteed to be unique within the same slice on the same segment?

PR #1580, opened 2026-02-23
Remove bogus loop in single-iteration code

reshke:
While casually reading SonarQube output, I spotted this oddly-fashioned code.

### DIFF
--- a/src/backend/access/appendonly/appendonlyam.c
+++ b/src/backend/access/appendonly/appendonlyam.c
@@ -1983,7 +1983,7 @@ appendonly_endscan(TableScanDesc scan)
 static pg_attribute_hot_inline bool
 appendonly_getnextslot_noqual(AppendOnlyScanDesc aoscan, ScanDirection direction, TupleTableSlot *slot)
 {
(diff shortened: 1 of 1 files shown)


src/backend/access/appendonly/appendonlyam.c
```diff
@@ -1983,7 +1983,7 @@ appendonly_endscan(TableScanDesc scan)
 static pg_attribute_hot_inline bool
 appendonly_getnextslot_noqual(AppendOnlyScanDesc aoscan, ScanDirection direction, TupleTableSlot *slot)
 {
-	while (appendonlygettup(aoscan, direction, aoscan->rs_base.rs_nkeys, aoscan->aos_key, slot))
```
yjhjstz, line 1986: if has bloom filter or min-max filter ?

reshke: Sorry, I did not get the actual question. 

My point is, there is no point in loop here?

```
while (cond) {
    f();
    return true;
}
```

isnt this equal to 
```
if (cond) {
    f();
    return true;
}
```

?

yjhjstz: Since `appendonlygettup` has loop,  you are right.

reshke:
Thanks for review.

PR #1586, opened 2026-03-01
Fix grammar in README events calendar description

04cb:
Fixes grammatical error in README where 'subscribe events calendar' should be 'subscribe to the events calendar'.


tuhaihe (approved):

tuhaihe:
Good catch! Thanks @04cb !

PR #1589, opened 2026-03-02
initdb: reset and preserve errno when scanning cdb_init.d

tuhaihe:
initdb's setup_cdb_schema() checked errno after readdir() iteration, but did not clear errno before entering the loop.

On some environments (observed on Ubuntu 24.04), a stale errno value can survive into this check and be misinterpreted as a readdir failure, causing:
  error while reading cdb_init.d directory: Function not implemented

Fix by:
- setting errno=0 before the readdir() loop
- storing readdir errno immediately after the loop
- handling closedir() errors separately
- using the saved readdir errno for the post-loop error check

This prevents false-positive failures during initdb while preserving proper I/O error reporting.

Fixes #ISSUE_Number

### Type of Change
- [ ] Documentation update

### Test Plan

### Impact

**Performance:**

**User-facing changes:**

**Dependencies:**

### Checklist

### CI Skip Instructions

---

### DIFF
--- a/src/bin/initdb/initdb.c
+++ b/src/bin/initdb/initdb.c
@@ -2024,7 +2024,7 @@ setup_cdb_schema(FILE *cmdfd)
 
 	/* Collect all files with .sql suffix in array. */
 	nscripts = 0;
-	while ((file = readdir(dir)) != NULL)
+	while (errno = 0, (file = readdir(dir)) != NULL)
 	{
 		int			namelen = strlen(file->d_name);
 
@@ -2054,12 +2054,16 @@ setup_cdb_schema(FILE *cmdfd)
 		errno = 0;
 #endif
 
-	closedir(dir);
-
 	if (errno != 0)
 	{
-		/* some kind of I/O error? */
 		pg_log_error("error while reading cdb_init.d directory: %m");
+		closedir(dir);
+		exit(1);
+	}
+
+	if (closedir(dir))
+	{
+		pg_log_error("error while closing cdb_init.d directory: %m");
 		exit(1);
 	}
 


src/bin/initdb/initdb.c
```diff
@@ -2054,11 +2056,18 @@ setup_cdb_schema(FILE *cmdfd)
 		errno = 0;
 #endif
 
-	closedir(dir);
+	readdir_errno = errno;
+
+	if (closedir(dir))
```
leborchuk: Yep, here is a bug, One should check errno only if syscall result is -1 https://man7.org/linux/man-pages/man3/errno.3.html

errno should be reset manually only In a rare cases https://stackoverflow.com/questions/25315191/need-to-clean-up-errno-before-calling-function-then-checking-errno 

`if (closedir(dir))` should be enough, but this fragment is quite unusual - we're trying to close the directory anyway. And only then check the reading error. And have lost the reading error if get the closing directory error, because `exit(1)`. How about printing reading error too if we get closing directory error?

tuhaihe: Thanks for the review! I've updated the code to follow the pattern used in  src/bin/pg_rewind/file_ops.c - checking the readdir error first before calling closedir. This ensures no error information is lost. See commit https://github.com/apache/cloudberry/pull/1589/commits/f91837baa26deca5a4aa56174a7ffebf07dbb5e3

src/bin/initdb/initdb.c
avamingli: Good Catch. 
But the current approach still has the same stale errno problem — just from a different source.

With errno = 0 set only once before the loop, consider: if pg_realloc() or pg_strdup() internally sets errno (e.g. via mmap/mremap fallback paths) during, say, the 8th iteration, and subsequent readdir() calls succeed (success does not clear errno), then when the loop ends normally with readdir() returning NULL, errno still holds the stale value from pg_realloc(). The if (errno) check then misinterprets it as a readdir failure — which is the same class of bug this patch is trying to fix.
The comma-expression idiom resets errno before every readdir() call, so when the loop exits, errno can only come from the final readdir():

```c
while (errno = 0, (file = readdir(dir)) != NULL)
```
This is the standard pattern in Postgres.

src/bin/initdb/initdb.c
```diff
@@ -2054,12 +2055,16 @@ setup_cdb_schema(FILE *cmdfd)
...
 	{
-		/* some kind of I/O error? */
 		pg_log_error("error while reading cdb_init.d directory: %m");
+		closedir(dir);
+		exit(1);
+	}
+
+	if (closedir(dir))
+	{
+		pg_log_error("error while closing cdb_init.d directory: %m");
```
avamingli, line 2066: closedir() rarely fails in practice — we have plenty of code that doesn't check its return value — but it's fine to keep the check here, no objection.

src/bin/initdb/initdb.c
tuhaihe: Thanks for your review.

PR #1595, opened 2026-03-02
Few doc & fixes patches from 14.5-14.6 series

reshke:
To make future and in-process rebases easier


reshke:
Closing per device list discussion

PR #1599, opened 2026-03-03
Fix assertion failure in getNdvBySegHeapTuple for empty partitions.

zhangwenchao-123:
When merging leaf partition statistics for a partitioned table, getNdvBySegHeapTuple could hit Assert(valuetype == FLOAT8OID) if a partition had relTuples == 0 and its pg_statistic entry lacked a valid STATISTIC_KIND_NDV_BY_SEGMENTS slot (or the slot had an unexpected element type).

The original guard condition only handled two cases:
  1. Non-empty partition with NDV value == 0
  2. Non-empty partition without NDV_BY_SEGMENTS It missed the case where an empty partition (relTuples == 0) has a pg_statistic entry but no valid FLOAT8OID NDV_BY_SEGMENTS slot, causing the code to fall through to the assertion.

Fix by checking valuetype != FLOAT8OID upfront: skip empty partitions gracefully, and mark non-empty partitions as invalid.

Fixes #ISSUE_Number

### Type of Change
- [ ] Documentation update

### Test Plan

### Impact

**Performance:**

**User-facing changes:**

**Dependencies:**

### Checklist

### CI Skip Instructions

---

### DIFF
--- a/src/backend/commands/analyzeutils.c
+++ b/src/backend/commands/analyzeutils.c
@@ -759,17 +759,38 @@ getNdvBySegHeapTuple(AttStatsSlot * *ndvbsSlots, HeapTuple *heaptupleStats, floa
 		(void) get_attstatsslot(ndvbsSlots[i], heaptupleStats[i],
 			STATISTIC_KIND_NDV_BY_SEGMENTS, InvalidOid, ATTSTATSSLOT_VALUES);
 
-		if ((InvalidOid != ndvbsSlots[i]->valuetype && // result is not empty
-			// not empty partition with invalid ndvbs
-			(relTuples[i] > 0 && DatumGetFloat8(ndvbsSlots[i]->values[0]) == 0)) ||
-			// not empty partition without ndvbs
-			(InvalidOid == ndvbsSlots[i]->valuetype && relTuples[i] > 0)) {
+		if (ndvbsSlots[i]->valuetype != FLOAT8OID)
+		{
+			/*
+			 * NDV_BY_SEGMENTS slot not found or has unexpected type.
+			 * Non-empty partitions must have valid NDV_BY_SEGMENTS;
+			 * empty partitions (relTuples == 0) can be skipped.
+			 */
+			if (relTuples[i] > 0)
+			{
+				valid = false;
+				break;
+			}
+			free_attstatsslot(ndvbsSlots[i]);	
+			pfree(ndvbsSlots[i]);
+			ndvbsSlots[i] = NULL;
+			continue;
+		}
+
+		Assert(ndvbsSlots[i]->valuetype == FLOAT8OID);
+
+		if (ndvbsSlots[i]->nvalues != 1)
+		{
 			valid = false;
 			break;
 		}
 
-		Assert(ndvbsSlots[i]->valuetype == FLOAT8OID);
-		Assert(ndvbsSlots[i]->nvalues == 1);
+		/* Non-empty partition with zero NDV is suspicious */
+		if (relTuples[i] > 0 && DatumGetFloat8(ndvbsSlots[i]->values[0]) == 0)
+		{
+			valid = false;
+			break;
+		}
 	}
 	return valid;
 }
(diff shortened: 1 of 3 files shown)


src/backend/commands/analyzeutils.c
```diff
@@ -759,17 +759,31 @@ getNdvBySegHeapTuple(AttStatsSlot * *ndvbsSlots, HeapTuple *heaptupleStats, floa
...
+			 * NDV_BY_SEGMENTS slot not found or has unexpected type.
+			 * Non-empty partitions must have valid NDV_BY_SEGMENTS;
+			 * empty partitions (relTuples == 0) can be skipped.
+			 */
+			if (relTuples[i] > 0)
+			{
+				valid = false;
+				break;
+			}
+			pfree(ndvbsSlots[i]);
```
my-ship-it, line 775: free_attstatsslot not called before pfree
 
When cleaning up a slot for an empty partition, the code does:
pfree(ndvbsSlots[i]);
ndvbsSlots[i] = NULL;

zhangwenchao-123: fixed

src/backend/commands/analyzeutils.c
```diff
@@ -759,17 +759,32 @@ getNdvBySegHeapTuple(AttStatsSlot * *ndvbsSlots, HeapTuple *heaptupleStats, floa
...
+				valid = false;
+				break;
+			}
+			free_attstatsslot(ndvbsSlots[i]);	
+			pfree(ndvbsSlots[i]);
+			ndvbsSlots[i] = NULL;
+			continue;
 		}
 
-		Assert(ndvbsSlots[i]->valuetype == FLOAT8OID);
```
gfphoenix78, line 771: This assert may keep here.

src/backend/commands/analyzeutils.c
```diff
@@ -759,17 +759,32 @@ getNdvBySegHeapTuple(AttStatsSlot * *ndvbsSlots, HeapTuple *heaptupleStats, floa
...
 		Assert(ndvbsSlots[i]->nvalues == 1);
+
+		/* Non-empty partition with zero NDV is suspicious */
+		if (relTuples[i] > 0 && DatumGetFloat8(ndvbsSlots[i]->values[0]) == 0)
+		{
+			valid = false;
+			break;
+		}
 	}
 	return valid;
```
gfphoenix78, line 795: All of the ndvbsSlots could be release if the valid becomes false. On the other side, the ndvbsSlots and its pointer array could be also freed in aggregate_leaf_partition_ndvbs no matter it's valid or not.

src/backend/commands/analyzeutils.c
```diff
@@ -759,17 +759,32 @@ getNdvBySegHeapTuple(AttStatsSlot * *ndvbsSlots, HeapTuple *heaptupleStats, floa
...
+			}
+			free_attstatsslot(ndvbsSlots[i]);	
+			pfree(ndvbsSlots[i]);
+			ndvbsSlots[i] = NULL;
+			continue;
 		}
 
-		Assert(ndvbsSlots[i]->valuetype == FLOAT8OID);
 		Assert(ndvbsSlots[i]->nvalues == 1);
+
```
yjhjstz, line 787:
```c++
if (ndvbsSlots[i]->nvalues != 1)
  {
      valid = false;
      break;
  }
```

src/backend/commands/analyzeutils.c
zhangwenchao-123: Good point. Added back `Assert(ndvbsSlots[i]->valuetype == FLOAT8OID)` after the type check block. Also changed `Assert(ndvbsSlots[i]->nvalues == 1)` to a runtime check as yjhjstz suggested.

src/backend/commands/analyzeutils.c
zhangwenchao-123: Right, `aggregate_leaf_partition_ndvbs` will free all ndvbsSlots regardless of validity. The early cleanup here for empty partitions (setting slot to NULL after free) is just to keep the slot array consistent during iteration — it won't cause a double-free since the cleanup function checks for NULL.

yjhjstz (approved):

my-ship-it (approved):

PR #1609, opened 2026-03-11
Fix compliance issues for RPM and DEB packages

tuhaihe:
As an Apache incubating project, convenience binaries must include LICENSE, NOTICE, and DISCLAIMER files. This commit adds these mandatory compliance files into the spec and rules definitions to ensure they are properly distributed with the binary RPM and DEB packages.

Additionally, this commit:
- Automates the cp of the .spec file into the ~/rpmbuild tree to prevent build failures for new users.

Fixes #ISSUE_Number

### Type of Change
- [ ] Documentation update

### Test Plan

### Impact

**Performance:**

**User-facing changes:**

**Dependencies:**

### Checklist

### CI Skip Instructions

---

### DIFF
--- a/devops/build/packaging/deb/build-deb.sh
+++ b/devops/build/packaging/deb/build-deb.sh
@@ -109,7 +109,7 @@ export CBDB_FULL_VERSION=$VERSION
 
 # Set version if not provided
 if [ -z "${VERSION}" ]; then
-  export CBDB_FULL_VERSION=$(./getversion | cut -d'-' -f 1 | cut -d'+' -f 1)
+  export CBDB_FULL_VERSION=$(./getversion 2>/dev/null | cut -d'-' -f 1 | cut -d'+' -f 1 || echo "unknown")
 fi
 
 if [[ ! $CBDB_FULL_VERSION =~ ^[0-9] ]]; then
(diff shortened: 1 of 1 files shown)


devops/build/packaging/deb/ubuntu22.04/rules
```diff
@@ -19,7 +19,22 @@ include /usr/share/dpkg/default.mk
...
+		cp -a ${CBDB_BIN_PATH}/* ${DEBIAN_DESTINATION}/; \
+	else \
+		echo "Build staging directory already populated, skipping copy."; \
+	fi
+	# Copy Apache compliance files into the build staging directory
+	cp -a LICENSE NOTICE DISCLAIMER ${DEBIAN_DESTINATION}/
+	cp -a licenses ${DEBIAN_DESTINATION}/
+	# Create debian/copyright for Debian policy compliance
+	mkdir -p $(shell pwd)/debian
+	cat LICENSE NOTICE > $(shell pwd)/debian/copyright
```
leborchuk, line 37: have lost here 

`	make install DESTDIR=${DEBIAN_DESTINATION} prefix=`

We need it to install all the binaries, without it we package only license files

tuhaihe: Thanks for catching this! The `make install` was replaced with a conditional copy mechanism, not simply removed. Here's the updated logic (lines 22–31 in the new rules file):

```makefile
@mkdir -p ${DEBIAN_DESTINATION}
@if [ -z "$$(ls -A ${DEBIAN_DESTINATION} 2>/dev/null)" ]; then \
    echo "Copying pre-built binaries from ${CBDB_BIN_PATH} to ${DEBIAN_DESTINATION}..."; \
    cp -a ${CBDB_BIN_PATH}/* ${DEBIAN_DESTINATION}/; \
else \
    echo "Build staging directory already populated, skipping copy."; \
fi
```

How it works:

- In CI: The build workflow already sets `BUILD_DESTINATION=${SRC_DIR}/debian/build`, so the configure and build scripts install all binaries directly into `debian/build/`. When `gpinstall` runs, the directory is already populated → skip the copy.
- For local/manual packaging: If a developer has already compiled and installed Cloudberry to `/usr/local/cloudberry-db`, they can run `build-deb.sh`
 directly. The staging directory will be empty → binaries are copied from the installed path.

The original `make install DESTDIR=... prefix=` would re-run the install step from the source tree, which requires the full build state to still be present. The new approach is more flexible and avoids redundant work in both scenarios.

All binaries are still fully included in the package — only the mechanism for populating the staging directory changed.

leborchuk: Got it! We have a script named build-cloudberry.sh but actually do `make` && `make install` inside it. So there is no need to install binaries once again

tuhaihe:
After this change, we can generate the PRM/DEB packages just by one command:

```bash
./devops/build/packaging/deb/build-deb.sh -v 2.1.0 # DEB


./devops/build/packaging/deb/build-deb.sh -v 2.1.0 -r 1 # RPM
```

Note: You must build the Cloudberry before running the above commands.

PR #1610, opened 2026-03-11
cherrypick from main to REL_2_STABLE part2

reshke:
cherrypick from commit: https://github.com/apache/cloudberry/commit/017b06b4f8698c47fa9f681103322aa85f167648 to commit: https://github.com/apache/cloudberry/commit/8ecc3267e4d1a7d73a8d4c4e28d2073162ba453d


reshke:
Hmm, looks like there is some unwanted changes to catalog (system_functions.sql), need to remove this

PR #1085, opened 2025-05-06
Introduce extension perfmon to monitor database cluster

fanfuxiaoran:
Fixes #ISSUE_Number

### What does this PR do?

This PR introduces the ​​perfmon​​ extension, designed to monitor the Cloudberry Database.

#### Key Features:
- ​**Cluster-wide Metrics Collection​​**: Gathers performance metrics from each machine in the database cluster.
- **Real-time Query Monitoring​​:**
   - Displays active queries with detailed execution plans.
   - Tracks progress of each query plan node (similar to EXPLAIN ANALYZE for running queries).
   - Captures performance metrics (CPU, memory, spill files) across multiple segments.
- ​​**Query History​​**: Provides a historical record of executed queries.

#### Components:
- **gpmmon**​​ (Shared Library)
Added to shared_preload_libraries for background operation.
Runs as a backgroud worker process to collect and store metrics in the gpperfmon database.
- **gpmon**​​ (Shared Library)
Added to shared_preload_libraries for query monitoring.
Implements hooks to capture runtime query data.
Provides the pg_query_state() function to inspect live query execution status.
- **gpsmon**​​ (Binary)
Deployed per-instance to collect system-level metrics (CPU, memory, etc.).

image
### Type of Change
- [ ] Bug fix (non-breaking change)
- [x] New feature (non-breaking change)
- [ ] Breaking change (fix or feature with breaking changes)
- [ ] Documentation update

### Test Plan

- [ ] Unit tests added/updated
- [x] Integration tests added/updated
- [x] Passed `make installcheck`
- [ ] Passed `make -C src/test installcheck-cbdb-parallel`

### Impact

**Performance:**

**User-facing changes:**

**Dependencies:**

### Checklist
- [ ] Followed [contribution guide](https://cloudberry.apache.org/contribute/code)
- [ ] Added/updated documentation
- [ ] Reviewed code for security implications
- [ ] Requested review from [cloudberry committers](https://github.com/orgs/apache/teams/cloudberry-committers)

### CI Skip Instructions

---

PR #1629, opened 2026-03-18
Add new extension yagp hooks collector

leborchuk:
A diagnostic and monitoring extension for Cloudberry clusters

Briefly, the interaction scheme is:

`Cloudberry` (Run Executor hooks) -> `hooks collector extension` (Create protobuf messages and send them using UDS) -> `local yagpcc agent`

This is the part that sends data from the PG instance to an external agent.

External agent receives telemetry, store, aggregate and expose to the consumers. The agent source code is  https://github.com/open-gpdb/yagpcc We're going to propose add it to Apache Cloudberry infrastructure later, after fixing all the tests.

This is quite similar to the https://github.com/apache/cloudberry/pull/1085 idea.  However, it has a different implementation. It is completely separate from database application and written in Go. Could be separately maintained. No influence to the main database, and so different development requirements. 

Extension has pg_regress tests, they were copied from pg_stat_statements (PG18) and adjusted to the extension behaviour.

Original extension authors are [Smyatkin-Maxim](https://github.com/Smyatkin-Maxim) and [NJrslv](https://github.com/NJrslv) I left their commits as-is to have blame history commited source.

### DIFF
--- a/devops/build/automation/cloudberry/scripts/configure-cloudberry.sh
+++ b/devops/build/automation/cloudberry/scripts/configure-cloudberry.sh
@@ -53,7 +53,6 @@
 #
 # Optional Environment Variables:
 #   LOG_DIR - Directory for logs (defaults to ${SRC_DIR}/build-logs)
-#   CONFIGURE_EXTRA_OPTS - Args to pass to configure command
 #   ENABLE_DEBUG - Enable debug build options (true/false, defaults to
 #                  false)
 #
@@ -179,8 +178,7 @@ execute_cmd ./configure --prefix=${BUILD_DESTINATION} \
             --with-uuid=e2fs \
             ${CONFIGURE_MDBLOCALES_OPTS} \
             --with-includes=/usr/local/xerces-c/include \
-            --with-libraries=${BUILD_DESTINATION}/lib \
-            ${CONFIGURE_EXTRA_OPTS:-""} || exit 4
+            --with-libraries=${BUILD_DESTINATION}/lib || exit 4
 log_section_end "Configure"
 
 # Capture version information


leborchuk:
Right now we are not fully support all the features from https://github.com/apache/cloudberry/pull/1085 but I believe one day solution will be won't be any worse.

pom.xml
```diff
@@ -154,6 +154,12 @@ code or new licensing patterns.
             <exclude>gpcontrib/gp_exttable_fdw/gp_exttable_fdw.control</exclude>
 
             <exclude>gpcontrib/diskquota/**</exclude>
+            <exclude>gpcontrib/yagp_hooks_collector/yagp_hooks_collector.control</exclude>
+            <exclude>gpcontrib/yagp_hooks_collector/protos/yagpcc_set_service.proto</exclude>
+            <exclude>gpcontrib/yagp_hooks_collector/protos/yagpcc_plan.proto</exclude>
+            <exclude>gpcontrib/yagp_hooks_collector/protos/yagpcc_metrics.proto</exclude>
+            <exclude>gpcontrib/yagp_hooks_collector/.clang-format</exclude>
+            <exclude>gpcontrib/yagp_hooks_collector/Makefile</exclude>
```
tuhaihe, line 157: can move these new lines to the `Cloudberry` part, like:

https://github.com/apache/cloudberry/blob/6c37f8f6732bbe18387b5500838409153793ec13/pom.xml#L1263-L1272

gpcontrib/yagp_hooks_collector/src/stat_statements_parser/README.md
```diff
@@ -0,0 +1 @@
+This directory contains a slightly modified subset of pg_stat_statements for PG v9.4 to be used in query and plan ID generation.
```
tuhaihe: Can add the standard ASF license header to the markdown file, like: https://github.com/apache/cloudberry/blob/6c37f8f6732bbe18387b5500838409153793ec13/CODE_OF_CONDUCT.md?plain=1#L1-L18

.github/workflows/build-cloudberry.yml
```diff
@@ -271,6 +271,10 @@ jobs:
                 },
                 "enable_core_check":false
               },
+              {"test":"gpcontrib-yagp-hooks-collector",
+               "make_configs":["gpcontrib/yagp_hooks_collector:installcheck"],
+               "extension":"yagp_hooks_collector"
+              },
```
tuhaihe, line 277: Is there a need to add the new test to `build-cloudberry-rocky8.yml` or `*.*-deb.yml` file? FYI.

gpcontrib/yagp_hooks_collector/protos/yagpcc_metrics.proto
```diff
@@ -0,0 +1,185 @@
+syntax = "proto3";
+
+package yagpcc;
+option java_outer_classname = "SegmentYAGPCCM";
+option go_package = "a.yandex-team.ru/cloud/mdb/yagpcc/api/proto/common;greenplum";
```
tuhaihe: Line 5: Is it private？ If so, do we need to make it public?

gpcontrib/yagp_hooks_collector/src/stat_statements_parser/README.md
tuhaihe: Can add an ASF license header. FYI.

gpcontrib/yagp_hooks_collector/metric.md
```diff
@@ -0,0 +1,126 @@
+## YAGP Hooks Collector Metrics
```
tuhaihe: Can add an ASF licenser header too. FYI.

gpcontrib/yagp_hooks_collector/README.md
```diff
@@ -0,0 +1,28 @@
+## YAGP Hooks Collector
+
+An extension for collecting greenplum query execution metrics and reporting them to an external agent.
```
tuhaihe:
```suggestion
An extension for collecting Apache Cloudberry/Greenplum query execution metrics and reporting them to an external agent.
```

gpcontrib/yagp_hooks_collector/README.md
```diff
@@ -0,0 +1,28 @@
...
+#### 2. `EXPLAIN` data
+-   **What:** Triggers generation of the `EXPLAIN (TEXT, COSTS, VERBOSE)` and captures it.
+-   **GUC:** `yagpcc.enable`.
+
+#### 3. `EXPLAIN ANALYZE` data
+-   **What:** Triggers generation of the `EXPLAIN (TEXT, ANALYZE, BUFFERS, TIMING, VERBOSE)` and captures it.
+-   **GUCs:** `yagpcc.enable`, `yagpcc.min_analyze_time`, `yagpcc.enable_cdbstats`(ANALYZE), `yagpcc.enable_analyze`(BUFFERS, TIMING, VERBOSE).
+
+#### 4. Other Metrics
+-   **What:** Captures Instrument, Greenplum, System, Network, Interconnect, Spill metrics.
```
tuhaihe:
```suggestion
-   **What:** Captures Instrument, Apache Cloudberry (Greenplum), System, Network, Interconnect, Spill metrics.
```

gpcontrib/yagp_hooks_collector/yagp_hooks_collector.control
```diff
@@ -0,0 +1,5 @@
+# yagp_hooks_collector extension
+comment = 'Intercept query and plan execution hooks and report them to Yandex GPCC agents'
```
tuhaihe:
```suggestion
comment = 'Intercept query and plan execution hooks and report them to Cloudberry monitor agents'
```
?

configure.ac
```diff
@@ -1365,6 +1365,13 @@ PGAC_ARG_BOOL(with, zstd, yes, [do not build with Zstandard],
 AC_MSG_RESULT([$with_zstd])
 AC_SUBST(with_zstd)
 
+#
+# yagp_hooks_collector
+#
+PGAC_ARG_BOOL(with, yagp_hooks_collector, no,
+              [build with YAGP hooks collector extension])
```
tuhaihe:
```suggestion
              [build with hooks collector extension])
```
?

Makefile
```diff
@@ -3,14 +3,12 @@
 # to build Postgres with a different make, we have this make file
 # that, as a service, will look for a GNU make and invoke it, or show
 # an error message if none could be found.
-
```
tuhaihe, line 6: If no actual changes to Makefile, we can restore it.

configure.ac
```diff
@@ -1365,6 +1365,13 @@ PGAC_ARG_BOOL(with, zstd, yes, [do not build with Zstandard],
 AC_MSG_RESULT([$with_zstd])
 AC_SUBST(with_zstd)
 
+#
+# yagp_hooks_collector
```
tuhaihe: Does this new extension need other dependencies? Not sure about this. If so, we can add the pre-required dependencies here. FYI.

tuhaihe (commented):
Thanks for your great work! Just left some comments, FYI.

Jaroslavian:
Hi all!

I want to extend the description of this project a bit.

Here we collect state of the query during the executor stages and send it via unix sockets. It was originally made for Greenplum, and I adapted it for Cloudberry [here](https://github.com/open-gpdb/cloudberry/pull/21).

For testing purposes only, we support writing data to the catalog table so it can be selected during regression tests. The tuples are logged even during transaction aborts, using frozen tuples (only for the tests).

The state of the query is created before the executor_start stage and released after the executor_end stage. If the executor_end is missing for some unknown reason, the state is not freed. To fix this problem, new changes to the code outside of the project were introduced to correctly handle queries with a missing executor_end stage.

configure.ac
Jaroslavian: We need the protoc, included it.

.github/workflows/build-cloudberry.yml
Jaroslavian: Done - I don't see a reason to not include it.

pom.xml
Jaroslavian: Done

gpcontrib/yagp_hooks_collector/src/stat_statements_parser/README.md
Jaroslavian: Done

gpcontrib/yagp_hooks_collector/protos/yagpcc_metrics.proto
Jaroslavian: I don't think we need them, so I just got rid of those options.

gpcontrib/yagp_hooks_collector/src/stat_statements_parser/README.md
Jaroslavian: done

gpcontrib/yagp_hooks_collector/metric.md
Jaroslavian: done

gpcontrib/yagp_hooks_collector/README.md
Jaroslavian: Just got rid of it - done

gpcontrib/yagp_hooks_collector/README.md
Jaroslavian: done

gpcontrib/yagp_hooks_collector/yagp_hooks_collector.control
Jaroslavian: done

configure.ac
Jaroslavian: done

Makefile
Jaroslavian: done

gpcontrib/gp_stats_collector/gmon.out
tuhaihe, line 1: Do we still need this binary file? FYI.

Jaroslavian: We don't - done.

tuhaihe:
> LGTM, though I'm kind of an interested party myself - so feel free to ignore my ship :)
> 
> It's been widely used in all our production gp6 clusters and in terms of production-readiness we're quite confident in it and it's way better than gpperfmon. I guess as a contrib package it's should be good enough for cloudberry as well.

Great! I also believe it so. I would like to test on a local machine and give my +1 earlier next week. Does it sound good?

configure
tuhaihe, line 1: Hi, I tested on a Rocky 8 machine -- it shows we still need to regenerate this file again based on the latest `configure.ac`. FYI.

.github/workflows/build-deb-cloudberry.yml
```diff
@@ -448,13 +452,14 @@ jobs:
         shell: bash
         env:
           SRC_DIR: ${{ github.workspace }}
+          CONFIGURE_EXTRA_OPTS: --with-gp-stats-collector
```
tuhaihe: Hi, just one idea on how to enable this new extension by default: could we add the option `--with-gp-stats-collector` directly in the `configure-cloudberry.sh` file? Maybe this way is simpler? We just need to add the new test here. FYI.

tuhaihe:
Hi, I did a test on a local machine. After building the extension successfully, it showed that some files are deleted when running `make distclean`:

```
[gpadmin@cdw cloudberry]$ git status
On branch mergeYagpHooks
Changes not staged for commit:
  (use "git add/rm <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)

        deleted:    gpcontrib/gp_stats_collector/results/gpsc_cursors.out
        deleted:    gpcontrib/gp_stats_collector/results/gpsc_dist.out
        deleted:    gpcontrib/gp_stats_collector/results/gpsc_guc_cache.out
        deleted:    gpcontrib/gp_stats_collector/results/gpsc_locale.out
        deleted:    gpcontrib/gp_stats_collector/results/gpsc_select.out
        deleted:    gpcontrib/gp_stats_collector/results/gpsc_uds.out
        deleted:    gpcontrib/gp_stats_collector/results/gpsc_utf8_trim.out
        deleted:    gpcontrib/gp_stats_collector/results/gpsc_utility.out
```

Do we need to remove or keep them in the source? FYI.

configure
tuhaihe: I run `autoconf` in Rocky Linux 8.

gpcontrib/gp_stats_collector/src/UDSConnector.cpp
```diff
@@ -0,0 +1,130 @@
...
+  const auto &uds_path = config.uds_path();
+
+  if (uds_path.size() >= sizeof(address.sun_path)) {
+    ereport(WARNING, (errmsg("UDS path is too long for socket buffer")));
+    GpscStat::report_error();
+    return false;
+  }
+  strcpy(address.sun_path, uds_path.c_str());
+
+  const auto sockfd = socket(AF_UNIX, SOCK_STREAM, 0);
```
my-ship-it: Thanks for the contribution. It is great but minor comments here:
 
Every single hook event creates a new socket, connects, sends, and closes. For a busy cluster this means thousands of socket()/connect()/close() syscalls per second. This will be a significant performance hit. Should use a persistent connection with reconnect-on-failure.

gpcontrib/gp_stats_collector/src/hook_wrappers.cpp
```diff
@@ -0,0 +1,414 @@
...
+  }
+  return sender;
+}
+
+template <typename T, typename R, typename... Args>
+R cpp_call(T *obj, R (T::*func)(Args...), Args... args) {
+  try {
+    return (obj->*func)(args...);
+  } catch (const std::exception &e) {
+    ereport(FATAL, (errmsg("Unexpected exception in gpsc %s", e.what())));
```
my-ship-it: Any C++ exception from the extension kills the entire backend with FATAL. This should be ERROR or WARNING at most — a monitoring extension should never crash the database.

gpcontrib/gp_stats_collector/src/memory/gpdbwrappers.h
```diff
@@ -0,0 +1,81 @@
...
+ *
+ * gpdbwrappers.h
+ *
+ * IDENTIFICATION
+ *	  gpcontrib/gp_stats_collector/src/memory/gpdbwrappers.h
+ *
+ *-------------------------------------------------------------------------
+ */
+
+#pragma once
```
my-ship-it: CBDB convention requires #ifndef FILENAME_H / #define FILENAME_H style include guards, not #pragma once.

gpcontrib/gp_stats_collector/.clang-format
```diff
@@ -0,0 +1,2 @@
+BasedOnStyle: LLVM
+SortIncludes: false
```
my-ship-it: The extension uses LLVM code style (brace on same line, etc.) which is inconsistent with the rest of CBDB (BSD/Allman style). This will cause confusion and merge conflicts if other contributors touch these files.

gpcontrib/gp_stats_collector/protos/gpsc_metrics.proto
```diff
@@ -0,0 +1,183 @@
+syntax = "proto3";
```
my-ship-it, line 19: No Apache license header on .proto files — The protobuf definition files lack the Apache 2.0 license header that other files have.

gpcontrib/gp_stats_collector/src/Config.cpp
```diff
@@ -0,0 +1,177 @@
...
+}
+
+void Config::init_gucs() {
+  DefineCustomStringVariable(
+      "gpsc.uds_path", "Sets filesystem path of the agent socket", 0LL,
+      &guc_uds_path, "/tmp/gpsc_agent.sock", PGC_SUSET,
+      GUC_NOT_IN_SAMPLE | GUC_GPDB_NEED_SYNC, 0LL, 0LL, 0LL);
+
+  DefineCustomBoolVariable(
+      "gpsc.enable", "Enable metrics collector", 0LL, &guc_enable_collector,
```
my-ship-it: When the extension is loaded via shared_preload_libraries, collection is ON by default. Combined with gpsc.ignored_users_list defaulting to "gpadmin,repl,gpperfmon,monitor", this means the extension silently starts collecting for non-default users. Default should be false for safety？

gpcontrib/gp_stats_collector/results/gpsc_cursors.out
```diff
@@ -0,0 +1,163 @@
+CREATE EXTENSION gp_stats_collector;
```
my-ship-it: Remove results/ from git？

gpcontrib/gp_stats_collector/src/UDSConnector.cpp
Jaroslavian: Hi!

These are AF_UNIX sockets (Unix sockets), not TCP. The networking stack is not involved, Unix domain sockets use in-kernel memory copies between processes, not the network stack path.

I think the cost of socket()/connect()/close() on AF_UNIX is comparable to open()/write()/close() on a local file (we actually have a file to do IPC).

Do you mean that even with AF_UNIX option there is a potential problem? Also, we were running regression tests with this extension enabled and haven't seen any regressions.

.github/workflows/build-deb-cloudberry.yml
Jaroslavian: My reasoning was not to build this extension for users who don't know about it. Including `--with-gp-stats-collector` in `configure-cloudberry.sh` makes it build by default.

Do you think it's better to always build our extension or build only for the tests (current version)?

tuhaihe: Thanks for the reply! I’d like to share a few perspectives on why including `--with-gp-stats-collector` in `configure-cloudberry.sh` might be a better long-term choice:

1. CI/CD Stability & Consistency: Since `configure-cloudberry.sh` is the foundational script for our PR generation, CI pipelines, and developer environments, adding the flag here ensures that the extension is consistently compiled and verified with every code change. This prevents "silent" build breakages that might only appear during specific tests.
2. Building vs. Enabling (Zero Runtime Impact): There is a clear distinction between `building` the binary and `enabling` the feature. As long as `shared_preload_libraries` doesn't include it and `gpsc.enable` defaults to off, the extension is essentially "dormant" on disk. It will have zero impact on database performance, memory, or stability for users who aren't aware of it.
3. Avoid the "Recompilation Hurdle": Cloudberry is a massive project. If a user discovers this powerful monitoring tool later but finds it wasn't built into their binary, they would have to reconfigure and recompile the entire database. This "secondary build" requirement is a significant barrier to adoption.
4. Alignment with Other Extensions: Following the precedent of `diskquota` or `PAX`, building by default significantly improves discoverability. It allows users to try this high-performance collector via a simple `CREATE EXTENSION` without needing a deep knowledge of build flags.

Finally, for those power users with highly customized requirements, they are already accustomed to modifying `configure` options or scripts to suit their specific needs. We’ve already provided that flexibility, so we don't need to worry about them. For the majority of users, keeping it "built-in but dormant" provides the best "out-of-the-box" experience. So let's enable it by default?

gpcontrib/gp_stats_collector/src/Config.cpp
NJrslv: I think from the user's perspective, once `gp_stats_collector` is loaded via `shared_preload_libraries`, they want it enabled by default. We exclude noisy users (`gpadmin`, `repl`, `gpperfmon`, `monitor`) to avoid log overflow (not in terms of space, but for the user to easily see what is happening). But it is also possible that the user logged in as some ignored_user, for this reason, I'd clear by default ignored_users.

What do I think about this, 2 different approaches: 
1. If not built by default → enable via GUCs by default (users loading it expect it active) and clear ignored_users.
2. If built by default → disable all activating GUCs and keep ignored users as right now.

What do you think about this?

.github/workflows/build-deb-cloudberry.yml
Jaroslavian: Sure, it makes total sense for me now, thanks.

configure
Jaroslavian: I didn't know that, thanks.

As far as I understand it, `configure` is generated from `configure.ac`.

If that's the case, we have a problem with `configure`: every commit that modifies `configure` without touching `configure.ac` gets overwritten after running autoconf.

For example, the latest commit has this issue: https://github.com/apache/cloudberry/commit/7e867f605c8798979405b163a7d5c65b572edb4c.

If I am not wrong, the solution would be, in this pr, to restore any changes to `configure`, create a separate pr, where we find every commit that might be overwritten by autoconf (the change is not present in `configure.ac`). We can do it in parallel, does not matter, just regenerate the `configure` in its dedicated pr.

tuhaihe: Good catch. Let me verify the commit 7e867f6 later. If it needs a fix, will create a separate PR. Then our PR can rebase.

tuhaihe: After a review, not only 7e867f6 but also https://github.com/apache/cloudberry/pull/1363 introduced a lot of changes to the configure -- resulting in regenerating a new configure file. Let me create a new PR on it. You can continue on this PR, then rebase.

gpcontrib/gp_stats_collector/results/gpsc_cursors.out
Jaroslavian: Done https://github.com/apache/cloudberry/pull/1629/commits/829ee53ef1b42ec481a9b495832a989666baa647

gpcontrib/gp_stats_collector/src/Config.cpp
Jaroslavian: Done https://github.com/apache/cloudberry/pull/1629/commits/ab4a54e6a6b22f347da5eb144c3ac716671cf73b

gpcontrib/gp_stats_collector/.clang-format
Jaroslavian: Done https://github.com/apache/cloudberry/pull/1629/commits/cf6ffce4b91964fd3149088c81935e56efc2bbc6

gpcontrib/gp_stats_collector/src/memory/gpdbwrappers.h
Jaroslavian: Done https://github.com/apache/cloudberry/pull/1629/commits/925de3f15796c9e381725a2ae3d6130b8fea6aca

gpcontrib/gp_stats_collector/src/hook_wrappers.cpp
Jaroslavian: Done https://github.com/apache/cloudberry/pull/1629/commits/c1ade83be919f2a0a6119b96fc4ba9de7015257e

.github/workflows/build-deb-cloudberry.yml
Jaroslavian: Done here https://github.com/apache/cloudberry/pull/1629/commits/ab4a54e6a6b22f347da5eb144c3ac716671cf73b

gpcontrib/gp_stats_collector/src/UDSConnector.cpp
my-ship-it: Thanks for the explanation. You're right that AF_UNIX sockets are significantly cheaper than TCP — the kernel memory copy path avoids most of the networking overhead I was concerned about.

I'm fine with the current implementation. If performance ever becomes a concern under very high QPS workloads, a persistent connection could be revisited as a future optimization. No action needed for now.

src/backend/commands/portalcmds.c
```diff
@@ -373,6 +374,10 @@ PortalCleanup(Portal portal)
 			FreeQueryDesc(queryDesc);
 
 			CurrentResourceOwner = saveResourceOwner;
+		} else {
```
my-ship-it: 1) Minor style nit: } else { does not follow CBDB's brace convention.

2) As comments says
```
 Note: if portal->status is PORTAL_FAILED, we are probably being called
 during error abort, and must be careful to avoid doing anything that
 is likely to fail again.
```

The backend may be in an unstable state during exception recovery. Calling extension code that does socket I/O and protobuf serialization here risks nested exceptions or hangs.

I'd suggest wrapping the hook call in its own PG_TRY/PG_CATCH block to isolate failures:
```
  else
  {
        /* GPDB hook for collecting query info */
        if (queryDesc->gpsc_query_key && query_info_collect_hook)
        {
                PG_TRY();
                {
                        (*query_info_collect_hook)(METRICS_QUERY_ERROR, queryDesc);
                }
                PG_CATCH();
                {
                        FlushErrorState();
                }
                PG_END_TRY();
        }
  }
```
This way, if the hook fails, the error is silently discarded and normal cleanup proceeds.

gpcontrib/gp_stats_collector/src/UDSConnector.cpp
```diff
@@ -0,0 +1,146 @@
...
+		sent = send(sockfd, buf + sent_total, total_size - sent_total,
+					MSG_DONTWAIT);
+		if (sent > 0)
+			sent_total += sent;
+	} while (sent > 0 && size_t(sent_total) != total_size &&
+			 // the line below is a small throttling hack:
+			 // if a message does not fit a single packet, we take a nap
+			 // before sending the next one.
+			 // Otherwise, MSG_DONTWAIT send might overflow the UDS
+			 (std::this_thread::sleep_for(std::chrono::milliseconds(1)), true));
```
my-ship-it: std::this_thread::sleep_for does not interact with PostgreSQL's signal handling infrastructure. During a backend shutdown or query cancel, this sleep will not be interrupted by CHECK_FOR_INTERRUPTS(). How about use pg_usleep() or WaitLatch() instead?

gpcontrib/gp_stats_collector/gp_stats_collector--1.0--1.1.sql
```diff
@@ -0,0 +1,113 @@
...
+DROP FUNCTION __gpsc_stat_messages_f_on_segments();
+DROP FUNCTION __gpsc_stat_messages_f_on_master();
+DROP FUNCTION __gpsc_stat_messages_reset_f_on_segments();
+DROP FUNCTION __gpsc_stat_messages_reset_f_on_master();
+
+-- Recreate functions and view in new schema.
+CREATE FUNCTION gpsc.__stat_messages_reset_f_on_master()
+RETURNS SETOF void
+AS 'MODULE_PATHNAME', 'gpsc_stat_messages_reset'
+LANGUAGE C EXECUTE ON MASTER;
```
my-ship-it: All SQL definition files (gp_stats_collector--1.0.sql, --1.1.sql, --1.0--1.1.sql) use EXECUTE ON MASTER. Cloudberry has moved to EXECUTE ON COORDINATOR. While the old syntax still works, better for new extensions to use the current terminology.

gpcontrib/gp_stats_collector/src/hook_wrappers.cpp
```diff
@@ -0,0 +1,473 @@
...
+cpp_call(T *obj, R (T::*func)(Args...), Args... args)
+{
+	try
+	{
+		return (obj->*func)(args...);
+	}
+	catch (const std::exception &e)
+	{
+		ereport(ERROR, (errmsg("Unexpected exception in gpsc %s", e.what())));
+	}
```
my-ship-it, line 120: The catch block calls ereport(ERROR, ...) which does not return, but the compiler doesn't know that.

When R is non-void, this is undefined behavior (missing return after catch). Add pg_unreachable() or __builtin_unreachable() after the ereport?

src/backend/commands/portalcmds.c
Jaroslavian: Good catch, done - https://github.com/apache/cloudberry/pull/1629/changes/4271712a94540d3223b8b286a8d1b183d5c17436

gpcontrib/gp_stats_collector/gp_stats_collector--1.0--1.1.sql
Jaroslavian: Done - https://github.com/apache/cloudberry/pull/1629/changes/733b0822415970e25ea3601af808b94f3152e692

gpcontrib/gp_stats_collector/src/hook_wrappers.cpp
Jaroslavian: Done - https://github.com/apache/cloudberry/pull/1629/changes/826d97d08c4c938d2d84032dcb4fb11e0e513516

gpcontrib/gp_stats_collector/src/UDSConnector.cpp
Jaroslavian: Partially done here - https://github.com/apache/cloudberry/pull/1629/changes/649da13fe7f06998d4e38e199884f67959ff2865.

What I am thinking here is that we need to use `CHECK_FOR_INTERRUPTS()` along with `pg_usleep()`, but `CHECK_FOR_INTERRUPTS()` internally might call `longjmp()` bypassing destructor call of C++ objects like `buf_guard` or `sock_guard` or some other object on the stack, from some previous outer function.

Using `WaitLatch()` seems like overkill for a 1ms sleep, so I just replaced `sleep_for()` with `pg_usleep()`.

my-ship-it (approved):
LGTM.

Great work on this extension — I can see a lot of effort went into it, and the earlier
review feedback has been well addressed.

Before merging, it might be nice to squash or tidy up the commit history a bit. The current log includes quite a few incremental fix-up commits from the review process.

Squashing these into a small number of logical commits — or even a single clean commit — would keep git log tidy and consistent with the project's commit conventions. Just a suggestion!

Other than these, everything looks good from my side. Thanks again for the great contribution!

devops/build/automation/cloudberry/scripts/configure-cloudberry.sh
```diff
@@ -177,7 +179,8 @@ execute_cmd ./configure --prefix=${BUILD_DESTINATION} \
             --with-uuid=e2fs \
             ${CONFIGURE_MDBLOCALES_OPTS} \
             --with-includes=/usr/local/xerces-c/include \
-            --with-libraries=${BUILD_DESTINATION}/lib || exit 4
+            --with-libraries=${BUILD_DESTINATION}/lib \
+            ${CONFIGURE_EXTRA_OPTS:-""} || exit 4
```
tuhaihe: It looks like this new option is no longer used. Can remove it. FYI.

tuhaihe: Same to Line 56.

configure
tuhaihe: The fix PR is still in progress. So let's merge this PR first. We can regenerate the `configure` file later.

devops/build/automation/cloudberry/scripts/configure-cloudberry.sh
Jaroslavian: Done - https://github.com/apache/cloudberry/pull/1629/commits/4509d30308a68ec2f90ea7084827772f8d2fa587

Jaroslavian:
@tuhaihe
Dianjin, Hi!

I restored the config file here (the one I was editing manually): https://github.com/apache/cloudberry/pull/1629/changes/f26c2877f03db0f0712f25770a169b35c17c4318. We'll regenerate it cleanly from configure.ac in the next PRs, I guess.

I think we're now good to merge.

The only question is: squash or merge? This PR has quite a history of commits, so it might be useful to preserve it. But throughout our work on this extension, we've been making fix-up commits, and the history should be cleaned up - grouped before the merge.

I don't think preserving the commit history is an extremely important option. The cleanest and easiest approach would be to squash all the commits into one and merge. What do you think?

*Also, give me some time to write a complete commit message, if we squash it.

tuhaihe:
> @tuhaihe Dianjin, Hi!
> 
> I restored the config file here (the one I was editing manually): [f26c287](https://github.com/apache/cloudberry/commit/f26c2877f03db0f0712f25770a169b35c17c4318). We'll regenerate it cleanly from configure.ac in the next PRs, I guess.
> 

Got it, thanks. Yes, we can regenerate it later, and welcome to create a new PR to update it then. BTW, my PR on generating the new `configure` file is still in review: #1651.

> 
> The only question is: squash or merge? This PR has quite a history of commits, so it might be useful to preserve it. But throughout our work on this extension, we've been making fix-up commits, and the history should be cleaned up - grouped before the merge.
> 
> I don't think preserving the commit history is an extremely important option. The cleanest and easiest approach would be to squash all the commits into one and merge. What do you think?

I prefer to keep the history commits, which will be important for the potential contributors to learn more about the context. Here just need some time to tidy them up ( my common way is squashing some simple and related continuous commits into one and rewording the commit title/commit message body ) and make the commits look more logical, as suggested by @my-ship-it. But it looks good to me for now. FYI. @NJrslv

Jaroslavian:
I reset this commit (https://github.com/apache/cloudberry/commit/f26c2877f03db0f0712f25770a169b35c17c4318) in the force push above because autoconf is not run during the configure-cloudberry.sh script. As a result, the extension was not built, so the tests failed.

The ideal case would be to make the change in configure.ac, regenerate configure, and push both files. However, since we agreed to regenerate configure later, the only thing that matters for now is configure.ac.

I left configure file as it was to ensure the tests pass.

Now I'm tidying up the commit history.

*I will approve the review once I am ready.

NJrslv (approved):
+1. I grouped the commits. We want to merge/(rebase + merge), with merge preferred. However, merging main into the current branch might cause a criss-cross merge later.

Folks, thanks all for the review. We fixed a couple of problems during it that we hadn't discovered previously.

tuhaihe:
> +1. I grouped the commits. We want to merge/(rebase + merge), with just merge preferred. However, merging main into the current branch might cause a criss-cross merge later.
> 
> Folks, thanks all for the review. We fixed a couple of problems during it that we hadn't discovered previously.

Merged! Thanks @NJrslv @Smyatkin-Maxim for your great work again！

I'd also like to share an experience regarding integrating a long-standing project (let's call it Project A) into the current repository, for future reference. Initially, Project A might have been maintained in a separate repository with a directory structure that doesn't align with the current project. During the rebase process, conflicts can arise between Project A's commit history and the current directory files. This is because rebase replays each of Project A's past commits one by one.

Therefore, when introducing similar projects into the current repository in the future, it's best to first add a target directory prefix to Project A, and then import its commit history into this new target directory. This way, the replayed commit history will only occur within the target directory, preventing conflicts with the main repository's top-level directories. This is an experience I gained while merging disktuqota, and I wanted to share it for your reference.

PR #1636, opened 2026-03-24
Backpatch 14.5-14.6 changes to REL_2_STABLE

reshke:
https://github.com/postgres/postgres/commit/aaad8adb02e - https://github.com/postgres/postgres/commit/e527cb8cecc range

### DIFF
--- a/doc/src/sgml/release-14.sgml
+++ b/doc/src/sgml/release-14.sgml
@@ -1,6 +1,1043 @@
 <!-- doc/src/sgml/release-14.sgml -->
 <!-- See header comment in release.sgml about typical markup -->
 
+ <sect1 id="release-14-6">
+  <title>Release 14.6</title>
+
+  <formalpara>
+  <title>Release date:</title>
+  <para>2022-11-10</para>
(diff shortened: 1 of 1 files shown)


reshke:
This pr completes back-patching process up to 14.6

tuhaihe (approved):
All tests passed.

tuhaihe:
Huge thanks to @reshke for all the hard work on this! 😄

(I had a quick thought: what if we try cherry-picking commits across multiple PG minor versions all at once? Each version has roughly 100 commits. I know this might be quite challenging and could put some pressure on us when resolving conflicts, but it might ultimately be much more efficient.

That being said, maintaining our current pace is also completely fine. This is just an idea for consideration. Thanks again!)

reshke:
> 
>  what if we try cherry-picking commits across multiple PG minor versions all at once? Each version has roughly 100 commits. 

Yeah, I also thought about that. 14.5-14.6 was actually done in two PRs. There are some really troublesome patches from upstream to work with... so I try to avoid doing many of such changes in single PR.

PR #1598, opened 2026-03-03
Fix sed -i compatibility on macOS in configure

zhangyue1818:
macOS BSD sed requires an explicit empty string argument after -i (sed -i '' 'script' file), unlike GNU sed which takes -i without a suffix argument. Without this fix, BSD sed misinterprets the sed script as a backup suffix and treats the filename as the script, causing "unterminated substitute pattern" error.

Fixes #28, #1335

### Type of Change
- [x] Bug fix (non-breaking change)
- [ ] New feature (non-breaking change)
- [ ] Breaking change (fix or feature with breaking changes)
- [ ] Documentation update

### Test Plan

- [ ] Unit tests added/updated
- [ ] Integration tests added/updated
- [ ] Passed `make installcheck`
- [ ] Passed `make -C src/test installcheck-cbdb-parallel`

### Impact

**Performance:**

**User-facing changes:**

**Dependencies:**

### Checklist
- [ ] Followed [contribution guide](https://cloudberry.apache.org/contribute/code)
- [ ] Added/updated documentation
- [ ] Reviewed code for security implications
- [ ] Requested review from [cloudberry committers](https://github.com/orgs/apache/teams/cloudberry-committers)

### CI Skip Instructions

---

PR #1650, opened 2026-03-31
Update configure with the latest configure.ac

tuhaihe:
Some changes to configure.ac have been made by the following PRs:
- https://github.com/apache/cloudberry/pull/1598
- https://github.com/apache/cloudberry/pull/1363

We need to regenerate the `configure` file by the command `autoconf`, not simply editing the configure directly. This commit is to generate the latest configure based the latest configure.ac.

OS: Rocky Linux 8
Autoconf: 2.69 (default version in Rocky Linux 8)

Fixes #ISSUE_Number

### Type of Change
- [ ] Documentation update

### Test Plan

### Impact

**Performance:**

**User-facing changes:**

**Dependencies:**

### Checklist

### CI Skip Instructions

---

### DIFF
--- a/configure
+++ b/configure
@@ -2912,6 +2912,7 @@ PG_PACKAGE_VERSION=14.4
 
 
 
+
 ac_aux_dir=
 for ac_dir in config "$srcdir"/config; do
   if test -f "$ac_dir/install-sh"; then
@@ -12235,7 +12236,9 @@ $as_echo "#define USE_MDBLOCALES 1" >>confdefs.h
   esac
 
 else
-  with_mdblocales=no
+  with_mdblocales=yes
+
+$as_echo "#define USE_MDBLOCALES 1" >>confdefs.h
 
 fi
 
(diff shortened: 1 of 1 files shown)


configure
```diff
@@ -12235,7 +12236,9 @@ $as_echo "#define USE_MDBLOCALES 1" >>confdefs.h
   esac
 
 else
-  with_mdblocales=no
+  with_mdblocales=yes
```
leborchuk, line 12239: Does it enable mdb_locales by default? If yes, it's bad. It's needed an additional package with custom locales (individual for every installation). For example, one installation could have a frozen libc since Rocky Linux 8 and not change the symbol orders anymore. The second could have a frozen libc in Rocky Linux 9, and so on. I prefer not to enable it. Let's do not give custom locales for users by default, it's only for those who fully understand what it is about.

tuhaihe: Based on the commit history, mdblocales has been enabled by default. The configure file in this PR was generated from the current configure.ac using autoconf, which reflects the existing default setting.

I agree with your concern about disabling mdblocales by default. Since this PR is focused on regenerating the configure file from configure.ac, how about we create a separate PR to change the default behavior of mdblocales? That way we can discuss and review the default setting change independently.

What do you think?

tuhaihe: I created a PR on disabling the mdblocales by default: https://github.com/apache/cloudberry/pull/1651

leborchuk: Yes, this has been done by mistake, let's fix it in another PR

tuhaihe:
Will use #1651. Close this PR.

PR #1653, opened 2026-04-01
ORCA: fall back to Postgres planner for KNN ORDER BY queries

yjhjstz:
ORCA is unaware of amcanorderbyop, so it plans "ORDER BY col <-> val"
  queries with a full Seq Scan + Sort instead of a native KNN ordered
  index scan.  Detect this pattern by checking whether any ORDER BY
  target is an operator with amoppurpose = AMOP_ORDER in pg_amop and at
  least one direct Var argument, then raise ExmiQuery2DXLUnsupportedFeature
  to hand the query off to the Postgres planner, which generates an
  efficient Index Only Scan with native KNN ordering.

  Queries where the ordering operator's arguments are entirely computed
  expressions (e.g. circle(col,1) <-> point(0,0)) are excluded from the
  fallback to avoid lossy-distance errors in index-only scans.

Fixes #ISSUE_Number

### Type of Change
- [ ] Documentation update

### Test Plan

### Impact

**Performance:**

**User-facing changes:**

**Dependencies:**

### Checklist

### CI Skip Instructions

---


my-ship-it (approved):

src/backend/optimizer/util/walkers.c
```diff
@@ -1011,3 +1017,102 @@ check_collation_walker(Node *node, check_collation_context *context)
...
+}
+
+/*
+ * has_plain_var_arg
+ *
+ * Return true if the OpExpr has at least one direct Var argument
+ * (not wrapped in a function or other expression).
+ */
+static bool
+has_plain_var_arg(OpExpr *op)
```
my-ship-it, line 1059: The check requires a direct Var node. A RelabelType wrapping a Var (common with type coercions) would not match:

-- If the column is int2 but the operator expects int4,
-- the Var gets wrapped in RelabelType
ORDER BY int2_col <-> 42

reshke:
will try to take a look soon. Changes looks sane

src/backend/optimizer/util/walkers.c
yjhjstz: fixed by strip_implicit_coercions

leborchuk (approved):
Checked code, cannot say I could add something meaningful here, but using index only sqan instead of seqscan is usually good

reshke (approved):

Issue #1298, opened 2025-08-08
COPY ... ON SEGMENT on REPLICATED tables: FROM returns N× in Greenplum but only 1× in Cloudberry


robertmu:
```sql

 -- Cloudberry cluster with 4 segments
 -- Greenplum cluster with 4 segments

  create table tab_replicated(a int, b int, c int) distributed replicated;
  insert into tab_replicated select i, i, i from generate_series(1, 100) i;

  copy tab_replicated to '/home/cbdb/tab_replicated_<SEGID>.txt' on segment;
  -- Greenplum: COPY 400   |  Cloudberry: COPY 400

  create table tab_replicated_new(a int, b int, c int) distributed replicated;
  copy tab_replicated_new from '/home/cbdb/tab_replicated_<SEGID>.txt' on segment;
  -- Greenplum: COPY 400   |  Cloudberry: COPY 100  (unexpected)
```

robertmu:
This bug also leads to illogical and unpredictable row counts in other scenarios(backup/restore). For example, on a 4-segment cluster performing a COPY ... FROM ... ON SEGMENT into a replicated table:

* If three segments are provided with identical data files, each containing 100 rows,
* And the fourth segment is provided with an empty file,
* The COPY command incorrectly reports that 75 rows were copied.

This result appears to be calculated as (100 rows * 3 files) / 4 segments = 75. This behavior is arbitrary and nonsensical for a data loading operation

PR #1667, opened 2026-04-08
Revert "Fix copy to replicated table display copy number is incorrect."

robertmu:
This reverts commit cfd2ce478fea76f6cfc355aa58ea1e90a069221e.

The reverted change normalizes COPY FROM row counts on replicated tables by dividing by the number of segments. This causes inconsistent and misleading completion tags (e.g. COPY TO reports N*segments while COPY FROM reports N for the same replicated data).

It also produces arbitrary results in ON SEGMENT workflows used by backup/restore. For example, on a 4-segment cluster, loading three 100-row files and one empty file can report 75 rows ((100*3)/4), which is not a meaningful load count.

Restore Greenplum-compatible behavior by reporting aggregated completed rows.

Refs:
- https://github.com/apache/cloudberry/issues/1298
- https://github.com/apache/cloudberry-backup/pull/36

### DIFF
--- a/src/backend/commands/copy.c
+++ b/src/backend/commands/copy.c
@@ -432,11 +432,6 @@ DoCopy(ParseState *pstate, const CopyStmt *stmt,
 			else
 				*processed = CopyFrom(cstate);	/* copy from file to database */
 
-			/* Handle copy to replicated table returns processed number */
-			if (Gp_role == GP_ROLE_DISPATCH &&
-				GpPolicyIsReplicated(cstate->rel->rd_cdbpolicy))
-				*processed = *processed / cstate->rel->rd_cdbpolicy->numsegments;
-
 			/*
 			 * Update view info if we actualy copy data from other place.
 			 */
--- a/src/test/isolation2/expected/ao_unique_index.out
+++ b/src/test/isolation2/expected/ao_unique_index.out
@@ -390,12 +390,12 @@ CREATE
 1: BEGIN;
 BEGIN
 1: COPY unique_index_ao_row FROM PROGRAM 'seq 1 10';
-COPY 10
+COPY 30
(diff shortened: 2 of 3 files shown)


zhangwenchao-123:
> This reverts commit [cfd2ce4](https://github.com/apache/cloudberry/commit/cfd2ce478fea76f6cfc355aa58ea1e90a069221e).
> 
> The reverted change normalizes COPY FROM row counts on replicated tables by dividing by the number of segments. This causes inconsistent and misleading completion tags (e.g. COPY TO reports N*segments while COPY FROM reports N for the same replicated data).
> 
> It also produces arbitrary results in ON SEGMENT workflows used by backup/restore. For example, on a 4-segment cluster, loading three 100-row files and one empty file can report 75 rows ((100*3)/4), which is not a meaningful load count.
> 
> Restore Greenplum-compatible behavior by reporting aggregated completed rows.
> 
> Refs:
> 
> * [COPY ... ON SEGMENT on REPLICATED tables: FROM returns N× in Greenplum but only 1× in Cloudberry #1298](https://github.com/apache/cloudberry/issues/1298)
> * [fix(test): Skip resize cluster test for Cloudberry due to replicated table issue cloudberry-backup#36](https://github.com/apache/cloudberry-backup/pull/36)

This purpose of this commit is to fix the number between copy to and copy from is different on QD when the table is replicated. Before this, the copy to always reports N rows while the copy from reports N * numsegs from the QD on replicated table. I have tested again and don't find the first problem you described. Can recheck it and give me new message? What's more, I don't understand what you mean 'on a 4-segment cluster, loading three 100-row files and one empty file can report 75 rows ((100*3)/4), which is not a meaningful load count', this commit will only influence the copy to/from on QD, can you give me more details or information to reproduce? Thx.

PR #1713, opened 2026-05-02
Feature: reject partition fullscan as gpcontrib extension

lqriu:
## What does this PR do?

Add a `planner_hook`-based extension under `gpcontrib/` that rejects queries
on partitioned tables when no effective partition pruning occurs. This prevents
unintended full partition scans in production environments.

**This replaces the previous PR #1711**, which modified core optimizer files.
Per reviewer feedback, the implementation is now fully self-contained as an
extension with **zero core file changes** (only `gpcontrib/Makefile` is touched
to register the extension in the build).

### Architecture

The extension installs a `planner_hook` that wraps `standard_planner()`.
After the planner (including ORCA) produces a `PlannedStmt`, the hook walks
the plan tree using three complementary detection strategies:

**Check 1 — PartitionPruneInfo check** (Planner + ORCA JOIN path):
For nodes with `part_prune_info != NULL`, compare `present_parts` vs `nparts`.
Exempt nodes with `initial_pruning_steps` or `exec_pruning_steps` (runtime
pruning capable, e.g., prepared statements).

**Check 2 — Append heuristic** (Planner path, no useful pruning quals):
When the standard Planner processes queries with no WHERE, WHERE 1=1, or
WHERE on non-partition-key columns, it does not generate `PartitionPruneInfo`.
We detect these by checking if `Append.apprelids` references a partitioned
table RTE (`relkind='p'`, `inh=true`).

**Check 3 — DynamicScan check** (ORCA path):
ORCA never sets `part_prune_info` on DynamicSeqScan nodes. We detect full
scans by comparing `list_length(partOids)` against the total partition count
from `pg_inherits`. Nodes with `join_prune_paramids` are skipped (JOIN
dynamic pruning).

### GUC parameters
- `reject_partition_fullscan` (bool, default ON) — enable/disable
- `partition_fullscan_threshold` (int, default 0) — max partitions allowed
  after pruning; 0 means reject only true fullscans

### Exemptions
| Scenario | Behavior |
|----------|----------|
| `reject_partition_fullscan = off` | Allowed |
| `enable_partition_pruning = off` | Allowed |
| Single-partition table (nparts <= 1) | Allowed |
| Runtime pruning steps present (prepared stmt) | Allowed |
| JOIN dynamic pruning (PartitionSelector / join_prune_paramids) | Allowed |
| No WHERE / WHERE 1=1 | **Rejected** |
| WHERE on non-partition-key column | **Rejected** |

### Error message
```
ERROR: partitioned table "schema.table" full partition scan is not
       allowed, N partitions would be scanned
HINT:  Add a WHERE clause on the partition key to enable partition pruning.
```

### Usage
```sql
-- Via shared_preload_libraries (recommended for production)
-- postgresql.conf: shared_preload_libraries = 'reject_partition_fullscan'

-- Or per-session via LOAD
LOAD 'reject_partition_fullscan';
```

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## Breaking Changes

When loaded, queries that previously performed full partition scans will
be rejected by default. Users can disable with:
`SET reject_partition_fullscan = off;`

## Test Plan
- [x] New regression test: `sql/partition_fullscan_reject.sql` (12 scenarios)
  - Basic rejection (no WHERE, WHERE 1=1, non-partition-key WHERE)
  - Pruning passes (WHERE on partition key)
  - GUC on/off toggle
  - enable_partition_pruning=off exemption
  - Single-partition table exemption
  - Threshold mode (threshold=2)
  - Prepared statement Param exemption
  - UPDATE/DELETE rejection
  - Subquery propagation
- [x] Compiled with `-Werror` — zero warnings, zero errors
- [x] Manual functional testing on utility-mode instance
- [ ] `make installcheck` (full cluster)

## Impact
- **Performance**: negligible — one lightweight `plan_tree_walker` after
  planning, added to the existing ~11 post-planning traversals
- **Core files changed**: zero (only `gpcontrib/Makefile` modified)
- **User-facing**: new ERROR for unfiltered partition queries; two new GUCs

## Compliance Checklist
- [x] I have read the [contribution guide](https://cloudberry.apache.org/contribute/code)
- [x] My code follows PostgreSQL coding conventions
- [x] New functionality includes tests
- [x] I have reviewed my changes for security implications
- [x] I have requested appropriate reviewers

## Files Changed

| File | Change |
|------|--------|
| `gpcontrib/Makefile` | Register extension in build (+4/-2) |
| `gpcontrib/reject_partition_fullscan/reject_partition_fullscan.c` | Extension source (+470) |
| `gpcontrib/reject_partition_fullscan/Makefile` | Build config (+18) |
| `gpcontrib/reject_partition_fullscan/reject_partition_fullscan.control` | Extension metadata (+4) |
| `gpcontrib/reject_partition_fullscan/reject_partition_fullscan--1.0.sql` | SQL install script (+8) |
| `gpcontrib/reject_partition_fullscan/sql/partition_fullscan_reject.sql` | Regression test (+113) |


gpcontrib/Makefile
```diff
@@ -22,15 +22,17 @@ ifeq "$(enable_debug_extensions)" "yes"
...
 else
 	recurse_targets = gp_sparse_vector \
 	       gp_distribution_policy \
                gp_internal_tools \
                gp_legacy_string_agg \
                gp_exttable_fdw \
                gp_toolkit \
-               pg_hint_plan
+               pg_hint_plan \
+               reject_partition_fullscan
```
yjhjstz: let's set only enable_debug_extensions=yes, enable reject_partition_fullscan extension

tuhaihe:
For the new created files, we need to add the standard ASF license header to them. Eg: https://github.com/apache/cloudberry/blob/f9ca4fac587c4d529d33c5d47eecaada9328f2a4/contrib/pax_storage/src/cpp/catalog/pax_fastsequence.cc#L1-L26

leborchuk:
This is just a remark. There is no need to fix anything here. Also, consider doing the same thing outside the cloudberry database code. We have a command center and information about requests being executed and their execution plans get there. It is much more convenient to analyze queries and other metrics in the command center rather than in an extension:
1. This is a separate process. It's easier to update the shared library. The whole cloudberry instance needs to be restarted. It often happens that something needs to be fixed quickly and rolling out a new version of an extension with a cloudberry restart takes a lot of time.
2. The process is written in Go. It's very easy to add new functionality - sending data to Kafka topics, coordinating actions of administrators, etc. 

Our operational experience shows that simple rules aren't enough and the rules need to change over time for each installation and analyst team in a unique way. And a companies with large installations prefers to create their own rules.

Here [gp_stats_collector ](https://github.com/apache/cloudberry/tree/main/gpcontrib/gp_stats_collector )is the source code for the extension that sends data from Cloudberry to the outside. [Here](https://github.com/open-gpdb/yagpcc) is the outside agent go code itself, which accepts and processes the data from the extension. The code is not yet in the Apache repository, but I plan to finish it this month and present to the community.

yjhjstz (approved):
LGTM

MisterRaindrop (approved):

PR #1716, opened 2026-05-02
Fix gppkg --clean error: 'SyncPackages' object has no attribute 'ret'.

woblerr:
The command `gppkg --clean` fails with the following error:  
```bash
gpadmin@master:~$ gppkg --clean
20260502:20:46:20:008305 gppkg:master:gpadmin-[INFO]:-Starting gppkg with args: --clean
20260502:20:46:20:008305 gppkg:master:gpadmin-[ERROR]:-gppkg error: SyncPackages failed:
'SyncPackages' object has no attribute 'ret'
...
```

This occurs because `operations` was being passed positionally during the OperationWorkerPool initialization, which incorrectly bound it to the `should_stop` argument instead of `items` in the base WorkerPool class.

The solution is to  pass `operations` as a keyword argument.

### DIFF
--- a/gpMgmt/bin/gppylib/commands/base.py
+++ b/gpMgmt/bin/gppylib/commands/base.py
@@ -230,7 +230,7 @@ class OperationWorkerPool(WorkerPool):
         if operations is not None:
             for operation in operations:
                 self._spoof_operation(operation)
-        super(OperationWorkerPool, self).__init__(numWorkers, operations)
+        super(OperationWorkerPool, self).__init__(numWorkers, items=operations)
 
     def check_results(self):
         raise NotImplementedError("OperationWorkerPool has no means of verifying success.")


tuhaihe:
Hi @woblerr Thanks for your fixes! 

One more question is that: are you using the _.gppkg_ package for your daily operations? If so, I think we can have a discussion on providing the gppkg for Cloudberry and other components.

woblerr:
> One more question is that: are you using the _.gppkg_ package for your daily operations? If so, I think we can have a discussion on providing the gppkg for Cloudberry and other components.

Not in everyday use. I came across a recommendation when there are problems with extensions synchronization when activating the standby coordinator.

```bash
20260502:20:16:32:005759 gpinitstandby:master:gpadmin-[INFO]:-Syncing Apache Cloudberry extensions to standby
20260502:20:16:32:005759 gpinitstandby:master:gpadmin-[WARNING]:-Syncing of Apache Cloudberry extensions has failed.
20260502:20:16:32:005759 gpinitstandby:master:gpadmin-[WARNING]:-Please run gppkg --clean after successful standby initialization.
```

[gpinitstandby#L397-L398](https://github.com/apache/cloudberry/blob/main/gpMgmt/bin/gpinitstandby#L397-L398) - current implimintation.

leborchuk (approved):
Sorry, I personally prefer not to use objects in Python, especially when one needs to call parent class constructors. 

But since this code exists, let's fix it )

tuhaihe (approved):
All tests passed.

PR #1727, opened 2026-05-09
Fix: FDW OPTIONS encoding accepts symbolic names (issue #1726)

### LINKED ISSUE #1726: [Bug] FDW OPTIONS 'encoding' silently parses non-numeric names (e.g. 'UTF8') as SQL_ASCII via atoi()
### Apache Cloudberry version

main (also reproduces on 1.x). Verified against current `main` at commit `ef0b6fb3462`.

### What happened

When defining a foreign table via the FDW path with a symbolic encoding name in `OPTIONS`:

```sql
CREATE FOREIGN TABLE ext_t (c text)
SERVER gp_exttable_server
OPTIONS (
    formatter     'pxfwritable_import',
    format        'custom',
    format_type   'b',
    location_uris 'pxf://t?PROFILE=jdbc&SERVER=s',
    encoding      'UTF8'
);
```

the DDL succeeds, but the encoding actually stored on the foreign table is **`SQL_ASCII` (0)**, not `UTF8` (`PG_UTF8 = 6`).

Root cause — both the FDW catalog reader and the option validator parse `encoding` with `atoi()`:

- `src/backend/access/external/external.c:280`
  ```c
  if (pg_strcasecmp(def->defname, "encoding") == 0)
  {
      extentry->encoding = atoi(defGetString(def));   /* 'UTF8' → 0 */
      ...
  }
  ```
- `gpcontrib/gp_exttable_fdw/option.c:136-143`
  ```c
  else if (pg_strcasecmp(def->defname, "encoding") == 0)
  {
      char *encoding = (char *) defGetString(def);
      if (!PG_VALID_ENCODING(atoi(encoding)))         /* atoi('UTF8') == 0 == SQL_ASCII, passes */
          ereport(ERROR, ...);
  }
  ```

`atoi("UTF8")` returns `0`, and `PG_VALID_ENCODING(0)` is true (`PG_SQL_ASCII`), so validation silently passes. Anyone who writes `encoding 'UTF8'`, `encoding 'utf-8'`, `encoding 'GBK'`, etc. ends up with `SQL_ASCII`.

Confirm with:

```sql
SELECT ftoptions FROM pg_foreign_table WHERE ftrelid = 'ext_t'::regclass;
-- {... ,encoding=UTF8, ...}     -- stored verbatim; atoi() at read time → 0
```

Downstream consequences:

1. On the PXF binary path (`format_type='b'`), the PXF server emits UTF-8 bytes via `String.getBytes(StandardCharsets.UTF_8)`. The segment's gpdbwritable formatter checks the foreign-table encoding and rejects anything other than UTF-8, surfacing as:
   > `gpdbwritable formatter can only import UTF8 formatted data. Define the external table with ENCODING UTF8...`

   The error message is misleading because the user's DDL **does** say `ENCODING UTF8` — but it has been silently turned into `SQL_ASCII` by `atoi()`.
2. On other formats and protocols, the misconfiguration is silent: multi-byte text gets stored as opaque bytes, and `length()` / `substring()` / `LIKE` / downstream re-encoding misbehave only when CJK or accented characters appear.

By contrast, the legacy `CREATE EXTERNAL TABLE ... ENCODING 'UTF8'` path is fine because `src/backend/commands/exttablecmds.c:209-216` resolves the name with `pg_char_to_encoding()` before persisting it as a numeric string into OPTIONS. Only the FDW OPTIONS entry point bypasses that translation.

`encoding` in FDW `OPTIONS` should accept the same forms accepted everywhere else in Postgres / Cloudberry — symbolic names (`UTF8`, `GBK`, `LATIN1`, …) **and** numeric IDs — and reject everything else with a clear error. Symbolic names should be resolved via `pg_char_to_encoding()`, just like the legacy `CREATE EXTERNAL TABLE` path.

Specifically:

| OPTIONS spelling      | Today                                | Expected                          |
| --------------------- | ------------------------------------ | --------------------------------- |
| `encoding '6'`        | UTF8 ✅                              | UTF8 ✅ (unchanged)               |
| `encoding 'UTF8'`     | SQL_ASCII ❌ (silent)                | UTF8 ✅                           |
| `encoding 'utf-8'`    | SQL_ASCII ❌ (silent)                | UTF8 ✅                           |
| `encoding 'GBK'`      | SQL_ASCII ❌ (silent)                | GBK ✅                            |
| `encoding 'bogus'`    | SQL_ASCII ❌ (silent)                | ERROR — unrecognized encoding ✅  |

#### Implementation note (FDW validator interface limitation)

The legacy `CREATE EXTERNAL TABLE` path achieves write-time normalization via a custom dispatch (`DefineExternalRelation`, `src/backend/commands/exttablecmds.c:76`) that pre-resolves the encoding and writes a numeric string into `stmt->options` **before** calling `CreateForeignTable`. The standard FDW validator contract (`src/backend/commands/foreigncmds.c:121-216` `transformGenericOptions` — the array to be persisted is built at line 190 *before* the validator is called at line 202-212, and the validator returns void) does not permit option mutation. So the FDW path needs a different injection point than just the validator. Two implementations are viable:

- **Write-time normalization**, either via a `ProcessUtility_hook` registered by `gp_exttable_fdw` or via a small site-specific branch in `CreateForeignTable` keyed on the `gp_exttable_fdw` validator. The read path (`external.c:280`) keeps its `atoi`. Existing rows with `encoding 'UTF8'` retain `SQL_ASCII` semantics on upgrade and only get fixed when the user re-issues `ALTER … OPTIONS (SET encoding 'UTF8')` — i.e. **no silent meaning change on upgrade**, which is the safer behavior. `pg_dump` output is also stabilized to numeric form.
- **Read-side name-awareness** in `GetExtFromForeignTableOptions` (`external.c:280`) and the validator (`option.c:139-142`). Simpler (~30 LOC), but existing rows that today silently render as `SQL_ASCII` would start being interpreted as actual UTF8 on upgrade — this **must** be called out in release notes.

**Update (2026-05-09): direction settled — going with the second shape.**

After implementing the first shape (write-time normalization via `ProcessUtility_hook` registered in `gp_exttable_fdw`'s `_PG_init`), the author hit a load-order issue that makes it unworkable without changing deployment expectations:

- The hook is installed by `_PG_init`, which only runs when the extension's shared library is `dlopen`-ed.
- That happens lazily, when one of the extension's `PG_FUNCTION_INFO_V1` functions is first invoked. None of them is called during connection setup — the first to fire is the validator (`gp_exttable_permission_check`), called from inside `transformGenericOptions` (`src/backend/commands/foreigncmds.c:202-212`) on the user's first `CREATE FOREIGN TABLE`.
- By the time the validator triggers `dlopen` and `_PG_init` registers our `ProcessUtility_hook`, the hook check for the *current* statement has already been passed (it happens at the top of `ProcessUtility`). So the very first `CREATE FOREIGN TABLE … encoding 'UTF8'` of every fresh backend would still bypass the hook — the exact reproduction documented at the top of this issue.
- Subsequent statements in the same session would work, but that defeats the goal: the bug bites at table-creation time, not later.

The standard cure in Cloudberry / PG (which `pg_hint_plan` follows — every one of its tests starts with `LOAD 'pg_hint_plan';` and its README requires `session_preload_libraries`) is to require the extension to be preloaded. That changes the deployment story for `gp_exttable_fdw`, which today does not need preloading, and was therefore not chosen here.

**The PR will take the second shape**: read-side name-awareness in both `GetExtFromForeignTableOptions` (`external.c:280`) and the validator (`option.c:139-142`), via a small shared helper `parse_fdw_encoding_option(const char *)` that resolves names through `pg_char_to_encoding()` and falls back to a strict numeric form (no `atoi`, since `atoi("UTF8")==0` is the bug). On-disk values stay as the user wrote them; correctness is established at read time.

Acknowledged trade-off (called out as a release-note item in the PR): existing rows whose `ftoptions` already contains a literal `encoding=UTF8` and which therefore silently render as `SQL_ASCII` today will start being interpreted as the requested encoding on upgrade. Since such rows are already misconfigured, "upgrade fixes them" is the desired behavior for nearly all affected users — but it is a behavior change. A detection query will ship in the PR body so operators who, for any reason, want to pin a specific table to its old (broken) `SQL_ASCII` semantics can re-issue `ALTER FOREIGN TABLE … OPTIONS (SET encoding '<numeric>')` before upgrade.

### How to reproduce

```sql
CREATE EXTENSION IF NOT EXISTS gp_exttable_fdw;

CREATE FOREIGN TABLE ext_utf8 (c text)
SERVER gp_exttable_server
OPTIONS (
    format        'text',
    format_type   't',
    location_uris 'file:///tmp/x.txt',
    encoding      'UTF8'
);

-- Bug: encoding is stored verbatim and atoi('UTF8')==0 → SQL_ASCII
SELECT ftoptions FROM pg_foreign_table WHERE ftrelid = 'ext_utf8'::regclass;
SELECT e.encoding, pg_encoding_to_char(e.encoding)
FROM   pg_exttable e
JOIN   pg_class c ON c.oid = e.reloid
WHERE  c.relname = 'ext_utf8';
--  encoding | pg_encoding_to_char
-- ----------+---------------------
--         0 | SQL_ASCII

-- Causal chain made explicit:
--   user wrote     :  encoding 'UTF8'
--   stored in      :  pg_foreign_table.ftoptions  →  encoding=UTF8   (literal, see ftoptions output above)
--   atoi("UTF8")   :  0                                              (stops at first non-digit)
--   effective enc  :  SQL_ASCII                                      (PG_SQL_ASCII = 0)
```

The same misbehavior applies whenever `pxf://` URIs are used together with `SERVER gp_exttable_server` (which dispatches through `gp_exttable_fdw`); in that combination the segment-side gpdbwritable formatter raises the misleading `Define the external table with ENCODING UTF8` error at SELECT time.

### Operating System

Linux 5.14.0-503.38.1.el9_5.x86_64 (Rocky / RHEL 9.5). Issue is OS-independent.

### Anything else

- **Affected scope: the `gp_exttable_fdw` FDW** (used by the built-in `gp_exttable_server`).
  - Bug sites: `src/backend/access/external/external.c:278-283` (read path) and `gpcontrib/gp_exttable_fdw/option.c:136-143` (validator).
  - The standalone `pxf_fdw` FDW is **not** affected: its validator routes `encoding` through `ProcessCopyOptions` (see `gpcontrib/pxf_fdw/pxf_option.c:367`), which natively accepts symbolic names. The PXF binary symptom described above shows up specifically when `pxf://` URIs are used with `SERVER gp_exttable_server`, which dispatches through `gp_exttable_fdw`, not through the standalone `pxf_fdw` FDW.
- The runtime "Define the external table with ENCODING UTF8" message lives in the separate `apache/cloudberry-pxf-server` repository (gpdbwritable formatter, Java); a follow-up issue/PR will land there to point users at the OPTIONS clause once the root cause here is fixed.
- `pg_char_to_encoding()` is already declared in `src/include/mb/pg_wchar.h:567` and used by the legacy `CREATE EXTERNAL TABLE` path, so no new dependency is needed.

Yes — a follow-up PR is in preparation.

talmacschen-arch:
Fixes #1726

### What does this PR do?

Make the `gp_exttable_fdw` FDW accept both **symbolic encoding names** (`UTF8`, `utf-8`, `GBK`, …) and numeric encoding IDs in the `encoding` OPTIONS entry, mirroring what the legacy `CREATE EXTERNAL TABLE … ENCODING …` path has always accepted, and what the rest of PostgreSQL accepts everywhere else (`client_encoding`, `pg_dump`, `pg_conversion`, `CREATE DATABASE`, …).

Before this PR, both the FDW catalog reader (`src/backend/access/external/external.c:280`) and the option validator (`gpcontrib/gp_exttable_fdw/option.c:139,142`) parsed the value with `atoi()`. `atoi("UTF8")` silently returns `0` (`SQL_ASCII`) and `PG_VALID_ENCODING(0)` is true, so:

| OPTIONS spelling      | Before              | After                       |
| --------------------- | ------------------- | --------------------------- |
| `encoding '6'`        | UTF8 ✅             | UTF8 ✅ (unchanged)         |
| `encoding 'UTF8'`     | SQL_ASCII (silent ❌) | UTF8 ✅                     |
| `encoding 'utf-8'`    | SQL_ASCII (silent ❌) | UTF8 ✅                     |
| `encoding 'GBK'`      | SQL_ASCII (silent ❌) | GBK ✅                      |
| `encoding 'bogus'`    | SQL_ASCII (silent ❌) | ERROR — unrecognized ✅     |

Affected scope is `gp_exttable_fdw` (used by `gp_exttable_server`). The standalone `pxf_fdw` is unaffected — its validator routes `encoding` through `ProcessCopyOptions`, which is already name-aware.

#### Implementation

A small shared helper `parse_fdw_encoding_option(const char *)` is added in `src/backend/access/external/external.c` (declared in `src/include/access/external.h`):

1. Try `pg_char_to_encoding(value)` — the same translation the legacy `CREATE EXTERNAL TABLE` path uses.
2. Otherwise try a strict numeric form (`strtol` + end-of-string + `PG_VALID_ENCODING` checks). `atoi` is intentionally avoided: `atoi("UTF8")` silently mistranslates and that is the bug being fixed.
3. Otherwise `ereport(ERROR)`.

Both the FDW validator (`gp_exttable_permission_check`) and the read path (`GetExtFromForeignTableOptions`) call this helper. No on-disk format change: values are stored verbatim in `pg_foreign_table.ftoptions` exactly as the user wrote them, and resolved at read time. This keeps `pg_dump` output round-trip-safe and avoids silently rewriting catalog rows during upgrade.

The bypass on the runtime PXF error message ("`Define the external table with ENCODING UTF8...`") lives in the separate `apache/cloudberry-pxf-server` repository and will be addressed in a follow-up PR there.

### Type of Change
- [x] **Behavior change on upgrade** — see "User-facing changes" below
- [ ] Documentation update

### Test Plan

Added cases to `gpcontrib/gp_exttable_fdw/input/gp_exttable_fdw.source`:

- `encoding '6'` → resolves to UTF8 (control)
- `encoding 'UTF8'` / `encoding 'utf-8'` → resolves to UTF8 (the fix)
- `encoding 'GBK'` → resolves to GBK
- `encoding 'bogus'` → ereports
- `ALTER FOREIGN TABLE … OPTIONS (SET encoding 'UTF8')` on a table that
  was originally created with `encoding '0'` → reads as UTF8 after the ALTER

The pre-existing `encoding '-1'` ereport test was kept; only the error message text was updated to the new `"\"-1\" is not a valid encoding name or code"` form.

- [x] Unit tests added/updated (`gpcontrib/gp_exttable_fdw/`)
- [ ] Passed `make installcheck` *(to be run by reviewer / CI)*
- [ ] Passed `make -C src/test installcheck-cbdb-parallel` *(to be run)*

### Impact

**User-facing changes (release-note worthy ⚠️):**

Foreign tables that were created **before** this fix with a non-numeric `encoding` option (e.g. `encoding 'UTF8'`) had the literal name silently stored in `pg_foreign_table.ftoptions`. With the old `atoi()` reader they took effect as `SQL_ASCII`. After this fix, the same on-disk row is interpreted as the named encoding (e.g. `UTF8`) — i.e. the encoding actually requested by the user.

For most affected users this is the desired correction (data was already being silently mishandled). Operators who want to identify which existing tables are about to flip semantics can run:

```sql
SELECT n.nspname, c.relname,
       (SELECT split_part(opt, '=', 2)
        FROM   unnest(ftoptions) AS opt
        WHERE  opt LIKE 'encoding=%') AS stored_encoding
FROM   pg_foreign_table ft
JOIN   pg_class c ON c.oid = ft.ftrelid
JOIN   pg_namespace n ON n.oid = c.relnamespace
WHERE  ft.ftserver IN (SELECT oid FROM pg_foreign_server
                       WHERE  srvname IN ('gp_exttable_server'))
       AND EXISTS (
         SELECT 1 FROM unnest(ftoptions) AS opt
         WHERE  opt LIKE 'encoding=%'
                AND substring(opt FROM 'encoding=(.*)') !~ '^[0-9]+$'
       );
```

If a table appears that should *not* flip, the operator can pin it by re-issuing
`ALTER FOREIGN TABLE … OPTIONS (SET encoding '<numeric>')` before upgrade.

**Performance:** none. Two extra string compares + one syscache-free name lookup on each foreign-table read; same for validation.

**Dependencies:** none. `pg_char_to_encoding()` is already declared in `src/include/mb/pg_wchar.h:567` and used by the legacy path.

### Checklist
- [ ] Added/updated documentation (release-note line; doc PR not required)
- [x] Reviewed code for security implications (no new external input surfaces; helper ereports on invalid input)

### Additional Context

The design discussion in the issue (`#1726`) considered a write-time normalization variant via `ProcessUtility_hook`, where the option value would be rewritten to its numeric form *before* persisting into `ftoptions`. That approach was abandoned because the hook is only installed by the extension's `_PG_init`, which does not run until the first call into the extension's shared library — and on the very first `CREATE FOREIGN TABLE` of a fresh backend, the hook check has already been passed by the time the validator triggers `dlopen`. Adopting it would require listing `gp_exttable_fdw` in `session_preload_libraries`, which changes deployment expectations. The chosen read-side resolution works without any preload requirement.

### DIFF
--- a/gpcontrib/gp_exttable_fdw/input/gp_exttable_fdw.source
+++ b/gpcontrib/gp_exttable_fdw/input/gp_exttable_fdw.source
@@ -53,10 +53,16 @@ OPTIONS (format_type 'c', delimiter ',',
          location_uris 'file://@hostname@@abs_srcdir@/data/tableless.csv',
          reject_limit_type 'p', reject_limit '120');
 
--- Error, invalid encoding
+-- Error, invalid encoding (negative numeric ID)
 CREATE FOREIGN TABLE tableless_ext_fdw(a int, b int)
 SERVER gp_exttable_server
 OPTIONS (format_type 'c', delimiter ',', encoding '-1',
+         location_uris 'file://@hostname@@abs_srcdir@/data/tableless.csv');
+
+-- Error, invalid encoding (unknown name)
+CREATE FOREIGN TABLE tableless_ext_fdw(a int, b int)
+SERVER gp_exttable_server
+OPTIONS (format_type 'c', delimiter ',', encoding 'bogus',
          location_uris 'file://@hostname@@abs_srcdir@/data/tableless.csv');
 
 -- OK, no execute_on | log_errors | encoding | is_writable option
@@ -79,3 +85,59 @@ SELECT urilocation FROM pg_exttable WHERE reloid = 'public.ext_special_uri'::reg
 SELECT ftoptions FROM pg_foreign_table WHERE ftrelid='public.ext_special_uri'::regclass;
 \a
 SELECT * FROM ext_special_uri ORDER BY a;
+
+-- ===================================================================
+-- Tests for issue #1726: FDW OPTIONS encoding accepts both numeric IDs
+-- and symbolic names (UTF8, utf-8, GBK, ...). Names previously parsed
+-- via atoi() and silently degraded to SQL_ASCII.
+-- ===================================================================
+
+-- Numeric form (baseline; worked before the fix as well).
+CREATE FOREIGN TABLE ext_enc_num (a int) SERVER gp_exttable_server
+OPTIONS (format_type 'c', delimiter ',',
+         location_uris 'file:///tmp/ext_enc_ignored.csv',
+         encoding '6');
+SELECT pg_encoding_to_char(encoding) FROM pg_exttable
+WHERE  reloid = 'ext_enc_num'::regclass;
+
+-- Symbolic name 'UTF8' — used to be silently SQL_ASCII (the bug).
+CREATE FOREIGN TABLE ext_enc_utf8 (a int) SERVER gp_exttable_server
+OPTIONS (format_type 'c', delimiter ',',
+         location_uris 'file:///tmp/ext_enc_ignored.csv',
+         encoding 'UTF8');
+SELECT pg_encoding_to_char(encoding) FROM pg_exttable
+WHERE  reloid = 'ext_enc_utf8'::regclass;
+
+-- Case + dash variant resolved by pg_char_to_encoding().
+CREATE FOREIGN TABLE ext_enc_utf8_dash (a int) SERVER gp_exttable_server
+OPTIONS (format_type 'c', delimiter ',',
+         location_uris 'file:///tmp/ext_enc_ignored.csv',
+         encoding 'utf-8');
+SELECT pg_encoding_to_char(encoding) FROM pg_exttable
+WHERE  reloid = 'ext_enc_utf8_dash'::regclass;
+
+-- Non-UTF8 symbolic name.
+CREATE FOREIGN TABLE ext_enc_gbk (a int) SERVER gp_exttable_server
+OPTIONS (format_type 'c', delimiter ',',
+         location_uris 'file:///tmp/ext_enc_ignored.csv',
+         encoding 'GBK');
+SELECT pg_encoding_to_char(encoding) FROM pg_exttable
+WHERE  reloid = 'ext_enc_gbk'::regclass;
+
+-- ALTER FOREIGN TABLE ... OPTIONS (SET encoding 'UTF8') — same code
+-- path, this proves the read-side resolution works after an ALTER too.
+CREATE FOREIGN TABLE ext_enc_alter (a int) SERVER gp_exttable_server
+OPTIONS (format_type 'c', delimiter ',',
+         location_uris 'file:///tmp/ext_enc_ignored.csv',
+         encoding '0');
+SELECT pg_encoding_to_char(encoding) FROM pg_exttable
+WHERE  reloid = 'ext_enc_alter'::regclass;
+ALTER FOREIGN TABLE ext_enc_alter OPTIONS (SET encoding 'UTF8');
+SELECT pg_encoding_to_char(encoding) FROM pg_exttable
+WHERE  reloid = 'ext_enc_alter'::regclass;
+
+DROP FOREIGN TABLE ext_enc_num;
+DROP FOREIGN TABLE ext_enc_utf8;
+DROP FOREIGN TABLE ext_enc_utf8_dash;
+DROP FOREIGN TABLE ext_enc_gbk;
+DROP FOREIGN TABLE ext_enc_alter;
--- a/gpcontrib/gp_exttable_fdw/input/gp_exttable_fdw.source
+++ b/gpcontrib/gp_exttable_fdw/input/gp_exttable_fdw.source
@@ -65,6 +65,14 @@ SERVER gp_exttable_server
 OPTIONS (format_type 'c', delimiter ',', encoding 'bogus',
          location_uris 'file://@hostname@@abs_srcdir@/data/tableless.csv');
 
+-- Error, mixed numeric+letters must not be silently truncated to a
+-- valid prefix (atoi('6abc') would return 6 = UTF8; strict parsing
+-- in parse_fdw_encoding_option() rejects it).
+CREATE FOREIGN TABLE tableless_ext_fdw(a int, b int)
+SERVER gp_exttable_server
+OPTIONS (format_type 'c', delimiter ',', encoding '6abc',
+         location_uris 'file://@hostname@@abs_srcdir@/data/tableless.csv');
+
 -- OK, no execute_on | log_errors | encoding | is_writable option
 CREATE FOREIGN TABLE tableless_ext_fdw(a int, b int)
 SERVER gp_exttable_server
--- a/gpcontrib/gp_exttable_fdw/option.c
+++ b/gpcontrib/gp_exttable_fdw/option.c
@@ -135,11 +135,13 @@ gp_exttable_permission_check(PG_FUNCTION_ARGS)
 		}
 		else if(pg_strcasecmp(def->defname, "encoding") == 0)
 		{
-			char	*encoding = (char *) defGetString(def);
-			if (!PG_VALID_ENCODING(atoi(encoding)))
-				ereport(ERROR,
-				        (errcode(ERRCODE_FDW_INVALID_ATTRIBUTE_VALUE),
-				         errmsg("%s is not a valid encoding code", encoding)));
+			/*
+			 * Accept either a symbolic encoding name (e.g. 'UTF8', 'GBK')
+			 * or a numeric encoding ID. Reject anything else explicitly,
+			 * rather than letting atoi() silently mistranslate non-numeric
+			 * names to SQL_ASCII.
+			 */
+			(void) parse_fdw_encoding_option((char *) defGetString(def));
 		}
 	}
 
--- a/gpcontrib/gp_exttable_fdw/output/gp_exttable_fdw.source
+++ b/gpcontrib/gp_exttable_fdw/output/gp_exttable_fdw.source
@@ -95,7 +95,7 @@ ftoptions
 (1 row)
 \a
 SELECT * FROM ext_special_uri ORDER BY a;
- a | b
+ a | b 
 ---+---
  1 | 1
  2 | 2
@@ -114,7 +114,7 @@ OPTIONS (format_type 'c', delimiter ',',
          encoding '6');
 SELECT pg_encoding_to_char(encoding) FROM pg_exttable
 WHERE  reloid = 'ext_enc_num'::regclass;
- pg_encoding_to_char
+ pg_encoding_to_char 
 ---------------------
  UTF8
 (1 row)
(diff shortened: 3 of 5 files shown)


HuSen8891 (approved):

talmacschen-arch:
Hi committers — could someone please approve the queued workflow runs on this PR?

The previous CI run had a single failure in `ic-cbdb-parallel` on the Debian pipeline: `alter_distribution_policy` flipped one boolean (`t` → `f`) on the line

```
select check_redistributed(
  'alter table t_reorganize set with (reorganize=true) distributed randomly',
  't_reorganize', 't_reorganize');
```

This test compares per-segment row distribution before/after a `reorganize=true ... distributed randomly`, and with only 1000 rows the outcome is sensitive to the random hash mapping — looks like a pre-existing flake (RPM pipeline's same job passed, all other 42 jobs are green, and this PR only touches FDW `OPTIONS encoding` parsing, unrelated to distribution).

I pushed an empty commit (2f0b3304) to retrigger CI, but the workflows are stuck in `action_required` since I'm an external contributor. Thanks!

leborchuk (approved):
LGTM, everything is clear, should be fixed as was described

my-ship-it (approved):

talmacschen-arch:
Hi @my-ship-it, thanks for the review and approval!

All CI checks are green and the PR is now in a clean mergeable state. Would you mind merging it when you have a moment? Happy to address any follow-ups if needed.

PR #1750, opened 2026-05-16
Fix infinite loop in replaceStringInfoString

Smyatkin-Maxim:
The function had a few problems:
- replace ="a" and replacement="ab" would give an infinite loop leading to OOM. We have a similar case in pg_dump: replace="range", replacement="multirange"
- Copying the whole string each time there is a match
- Empty replace pattern also leads to infinite loop

We can still do better: for example count number of matches and do a single allocation, or even run the replacement inplace. But it's probably not worth it.

### Type of Change
- [ ] Documentation update

### Test Plan

- [x] gpupgrade of regression database

### Impact

**Performance:**

**User-facing changes:**

**Dependencies:**

### Checklist

### CI Skip Instructions

---

### DIFF
--- a/src/common/stringinfo.c
+++ b/src/common/stringinfo.c
@@ -361,16 +361,32 @@ enlargeStringInfo(StringInfo str, int needed)
 void
 replaceStringInfoString(StringInfo str, char *replace, char *replacement)
 {
-	char	*ptr;
+	char	*match_ptr = NULL;
+	char	*start_ptr = str->data;
+	char	*dup = NULL;
+	size_t	replace_len = strlen(replace);
 
-	while ((ptr = strstr(str->data, replace)) != NULL)
+	// prevent empty loop, cuz strstr will always return start_ptr
+	if (replace_len == 0)
+		return;
+
+	while ((match_ptr = strstr(start_ptr, replace)) != NULL)
 	{
-		char	   *dup = pstrdup(str->data);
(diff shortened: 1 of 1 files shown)


leborchuk (approved):
There is no replace String Info in https://github.com/postgres/postgres/blob/master/src/common/stringinfo.c so we need to implement our own solution

yjhjstz (approved):
LGTM

src/common/stringinfo.c
```diff
@@ -364,16 +364,32 @@ enlargeStringInfo(StringInfo str, int needed)
...
-	char	*ptr;
+	char	*match_ptr = NULL;
+	char	*start_ptr = str->data;
+	char	*dup = NULL;
+	size_t	replace_len = strlen(replace);
 
-	while ((ptr = strstr(str->data, replace)) != NULL)
-	{
-		char	   *dup = pstrdup(str->data);
+	// prevent empty loop, cuz strstr will always return start_ptr
```
reshke: cuz -> because

src/common/stringinfo.c
```diff
@@ -361,16 +361,32 @@ enlargeStringInfo(StringInfo str, int needed)
...
```
tuhaihe:
```suggestion
	// prevent empty loop, because strstr will always return start_ptr
```

PR #1754, opened 2026-05-18
ORCA: align CBitSet vec_size for grouping-set bitsets

### LINKED ISSUE #1753: [Bug] Wrong result on ROLLUP/CUBE when a SortGroupRef exceeds num_cols
### Apache Cloudberry version

  
 main branch

### What happened

 When ORCA translates a `GROUP BY ROLLUP(...)` / `CUBE(...)` / `GROUPING SETS ()` query, columns can be **silently NULL-ed out in the wrong grouping levels**, producing incorrect query
  results — no error, no crash, just wrong output.

 
  

## Reproduction

  ```sql
  -- Apache Cloudberry / GPDB, ORCA enabled (default)
  SET optimizer = on;

  SELECT generate_series(1, a) g, a+b ab
    FROM (VALUES (1,1),(2,2)) t(a,b)
    GROUP BY ROLLUP(a, ab)
    ORDER BY 1, 2;

  Actual (buggy) result with ORCA

   g | ab
  ---+----
  (0 rows)

Expected result (matches PostgreSQL planner, SET optimizer = off)

   g | ab
  ---+----
   1 |  2
   1 |  4
   1 |
   1 |
   2 |  4
   2 |
  (6 rows)

### How to reproduce

RT

### Operating System

centos9

### Anything else

_No response_

### Code of Conduct

yjhjstz:
CreateGroupingSetsForRollup / Cube and the GROUPING_SET_EMPTY case in GetColumnAttnosForGroupBy were constructing their accumulator/seed CBitSets via the default ctor (vec_size = 256), then Union'ing in per-grouping-set bitsets built with vec_size = num_cols.  CBitSet::Union just splices in any missing CBitSetLinks wholesale, so the accumulator ended up with a link at offset 0 (vec_size 256) plus a stray link at offset num_cols (vec_size num_cols) covering the high tleSortGroupRef. CBitSet::Get then computed the offset using the destination's m_vector_size = 256 and never consulted the stray link, while CBitSetIter happily walked both -- so Get(k) disagreed with the iterator for any k >= num_cols.

In CreateDXLProjectNullsForGroupingSets this caused tleSortGroupRefs >= num_cols to be misclassified as non-grouping columns and NULL'd out, even in grouping sets that included them.  Visible as:

  select generate_series(1, a) g, a+b ab
    from (values (1,1),(2,2)) t(a,b)
    group by rollup(a, ab) order by 1,2;

returning 0 rows instead of 6 -- the rollup(a, ab) branch projected the a column as NULL, so generate_series(1, NULL) produced no rows.

Fix by passing num_cols when constructing the accumulator and seed bitsets so all participants in the Union share m_vector_size.

Add the repro to groupingsets.sql.

Fixes #1753

### Type of Change
- [ ] Documentation update

### Test Plan

### Impact

**Performance:**

**User-facing changes:**

**Dependencies:**

### Checklist

### CI Skip Instructions

---

### DIFF
--- a/src/backend/gpopt/translate/CTranslatorUtils.cpp
+++ b/src/backend/gpopt/translate/CTranslatorUtils.cpp
@@ -916,7 +916,7 @@ CTranslatorUtils::GetColumnAttnosForGroupBy(
 			case GROUPING_SET_EMPTY:
 			{
 				col_attnos_arr_current = GPOS_NEW(mp) CBitSetArray(mp);
-				CBitSet *bset = GPOS_NEW(mp) CBitSet(mp);
+				CBitSet *bset = GPOS_NEW(mp) CBitSet(mp, num_cols);
 				col_attnos_arr_current->Append(bset);
 				break;
 			}
@@ -1111,11 +1111,12 @@ CTranslatorUtils::CreateGroupingSetsForRollup(CMemoryPool *mp,
 	GPOS_ASSERT(grouping_set->kind == GROUPING_SET_ROLLUP);
 	CBitSetArray *col_attnos_arr = GPOS_NEW(mp) CBitSetArray(mp);
 	ListCell *lc = nullptr;
-	CBitSet *current_result = GPOS_NEW(mp) CBitSet(mp);
+
+	CBitSet *current_result = GPOS_NEW(mp) CBitSet(mp, num_cols);
 	// Maintaining the order of grouping sets is essential because the
 	// UnionAll operator matches each child's distribution with the
 	// distribution of the first child
-	col_attnos_arr->Append(GPOS_NEW(mp) CBitSet(mp));
+	col_attnos_arr->Append(GPOS_NEW(mp) CBitSet(mp, num_cols));
 	ForEach(lc, grouping_set->content)
 	{
 		GroupingSet *gs_current = (GroupingSet *) lfirst(lc);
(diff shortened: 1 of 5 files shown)


src/test/regress/expected/groupingsets_optimizer.out
```diff
@@ -2645,4 +2645,18 @@ group by rollup (a,b) order by a;
    |   |    6
 (8 rows)
 
+-- ORCA: rollup over a derived-expression group alias with a target-list SRF.
+select generate_series(1, a) g, a+b ab
+  from (values (1,1),(2,2)) t(a,b)
+  group by rollup(a, ab) order by 1,2;
```
leborchuk, line 2652: I checked also cube operation - it also returns 0 rows:

```
select generate_series(1, a) g, a+b ab
  from (values (1,1),(2,2)) t(a,b)
  group by cube(a, ab) order by 1,2;
 g | ab
---+----
(0 rows)
```

Could we add also query with cube to regress tests for better coverage?

src/backend/gpopt/translate/CTranslatorUtils.cpp
```diff
@@ -918,7 +918,7 @@ CTranslatorUtils::GetColumnAttnosForGroupBy(
 			case GROUPING_SET_EMPTY:
 			{
 				col_attnos_arr_current = GPOS_NEW(mp) CBitSetArray(mp);
-				CBitSet *bset = GPOS_NEW(mp) CBitSet(mp);
+				CBitSet *bset = GPOS_NEW(mp) CBitSet(mp, num_cols);
```
leborchuk, line 919: Here we fixed issue by setting the same BitSet size for all used bitsets. But shouldn't we check if the sizes are equal in CBitSet::Union? Is it possible to perform Union with bitsets of different sizes?

yjhjstz: good suggestion.

src/test/regress/expected/groupingsets_optimizer.out
yjhjstz: fixed.

leborchuk (approved):
LGTM

zhangwenchao-123 (approved):
LGTM

PR #1756, opened 2026-05-19
ORCA: handle INDF predicate in PdxlnMergeJoin

yjhjstz:
PdxlnMergeJoin extracted merge-key operands via blind indexing:

    CExpression *pexprPredOuter = (*pexprPred)[0];
    CExpression *pexprPredInner = (*pexprPred)[1];

That works for plain equality `a = b` (binary ScalarCmp) but a NULL- safe equality clause is `NOT (a IS DISTINCT FROM b)` -- a *unary* NOT wrapping the binary IsDistinctFrom.  Indexing [1] on the NOT walks off the end of its 1-element child array, then DeriveUsedColumns on the garbage pointer segfaults.

Use CPhysicalJoin::AlignJoinKeyOuterInner -- the same helper PdxlnHashJoin already uses (line 5139) -- which handles both equality and INDF shapes and strips binary-coercible casts.  Reconstruct the DXL predicate from the aligned operands via PexprScalarCmp / PexprINDF so the downstream PG translator sees an operand order matching the children we emit.

Fixes #ISSUE_Number

### Type of Change
- [ ] Documentation update

### Test Plan

### Impact

**Performance:**

**User-facing changes:**

**Dependencies:**

### Checklist

### CI Skip Instructions

---

### DIFF
--- a/src/backend/gporca/libgpopt/src/translate/CTranslatorExprToDXL.cpp
+++ b/src/backend/gporca/libgpopt/src/translate/CTranslatorExprToDXL.cpp
@@ -4964,41 +4964,46 @@ CTranslatorExprToDXL::PdxlnMergeJoin(CExpression *pexprMJ,
 		// At this point, they all better be merge joinable
 		GPOS_ASSERT(CPhysicalJoin::FMergeJoinCompatible(
 			pexprPred, pexprOuterChild, pexprInnerChild));
-		CExpression *pexprPredOuter = (*pexprPred)[0];
-		CExpression *pexprPredInner = (*pexprPred)[1];
 
-		// align extracted columns with outer and inner children of the join
-		CColRefSet *pcrsOuterChild = pexprOuterChild->DeriveOutputColumns();
-		CColRefSet *pcrsPredInner = pexprPredInner->DeriveUsedColumns();
+		// Extract the two key columns out of pexprPred.  Plain equality
+		// is `a = b` (binary ScalarCmp) but NULL-safe equality (INDF) is
+		// `NOT (a IS DISTINCT FROM b)`, a *unary* NOT around the binary
+		// IsDistinctFrom.  Indexing [0]/[1] directly works for equality
+		// but accesses out of bounds for INDF.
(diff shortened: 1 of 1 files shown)


src/backend/gporca/libgpopt/src/translate/CTranslatorExprToDXL.cpp
```diff
@@ -4964,41 +4964,46 @@ CTranslatorExprToDXL::PdxlnMergeJoin(CExpression *pexprMJ,
...
-			pexprPredInner->AddRef();
-			pexprPred =
-				CUtils::PexprScalarEqCmp(m_mp, pexprPredOuter, pexprPredInner);
+			pexprPredNew = CUtils::PexprScalarCmp(m_mp, pexprPredOuter,
+												  pexprPredInner, mdid_scop);
 		}
 		else
 		{
-			pexprPred->AddRef();
+			GPOS_ASSERT(CPredicateUtils::FINDF(pexprPred));
```
leborchuk, line 5000: Is it possible to have INDF for merge joins? Because in `CPhysicalJoin::FMergeJoinCompatible ` we check:

```
// Only merge join between ScalarIdents of the same types is currently supported
	if (CPredicateUtils::FEqIdentsOfSameType(pexprPred))
	{
		pexprPredOuter = (*pexprPred)[0];
		pexprPredInner = (*pexprPred)[1];
		mdid_scop = CScalarCmp::PopConvert(pexprPred->Pop())->MdIdOp();
		GPOS_ASSERT(CUtils::FScalarIdent(pexprPredOuter));
		GPOS_ASSERT(CUtils::FScalarIdent(pexprPredInner));
	}
	else
	{
		return false;
	}
```

So the old code should work since we always check only for valid merge join operands.

Maybe we want to change it in future PR's?

yjhjstz: yes.

leborchuk (approved):
LGTM. We check expressions the same way we did it for Hash Join

PR #1758, opened 2026-05-20
Packaging: Enable RPM package relocation via --prefix

tuhaihe:
Refactor the RPM spec to properly support rpm --prefix for custom install locations.

- Split hardcoded install path into base_dir and name components
- Set Prefix to the base directory (/usr/local) so RPM's relocate engine can substitute it correctly
- Change symlink from absolute to relative so it remains valid after relocation
- Update %post/%postun scriptlets to use RPM macro expansion for the actual relocated path instead of hardcoded paths

Assisted-by: Claude Code

Fixes #ISSUE_Number

### Type of Change
- [ ] Documentation update

### Test Plan

### Impact

**Performance:**

**User-facing changes:**

**Dependencies:**

### Checklist

### CI Skip Instructions

---

### DIFF
--- a/.github/workflows/build-cloudberry-rocky10.yml
+++ b/.github/workflows/build-cloudberry-rocky10.yml
@@ -661,6 +661,9 @@ jobs:
             "${SRC_DIR}"/devops/build/packaging/rpm/build-rpm.sh --version "${CBDB_VERSION}" --release "${BUILD_NUMBER}" "${DEBUG_RPMBUILD_OPT}"
 
             # Get OS version and move RPM
+            # build-rpm.sh normalizes the published file name to the historical
+            # format (without the "-<major>" segment); the package Name metadata
+            # still embeds the major version.
             os_version=$(grep -oP '(?<=^VERSION_ID=")[0-9]+' /etc/os-release)
             RPM_FILE="${HOME}"/rpmbuild/RPMS/x86_64/apache-cloudberry-db-incubating-"${CBDB_VERSION}"-"${BUILD_NUMBER}""${DEBUG_IDENTIFIER}".el"${os_version}".x86_64.rpm
             cp "${RPM_FILE}" "${SRC_DIR}"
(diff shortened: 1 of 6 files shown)


tuhaihe:
No rush to review. I will test it by packaging a new RPM.

tuhaihe:
Updates:

```
Rework the RPM packaging so Cloudberry can be installed to a custom
location and so multiple major versions can coexist, aligning the
behavior with Greenplum's packaging model.

Relocation (rpm --prefix):
- Split the hardcoded install path into base_dir + name components and
  set Prefix to the base directory (/usr/local) so RPM's relocate
  engine can substitute it correctly.
- Move the convenience symlink out of %files and create it in %post
  using RPM_INSTALL_PREFIX, so it follows the actual --prefix (with a
  fallback to the default base_dir). %postun removes the symlink only
  when it still points to the version being erased.
- Disable build-id links so they are not emitted outside the
  relocatable prefix.

Major-version coexistence:
- Embed the major version in the package Name
  (apache-cloudberry-db-incubating-<major>), derived from %{version},
  so different major versions register as distinct packages and can be
  installed side by side (dnf) or under separate prefixes (rpm -i).
- Obsolete the previous unversioned package name on upgrade; this does
  not match the versioned names, so majors still coexist.
- In %post, move the generic symlink only when the existing target is
  the same major version, leaving other majors untouched. The version
  is parsed from the target directory name rather than by sourcing
  environment files or executing installed binaries.
- Keep the published file name in the historical format (without the
  "-<major>" segment) by renaming artifacts in build-rpm.sh; the file
  name does not affect the Name/Version stored in the RPM header.

Packaging correctness and hardening:
- Stop changing installed files to gpadmin ownership in %post; the
  package tree stays root-owned.
- Do not expose bundled private shared libraries (libpq.so.5, etc.) as
  Provides, and do not require them from the system; they ship in the
  package and are resolved via RPATH.
- Validate that version/release macros are supplied, mark
  cloudberry-env.sh as %config(noreplace), copy the tree with cp -a to
  preserve hidden files and attributes, add coreutils scriptlet
  dependencies, add python3 to the el9 runtime requires, and update the
  license tag to the SPDX identifier Apache-2.0.

CI workflows:
- Query the installed package by glob (rpm -qa 'apache-...*') instead of
  the fixed name, since the registered Name now carries the major.
- Keep locating the published artifact by its historical file name,
  which build-rpm.sh restores after the build.

```

tuhaihe:
Now this PR is ready for review.

cc @leborchuk @my-ship-it PTAL, thanks!

Would like to have these changes for coming 2.2 release.

leborchuk (approved):
LGTM

my-ship-it (approved):

PR #392, opened 2024-03-11
Export numeric structure and interface to public

jiaqizho:
fix #ISSUE_Number

---

### Change logs

CBDB not allow building pg numeric from other formats. Numeric objects can only be constructed from string, which means that the caller needs to convert its own data into a specific format string, then convert the string into numeric. This is a very inefficient method.

In order to better convert data of other structures into pg numeric, The current changes will expose the main structures and methods defined in `numeric.c` to `numeric.h`

### Why are the changes needed?

nope

### Does this PR introduce any user-facing change?

nope

### How was this patch tested?

nope

### Contributor's Checklist

Here are some reminders and checklists before/when submitting your pull request, please check them:

- [ ] Make sure your Pull Request has a clear title and commit message. You can take [git-commit](https://github.com/cloudberrydb/cloudberrydb/blob/main/.gitmessage) template as a reference.
- [ ] Sign the Contributor License Agreement as prompted for your first-time contribution(*One-time setup*).
- [ ] Learn the [coding contribution guide](https://cloudberrydb.org/contribute/code), including our code conventions, workflow and more.
- [ ] List your communication in the [GitHub Issues](https://github.com/cloudberrydb/cloudberrydb/issues) or [Discussions](https://github.com/orgs/cloudberrydb/discussions) (if has or needed).
- [ ] Document changes.
- [ ] Add tests for the change
- [ ] Pass `make installcheck`
- [ ] Pass `make -C src/test installcheck-cbdb-parallel`
- [ ] Feel free to request `cloudberrydb/dev` team for review and approval when your PR is ready🥳

PR #1774, opened 2026-05-29
macOS build support (Apple clang, PG 16.9)

avamingli:
## Background

Apache Cloudberry today builds cleanly on Linux only. macOS developers
have to spin up a Linux VM or container to compile and run the
project locally, which slows the inner loop on day-to-day work (read
code, change a line, see it run).

This PR brings the macOS host (Intel Mac tested; Apple Silicon
should follow the same pattern) up to first-class buildable status:

- `./configure` succeeds with the full feature set
  (PAX, ic-proxy, ORCA, orafce, mapreduce, gpcloud, paxformat,
  tap-tests, plperl, plpython3, GSSAPI, PAM, LDAP, LZ4, ZSTD, …).
- `make && make install` complete cleanly.
- `gpAux/gpdemo` produces a 3-pair demo cluster with coordinator +
  standby + 3 primaries + 3 mirrors, all in sync.

Tested on macOS 13.x x86_64 with Xcode CommandLineTools (Apple clang
14.0.3) and Homebrew at `/usr/local`.

## Strict guarantee: zero impact on existing Linux/gcc builds

Every change in this PR is gated so the existing Linux build path is
**byte-for-byte unchanged**. Specifically, each modified file falls
into one of these categories:

| Mechanism | Files |
|---|---|
| `PORTNAME=darwin` Makefile gate | `src/backend/Makefile`, `src/backend/gporca/gporca.mk`, `contrib/pax_storage/Makefile` |
| `if(APPLE) / NOT APPLE / CMAKE_SYSTEM_NAME` cmake gate | `contrib/pax_storage/CMakeLists.txt`, `contrib/pax_storage/src/cpp/CMakeLists.txt`, `contrib/pax_storage/src/cpp/cmake/{pax,pax_format}.cmake` |
| `case $host_os in linux*) ;;` autoconf gate | `configure.ac`, `configure` (for liburing + Python probe) |
| `src/template/darwin` — loaded only on darwin | (the DLSUFFIX pin) |
| `#ifdef __APPLE__` / `#ifdef __linux__` C preprocessor gate | `contrib/pax_storage/src/cpp/comm/fast_io.{h,cc}`, `contrib/pax_storage/src/cpp/storage/local_file_system.cc`, `contrib/pax_storage/src/cpp/storage/file_system.h` |
| `#if defined(__clang__)` C++ gate (preserves gcc codegen) | `contrib/pax_storage/src/cpp/storage/columns/pax_delta_encoding.cc` (VLA vs std::vector) |
| Identical compilation on Linux (zero functional change) | `src/include/utils/numeric.h` (drop bogus `const` on non-pointer return type), `contrib/pax_storage/src/cpp/storage/proto/proto_wrappers.h` (defensive `#undef`/redef of PG macros around protobuf include), `contrib/pax_storage/src/cpp/storage/proto/protobuf_stream.{h,cc}` (`google::protobuf::int64` → `int64_t`, equivalent typedef on older protobuf), `contrib/pax_storage/src/cpp/storage/columns/pax_encoding_utils.{h,cc}` (`int64_t *` → `int64 *`, identical LP64 type on Linux), `src/test/regress/GNUmakefile` (`.so` → `$(DLSUFFIX)`, which expands to `.so` on Linux) |

For every gated change the `else` / non-gated branch retains the
exact pre-PR code (verbatim). For every non-gated change the
generated machine code on Linux gcc is identical.

`make installcheck` on Linux should be entirely unaffected.

## What the changes touch (high level)

Roughly four buckets, **9 atomic commits, 23 files, +280 / −67**.

1. **Two universally-beneficial fixes** for clang strictness that
   *happen* to unblock macOS:
   - Drop the gcc-only `-Werror -Wextra -Wpedantic` triple from
     `gporca.mk` **on darwin only**. clang reports far more
     diagnostics than gcc under these flags and breaks the build
     for no real safety benefit (per-feature `-Werror=` flags in
     the base CXXFLAGS still catch real bugs). Linux retains the
     original triple.
   - Fix a typo in `init_var_from_str`'s prototype
     (`extern const bool` → `extern bool`) that gcc silently
     tolerated but clang correctly flags as a conflicting type
     against the definition. Identical compilation on gcc;
     introduced by upstream PR
     *#392 Export numeric interface to public*.

2. **macOS link-time differences** vs Linux:
   - macOS `<sys/param.h>` doesn't define `HZ`; provide a 100
     fallback so `contrib/interconnect/udp/ic_udpifc.c` compiles.
     `#ifndef HZ` makes this a no-op on Linux glibc.
   - macOS `ld` rejects unresolved symbols in shared libraries
     by default. With `--enable-shared-postgres-backend`, the
     `libpostgres.so` recipe deliberately omits `main/main.o`
     but other objects reference its symbols (`progname`, …).
     Add `-Wl,-undefined,dynamic_lookup` only when
     `PORTNAME=darwin` so those resolve at load time.

3. **PG 16 ↔ Cloudberry DLSUFFIX cohesion** on darwin only.
   Upstream PG 16 commit `b55f62abb2c "Unify DLSUFFIX on Darwin"`
   changed macOS module suffix from `.so` to `.dylib`.
   Cloudberry's catalog SQL, cdb_init.d scripts, expected/*.out
   files still hard-code `$libdir/foo.so`. When PG sees an
   explicit `.so` it does NOT re-append `DLSUFFIX`, so catalog
   bootstrap breaks (`FATAL: could not access file "foo"`).
   This PR pins `DLSUFFIX=.so` on darwin via `src/template/darwin`
   (file loaded only on darwin), plus two small follow-ons that
   no-op on Linux:
   - `configure`'s Python-shared-library probe gets a `.dylib`
     fallback only when `$PORTNAME = darwin` (macOS `libpython`
     ships as `.dylib` regardless of module DLSUFFIX).
   - `src/test/regress/GNUmakefile` uses `$(DLSUFFIX)` instead
     of hardcoded `.so` for install/uninstall lines — on Linux
     `$(DLSUFFIX)` expands to `.so`, so this is a textual
     no-change after make-time substitution.

4. **PAX portability for non-Linux**. Four focused commits:
   - **liburing optional**: `configure` keeps the hard error on
     Linux (`case $host_os in linux*) AC_MSG_ERROR ;;`) but
     allows missing on non-Linux. PAX has a pread-based
     `SyncFastIO` fallback selected at runtime by
     `IOUringFastIO::available()`. C++ code wraps the
     `IOUringFastIO` class with `#ifdef __linux__`. Linux gets
     identical behavior.
   - **C++ portability**: `off64_t` typedef
     (`#ifdef __APPLE__`-only), `int64_t` ↔ `int64` signature
     alignment (identical type on Linux LP64 — same `long int`),
     defensive `#undef Min/Max/IsPowerOf2` around protobuf
     includes (no-op on Linux: if old protobuf, no abseil
     shadowing; if new protobuf, the same fix is needed there
     too), VLA → `std::vector` *only under clang* via
     `#if defined(__clang__)` (gcc keeps zero-overhead stack
     VLA), `google::protobuf::int64` → `int64_t` (typedef
     identical for older protobuf, required for v22+).
   - **cmake portability**: gcc-only flags moved behind
     `if(APPLE)/else()`; `BUILD_GTEST=OFF` only on darwin so
     Linux `make pax-unit-test` continues to work; `-luuid` only
     on Linux (`libSystem` provides it on macOS); on macOS only,
     build `pax` as a `MODULE` (Mach-O bundle) and link with
     `-bundle_loader <postgres> -undefined dynamic_lookup` so
     backend globals have a single shared instance between
     postgres and pax.so; abseil deps via pkg-config on macOS
     (Homebrew protobuf v22+ splits them); `@loader_path`
     instead of `$ORIGIN` on macOS. Linux uses the `else`
     branches in all cases — original behavior verbatim.
   - **paxformat**: the standalone reader. On macOS add
     `-Wl,-undefined,dynamic_lookup` to defer PG-backend symbol
     references; Linux build path unchanged.

## Commits (kept atomic — order matters for builds)

```
Drop -Werror -Wextra -Wpedantic from ORCA C++ build
numeric: drop bogus 'const' qualifier on init_var_from_str
macOS: provide HZ fallback for UDP interconnect
macOS: defer libpostgres.so's undefined symbols to load time
macOS: pin DLSUFFIX=.so for PG 16 compatibility
macOS: make PAX's liburing dependency optional
macOS: PAX C++ portability
macOS: PAX cmake portability
macOS: build the standalone libpaxformat reader
```

## How I verified

Fresh checkout of this branch, on macOS 13.x x86_64 / Apple clang
14.0.3, with these Homebrew packages installed:

```
bison flex pkg-config libxml2 zstd openssl@3 readline apr apr-util
libevent xerces-c icu4c gettext libyaml lz4 perl protobuf libuv krb5
```

```sh
./configure --prefix=$HOME/install/cbdb-mac --enable-debug --enable-cassert \
  --enable-pax --enable-ic-proxy --enable-orafce --enable-tap-tests \
  --enable-mapreduce --enable-gpcloud \
  --with-perl --with-python --with-libxml --with-openssl \
  --with-lz4 --with-pam --with-ldap --with-gssapi \
  --with-apr-config=$(brew --prefix apr)/bin/apr-1-config \
  --with-includes=<homebrew prefixes joined by ':'> \
  --with-libs=<homebrew prefixes joined by ':'>

make -j$(sysctl -n hw.ncpu) \
  CUSTOM_COPT="-Wno-error=uninitialized -Wno-error=gnu-variable-sized-type-not-at-end"
make install
ln -sf $PREFIX/cloudberry-env.sh $PREFIX/greenplum_path.sh

cd gpAux/gpdemo
NUM_PRIMARY_MIRROR_PAIRS=3 make create-demo-cluster
```

After which:
- `gp_segment_configuration` shows 1 coordinator + 1 standby +
  3 primaries + 3 mirrors, all `status='u'` / `mode='s'`.
- `CREATE TABLE t USING pax` → insert 1000 rows → `SELECT
  count(*)` returns 1000 (ORCA plan used).
- `orafce`, `plperl`, `plpython3u` extensions install + work.
- `gpstate -s` reports `Mirror status = Synchronized` for all
  three mirrors.

Linux build path was not re-verified in this PR (no Linux machine
handy), but is structurally unchanged — see "Strict guarantee"
section above. Welcome any CI run to confirm.

## Caveats / out of scope

- **PAX's `io_uring` fast path stays Linux-only.** macOS has no
  equivalent. On Linux: unchanged. On macOS: PAX falls back to
  the existing pread-based `SyncFastIO`.
- **`uuid-ossp` extension** is not enabled in the configure
  command above because Homebrew `e2fsprogs` / `ossp-uuid` are
  heavy to pull in. Orthogonal to this PR.
- **GSSAPI on macOS requires Homebrew MIT `krb5`** in `PATH` /
  `PKG_CONFIG_PATH`; macOS's system Heimdal lacks
  `gss_store_cred_into` (used by PG 16). Environment setup
  detail, no code change in this PR.

### Type of Change
- [ ] Documentation update

### Test Plan

### Impact

**Performance:**

**User-facing changes:**

**Dependencies:**

### Checklist

### CI Skip Instructions

---

### DIFF
--- a/configure
+++ b/configure
@@ -9891,7 +9891,16 @@ _ACEOF
   LIBS="-luring $LIBS"
 
 else
-  as_fn_error $? "library 'uring' is required for PAX support" "$LINENO" 5
+  # liburing is required on Linux (io_uring is a Linux 5.1+ kernel iface);
+  # on non-Linux hosts PAX falls back to pread-based SyncFastIO.
+  case $host_os in
+    linux*)
+      as_fn_error $? "library 'uring' is required for PAX support on Linux" "$LINENO" 5
+      ;;
+    *)
+      :
+      ;;
+  esac
 fi
 
 
--- a/configure
+++ b/configure
@@ -12977,13 +12977,24 @@ else
 	fi
 	# Search for a likely-looking file.
 	found_shlib=0
+	# On Darwin, Python ships its shared lib as .dylib regardless of
+	# what DLSUFFIX is set to for modules. Try .dylib in addition to
+	# the configured DLSUFFIX so this check still finds libpython.
+	if test "$PORTNAME" = darwin; then
+		python_shlib_suffixes="$DLSUFFIX .dylib"
+	else
+		python_shlib_suffixes="$DLSUFFIX"
+	fi
 	for d in "${python_libdir}" "${python_configdir}" /usr/lib64 /usr/lib
 	do
-		if test -e "$d/lib${ldlibrary}${DLSUFFIX}"; then
-			python_libdir="$d"
-			found_shlib=1
-			break 2
-		fi
+		for s in $python_shlib_suffixes
+		do
+			if test -e "$d/lib${ldlibrary}${s}"; then
+				python_libdir="$d"
+				found_shlib=1
+				break 3
+			fi
+		done
 	done
 	# Some platforms (OpenBSD) require us to accept a bare versioned shlib
 	# (".so.n.n") as well. However, check this only after failing to find
--- a/configure.ac
+++ b/configure.ac
@@ -1035,9 +1035,19 @@ if test "$enable_pax" = yes; then
     [AC_MSG_ERROR([libzstd >= 1.4.0 is required for PAX support])]
   )
 
-  # Check liburing
-  AC_CHECK_LIB(uring, io_uring_queue_init, [],
-               [AC_MSG_ERROR([library 'uring' is required for PAX support])])
+  # Check liburing. liburing is Linux-only (io_uring kernel iface, 5.1+).
+  # On Linux it is required as before; on non-Linux hosts (macOS / *BSD)
+  # PAX falls back to pread-based SyncFastIO and the IOUringFastIO path
+  # is conditionally compiled — see contrib/pax_storage/src/cpp/comm/fast_io.*
+  case $host_os in
+    linux*)
+      AC_CHECK_LIB(uring, io_uring_queue_init, [],
+                   [AC_MSG_ERROR([library 'uring' is required for PAX support on Linux])])
+      ;;
+    *)
+      AC_CHECK_LIB(uring, io_uring_queue_init, [], [])
+      ;;
+  esac
 
   # Check cmake >= 3.11.0 using AX_COMPARE_VERSION
   AC_PATH_PROG([CMAKE], [cmake], [no])
--- a/contrib/interconnect/udp/ic_udpifc.c
+++ b/contrib/interconnect/udp/ic_udpifc.c
@@ -131,6 +131,13 @@ WSAPoll(
 #define SEC_TO_MSEC(t)                  ((t) * 1000)
 #define MSEC_TO_USEC(t)                 ((t) * 1000)
 #define USEC_TO_SEC(t)                  ((t) / 1000000)
+
+/* HZ is the kernel timer frequency. Linux defines it in <asm/param.h>
+ * (typically 100). macOS's sys/param.h does not, so provide a fallback. */
+#ifndef HZ
+#define HZ 100
+#endif
+
 #define TIME_TICK (1000000/HZ)/* in us */
 
 #define UDP_INITIAL_RTO                 (MSEC_TO_USEC(200))
--- a/contrib/interconnect/udp/ic_udpifc.c
+++ b/contrib/interconnect/udp/ic_udpifc.c
@@ -835,7 +835,7 @@ static void sendOnce(ChunkTransportState *transportStates, ChunkTransportStateEn
 static inline uint64 computeExpirationPeriod(MotionConn *conn, uint32 retry);
 
 static ICBuffer *getSndBuffer(MotionConn *conn);
-static void initSndBufferPool();
+static void initSndBufferPool(SendBufferPool *p);
 
 static void putIntoUnackQueueRing(UnackQueueRing *uqr, ICBuffer *buf, uint64 expTime, uint64 now);
 static void initUnackQueueRing(UnackQueueRing *uqr);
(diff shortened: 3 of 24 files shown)


zhangwenchao-123 (approved):

jiaqizho (approved):
/lgtm

PR #1795, opened 2026-06-02
Add predefined role pg_manage_resource_groups

Alena0704:
Add predefined role pg_manage_resource_groups
In Cloudberry/Greenplum only a superuser can manage resource groups.
In managed-service deployments superuser cannot be granted to the
client, so the cloud admin has no way to tune their own CPU and
memory limits.

Add a predefined role pg_manage_resource_groups (OID 6300) in
pg_authid.dat, following the upstream convention used by pg_monitor
and pg_read_all_data. Members of this role can CREATE, ALTER and DROP
resource groups and call pg_resgroup_move_query() without being a
superuser. Delegation is done with standard SQL:

    GRANT pg_manage_resource_groups TO <user>;

The regression and isolation2 tests exercise the new role via GRANT.
The basic CREATE/ALTER/DROP test cases in the regression test are
adapted from open-gpdb/gpdb commit 3ac99962ad2.

Co-authored-by: Andrey Borodin <x4mmm@yandex-team.ru>

Fixes #ISSUE_Number

### Type of Change
- [ ] Bug fix (non-breaking change)
- [x] New feature (non-breaking change)
- [ ] Breaking change (fix or feature with breaking changes)
- [x] Documentation update

### Test Plan

- [ ] Unit tests added/updated
- [x] Integration tests added/updated
- [x] Passed `make installcheck`
- [x] Passed `make -C src/test installcheck-cbdb-parallel`

### Impact

**Performance:**

**User-facing changes:**

**Dependencies:**

### Checklist
- [x] Followed [contribution guide](https://cloudberry.apache.org/contribute/code)
- [x] Added/updated documentation
- [x] Reviewed code for security implications
- [ ] This PR contains AI-assisted code generation
- [x] Requested review from [cloudberry committers](https://github.com/orgs/apache/teams/cloudberry-committers)

### CI Skip Instructions

---

PR #1804, opened 2026-06-04
Let mdb_admin manage resource groups

Alena0704:
Let mdb_admin manage resource groups
Add gp_toolkit.pg_create_mdb_admin_role() (gp_toolkit 1.7) to provision the
role with its fixed OID.

Allow members of mdb_admin — not just superusers — to CREATE/ALTER/DROP
resource groups and run pg_resgroup_move_query(), so a cloud admin can tune
their own CPU/memory limits.

Based on pg-sharding/cpg 7b8c912. Some tests are adapted from
open-gpdb/gpdb 3ac99962ad2.

Co-authored-by: reshke<reshke@double.cloud>
Co-authored-by: Andrey Borodin <x4mmm@yandex-team.ru>

Fixes #ISSUE_Number

### Type of Change
- [x] Documentation update

### Test Plan

### Impact

**Performance:**

**User-facing changes:**

**Dependencies:**

### Checklist

### CI Skip Instructions

---

### DIFF
--- a/src/backend/commands/resgroupcmds.c
+++ b/src/backend/commands/resgroupcmds.c
@@ -34,6 +34,7 @@
 #include "commands/resgroupcmds.h"
 #include "miscadmin.h"
 #include "nodes/pg_list.h"
+#include "utils/acl.h"
 #include "utils/builtins.h"
 #include "utils/datetime.h"
 #include "utils/fmgroids.h"
@@ -103,11 +104,11 @@ CreateResourceGroup(CreateResourceGroupStmt *stmt)
 	int			nResGroups;
 	MemoryContext oldContext;
 
-	/* Permission check - only superuser can create groups. */
-	if (!superuser())
+	/* Permission check - only superuser or mdb_admin can create groups. */
+	if (!is_member_of_role(GetUserId(), MDB_ADMIN_ROLEID))
 		ereport(ERROR,
 				(errcode(ERRCODE_INSUFFICIENT_PRIVILEGE),
-				 errmsg("must be superuser to create resource groups")));
+				 errmsg("must be mdb_admin to create resource groups")));
 
 	/*
 	 * Check for an illegal name ('none' is used to signify no group in ALTER ROLE).
@@ -269,11 +270,11 @@ DropResourceGroup(DropResourceGroupStmt *stmt)
 	Oid			 groupid;
 	ResourceGroupCallbackContext	*callbackCtx;
 
-	/* Permission check - only superuser can drop resource groups. */
-	if (!superuser())
+	/* Permission check - only superuser or mdb_admin can drop resource groups. */
+	if (!is_member_of_role(GetUserId(), MDB_ADMIN_ROLEID))
 		ereport(ERROR,
 				(errcode(ERRCODE_INSUFFICIENT_PRIVILEGE),
-				 errmsg("must be superuser to drop resource groups")));
+				 errmsg("must be mdb_admin to drop resource groups")));
 
 	/*
 	 * Check the pg_resgroup relation to be certain the resource group already
@@ -301,6 +302,13 @@ DropResourceGroup(DropResourceGroupStmt *stmt)
 	 * resource group later.
 	 */
 	groupid = ((Form_pg_resgroup) GETSTRUCT(tuple))->oid;
+
+	/* Permission check - only superuser can drop the admin/system resource groups. */
+	if (!superuser() && (groupid == ADMINRESGROUP_OID || groupid == SYSTEMRESGROUP_OID))
+		ereport(ERROR,
+				(errcode(ERRCODE_INSUFFICIENT_PRIVILEGE),
+				 errmsg("must be superuser to drop resource group \"%s\"",
+						stmt->name)));
 
 	/* cannot DROP default resource groups  */
 	if (groupid == DEFAULTRESGROUP_OID
@@ -375,11 +383,24 @@ AlterResourceGroup(AlterResourceGroupStmt *stmt)
 	ResourceGroupCallbackContext	*callbackCtx;
 	MemoryContext oldContext;
 
-	/* Permission check - only superuser can alter resource groups. */
-	if (!superuser())
+	/* Permission check - only mdb_admin can alter resource groups. */
+	if (!is_member_of_role(GetUserId(), MDB_ADMIN_ROLEID))
 		ereport(ERROR,
 				(errcode(ERRCODE_INSUFFICIENT_PRIVILEGE),
-				 errmsg("must be superuser to alter resource groups")));
+				 errmsg("must be mdb_admin to alter resource groups")));
+
+	/*
+	 * Check the pg_resgroup relation to be certain the resource group already
+	 * exists.
+	 */
+	groupid = get_resgroup_oid(stmt->name, false);
+
+	/* Permission check - only superuser can alter the admin/system resource groups. */
+	if (!superuser() && (groupid == ADMINRESGROUP_OID || groupid == SYSTEMRESGROUP_OID))
+		ereport(ERROR,
+				(errcode(ERRCODE_INSUFFICIENT_PRIVILEGE),
+				 errmsg("must be superuser to alter resource group \"%s\"",
+						stmt->name)));
 
 	/* Currently we only support to ALTER one limit at one time */
 	Assert(list_length(stmt->options) == 1);
@@ -405,12 +426,6 @@ AlterResourceGroup(AlterResourceGroupStmt *stmt)
 		value = getResgroupOptionValue(defel);
 		checkResgroupCapLimit(limitType, value);
 	}
-
-	/*
-	 * Check the pg_resgroup relation to be certain the resource group already
-	 * exists.
-	 */
-	groupid = get_resgroup_oid(stmt->name, false);
 
 	if (limitType == RESGROUP_LIMIT_TYPE_CONCURRENCY &&
 		value == 0 &&
@@ -500,7 +515,7 @@ AlterResourceGroup(AlterResourceGroupStmt *stmt)
 									  RESGROUP_DEFAULT_CPU_WEIGHT, "");
 
 		updateResgroupCapabilityEntry(pg_resgroupcapability_rel,
-									  groupid, RESGROUP_LIMIT_TYPE_CPUSET, 
+									  groupid, RESGROUP_LIMIT_TYPE_CPUSET,
 									  0, caps.cpuset);
 	}
 	else if (limitType == RESGROUP_LIMIT_TYPE_CPU)
@@ -1007,7 +1022,7 @@ parseStmtOptions(CreateResourceGroupStmt *stmt, ResGroupCaps *caps)
 		else
 			mask |= 1 << type;
 
-		if (type == RESGROUP_LIMIT_TYPE_CPUSET) 
+		if (type == RESGROUP_LIMIT_TYPE_CPUSET)
 		{
 			const char *cpuset = defGetString(defel);
 			strlcpy(caps->cpuset, cpuset, sizeof(caps->cpuset));
@@ -1611,7 +1626,7 @@ checkCpuSetByRole(const char *cpuset)
  * ex:
  * cpuset = "1;4"
  * then we should assign '1' to corrdinator and '4' to segment
- * 
+ *
  * cpuset = "1"
  * assign '1' to both coordinator and segment
  */
--- a/gpcontrib/gp_toolkit/Makefile
+++ b/gpcontrib/gp_toolkit/Makefile
@@ -1,15 +1,15 @@
 EXTENSION = gp_toolkit
 DATA = gp_toolkit--1.1--1.2.sql gp_toolkit--1.0--1.1.sql gp_toolkit--1.0.sql \
 		gp_toolkit--1.2--1.3.sql gp_toolkit--1.3.sql gp_toolkit--1.3--1.4.sql \
-		gp_toolkit--1.4--1.5.sql gp_toolkit--1.5--1.6.sql
+		gp_toolkit--1.4--1.5.sql gp_toolkit--1.5--1.6.sql gp_toolkit--1.6--1.7.sql
 MODULE_big = gp_toolkit
 ifeq ($(shell uname -s), Linux)
-OBJS = resgroup.o gp_partition_maint.o
+OBJS = resgroup.o gp_partition_maint.o mdb_admin.o
 else
-OBJS = resgroup-dummy.o gp_partition_maint.o
+OBJS = resgroup-dummy.o gp_partition_maint.o mdb_admin.o
 endif
 
-REGRESS = resource_manager_restore_to_none gp_toolkit resource_manager_switch_to_queue gp_toolkit_resqueue gp_toolkit_ao_funcs gp_partition_maint
+REGRESS = resource_manager_restore_to_none gp_toolkit gp_toolkit_mdb_admin resource_manager_switch_to_queue gp_toolkit_resqueue gp_toolkit_ao_funcs gp_partition_maint
 EXTRA_REGRESS_OPTS = --init-file=$(top_builddir)/src/test/regress/init_file
 
 ifdef USE_PGXS
--- a/gpcontrib/gp_toolkit/expected/gp_toolkit_mdb_admin.out
+++ b/gpcontrib/gp_toolkit/expected/gp_toolkit_mdb_admin.out
@@ -1,0 +1,46 @@
+-- Tests for gp_toolkit.pg_create_mdb_admin_role(): creation of the
+-- fixed-OID mdb_admin role and the resource-group permission gate it
+-- enables.  gp_toolkit is installed in template1 by gpinitsystem, so the
+-- extension is already present in this database.
+-- ---------------------------------------------------------------------
+-- pg_create_mdb_admin_role() creates the mdb_admin role with its fixed OID.
+-- ---------------------------------------------------------------------
+SELECT gp_toolkit.pg_create_mdb_admin_role() AS mdb_admin_oid;
+ mdb_admin_oid 
+---------------
+          8067
+(1 row)
+
+-- The role exists with the fixed OID and is a non-login, non-superuser,
+-- connection-limited role.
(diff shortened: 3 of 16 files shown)


src/backend/commands/resgroupcmds.c
```diff
@@ -375,11 +391,22 @@ AlterResourceGroup(AlterResourceGroupStmt *stmt)
...
+	 * pg_manage_resource_groups can alter resource groups.
+	 */
+	if (!has_privs_of_role(GetUserId(), ROLE_PG_MANAGE_RESOURCE_GROUPS))
+		ereport(ERROR,
+				(errcode(ERRCODE_INSUFFICIENT_PRIVILEGE),
+				 errmsg("permission denied to alter resource group \"%s\"", stmt->name),
+				 errhint("Must be superuser or have privileges of the pg_manage_resource_groups role.")));
+
+	/* The system admin_group can only be altered by a superuser. */
+	if (!superuser() && (strcmp(stmt->name, "admin_group") == 0))
```
my-ship-it: Better to use OID comparison instead of string?

src/include/catalog/pg_authid.dat
```diff
@@ -79,5 +79,10 @@
   rolcreaterole => 'f', rolcreatedb => 'f', rolcanlogin => 'f',
   rolreplication => 'f', rolbypassrls => 'f', rolconnlimit => '-1',
   rolpassword => '_null_', rolvaliduntil => '_null_' },
+{ oid => '6312', oid_symbol => 'ROLE_PG_MANAGE_RESOURCE_GROUPS',
```
my-ship-it: Can ROLE_PG_MANAGE_RESOURCE_GROUPS be dropped?  If dropped, we cannot create it again.

src/include/catalog/pg_authid.dat
```diff
@@ -79,5 +79,10 @@
...
+  rolname => 'pg_manage_resource_groups', rolsuper => 'f', rolinherit => 't',
```
my-ship-it: We change catalog here, but for 2X_STABLE, we suppose not.

src/backend/commands/resgroupcmds.c
```diff
@@ -375,11 +391,22 @@ AlterResourceGroup(AlterResourceGroupStmt *stmt)
...
 		ereport(ERROR,
```
my-ship-it, line 400: Should we also protect SYSTEMRESGROUP_OID?

src/backend/commands/resgroupcmds.c
Alena0704: Agree

src/include/catalog/pg_authid.dat
Alena0704: Yes, agree and because of this I'm going to return to this solution https://github.com/apache/cloudberry/pull/1763 but I will save tests that I have added before here. But I will take into account your current suggestions here https://github.com/apache/cloudberry/pull/1795

src/backend/commands/resgroupcmds.c
Alena0704: I think we should compare it with admin_group and default_group too. because the first one is a default group for superuser sessions and default_group contains system roles without superuser privileges.

src/include/catalog/pg_authid.dat
Alena0704: We can't drop this role. This function checks it:

```
/* This case can be dispatched quickly */
	if (IsPinnedObject(classId, objectId))
	{
		object.classId = classId;
		object.objectId = objectId;
		object.objectSubId = 0;
		ereport(ERROR,
				(errcode(ERRCODE_DEPENDENT_OBJECTS_STILL_EXIST),
				 errmsg("cannot drop %s because it is required by the database system",
						getObjectDescription(&object, false))));
	}
```
src/backend/catalog/pg_shdepend.c:654

Alena0704:
I rewrote the patch completely because a new bootstrap catalog entry bumps CATALOG_VERSION_NO and incompatible with minor gpupgrade - I adapted the commit from open-gpdb [3ac99962ad2](https://github.com/open-gpdb/gpdb/commit/3ac99962ad2) and added my tests.

src/include/catalog/pg_authid.dat
my-ship-it: > Yes, agree and because of this I'm going to return to this solution #1763 but I will save tests that I have added before here. But I will take into account your current suggestions here #1795

Got it, thanks for the explaination.

Alena0704: It is ready - you can take a look at it

src/backend/commands/resgroupcmds.c
Alena0704: I have added the check only for system and admin groups here. Only here super roles can be found. I save this check only for alter operation. For drop operation it still compares with ‘mdb_admin’ to avoid unnecessary scanning catalog to find particular record to delete it - you can notice it checks the same.

my-ship-it:
This PR's permission check looks up the role by name, not by a fixed OID:
```
  role = get_role_oid("mdb_admin", true);   /* look up the role by the name "mdb_admin" */
  if (!is_member_of_role(GetUserId(), role))
      ereport(ERROR, ...);
```

This means: members of any role that happens to be named mdb_admin can manage resource groups. The code doesn't care who created that role or when.

Where the hole is: This assumption overlooks CREATEROLE users. Suppose a DBA has granted some user CREATEROLE (quite common on self-hosted clusters,
  e.g. to let someone manage ordinary user accounts):

```
  -- Executed as an ordinary CREATEROLE user:
  CREATE ROLE mdb_admin;          -- CREATEROLE allows creating any non-superuser role ✓
  GRANT mdb_admin TO myself;      -- Under PG14 semantics, a CREATEROLE user can grant
                                  -- any non-superuser role to anyone, including themselves ✓
  ALTER RESOURCE GROUP default_group SET concurrency 1;  -- now passes the permission check!
```
  Three SQL statements, and a user who was only supposed to "manage accounts" has self-granted resource group administration rights.

Alena0704:
> This PR's permission check looks up the role by name, not by a fixed OID:
> 
> ```
>   role = get_role_oid("mdb_admin", true);   /* look up the role by the name "mdb_admin" */
>   if (!is_member_of_role(GetUserId(), role))
>       ereport(ERROR, ...);
> ```
> 
> This means: members of any role that happens to be named mdb_admin can manage resource groups. The code doesn't care who created that role or when.
> 
> Where the hole is: This assumption overlooks CREATEROLE users. Suppose a DBA has granted some user CREATEROLE (quite common on self-hosted clusters, e.g. to let someone manage ordinary user accounts):
> 
> ```
>   -- Executed as an ordinary CREATEROLE user:
>   CREATE ROLE mdb_admin;          -- CREATEROLE allows creating any non-superuser role ✓
>   GRANT mdb_admin TO myself;      -- Under PG14 semantics, a CREATEROLE user can grant
>                                   -- any non-superuser role to anyone, including themselves ✓
>   ALTER RESOURCE GROUP default_group SET concurrency 1;  -- now passes the permission check!
> ```
> 
> Three SQL statements, and a user who was only supposed to "manage accounts" has self-granted resource group administration rights.

Agree, I have fixed it by checking that it is included in admin and system group.

reshke:
> Where the hole is: This assumption overlooks CREATEROLE users. Suppose a DBA has granted some user CREATEROLE (quite common on self-hosted clusters, e.g. to let someone manage ordinary user accounts):
> 
> ```
>   -- Executed as an ordinary CREATEROLE user:
>   CREATE ROLE mdb_admin;          -- CREATEROLE allows creating any non-superuser role ✓
>   GRANT mdb_admin TO myself;      -- Under PG14 semantics, a CREATEROLE user can grant
>                                   -- any non-superuser role to anyone, including themselves ✓
>   ALTER RESOURCE GROUP default_group SET concurrency 1;  -- now passes the permission check!
> ```
> 
> Three SQL statements, and a user who was only supposed to "manage accounts" has self-granted resource group administration rights.

for our use in Cloud we just do not consider this as privilege escalation, because we grant this role to cloud users(non-superuser) by request (you just click in UI like that this role to this role)

Anyway issue looks valid in general use

reshke:
> This PR's permission check looks up the role by name, not by a fixed OID:
> 
> ```
>   role = get_role_oid("mdb_admin", true);   /* look up the role by the name "mdb_admin" */
>   if (!is_member_of_role(GetUserId(), role))
>       ereport(ERROR, ...);
> ```
> 
> This means: members of any role that happens to be named mdb_admin can manage resource groups. The code doesn't care who created that role or when.

Yep, exactly :) We only care about rolsuper or not, if mdb-admin patch does not allow to gain superuser priviledge - we consider it safe :)

my-ship-it:
> > This PR's permission check looks up the role by name, not by a fixed OID:
> > ```
> >   role = get_role_oid("mdb_admin", true);   /* look up the role by the name "mdb_admin" */
> >   if (!is_member_of_role(GetUserId(), role))
> >       ereport(ERROR, ...);
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This means: members of any role that happens to be named mdb_admin can manage resource groups. The code doesn't care who created that role or when.
> 
> Yep, exactly :) We only care about rolsuper or not, if mdb-admin patch does not allow to gain superuser priviledge - we consider it safe :)

Thanks for the iterations on this. I agree with the constraint driving the current design — a
new bootstrap entry in pg_authid.dat bumps CATALOG_VERSION_NO and breaks minor-version upgrades,
so a runtime lookup is the right shape for the stable branch.

How about Proposal: use a reserved-prefix role name

Rename the role to e.g. pg_manage_resgroup and keep the same runtime get_role_oid() lookup. The
point is that the pg_ namespace is already reserved for roles:

- CreateRole() rejects any pg_-prefixed name via IsReservedName() (src/backend/commands/user.c),
so a CREATEROLE user cannot forge it;
- RenameRole() rejects renaming any role to a pg_ name, so it cannot be smuggled in via ALTER
ROLE ... RENAME either;

One caveat: DROP is not name-protected

Alena0704:
> > > This PR's permission check looks up the role by name, not by a fixed OID:
> > > ```
> > >   role = get_role_oid("mdb_admin", true);   /* look up the role by the name "mdb_admin" */
> > >   if (!is_member_of_role(GetUserId(), role))
> > >       ereport(ERROR, ...);
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > This means: members of any role that happens to be named mdb_admin can manage resource groups. The code doesn't care who created that role or when.
> > 
> > 
> > Yep, exactly :) We only care about rolsuper or not, if mdb-admin patch does not allow to gain superuser priviledge - we consider it safe :)
> 
> Thanks for the iterations on this. I agree with the constraint driving the current design — a new bootstrap entry in pg_authid.dat bumps CATALOG_VERSION_NO and breaks minor-version upgrades, so a runtime lookup is the right shape for the stable branch.
> 
> How about Proposal: use a reserved-prefix role name
> 
> Rename the role to e.g. pg_manage_resgroup and keep the same runtime get_role_oid() lookup. The point is that the pg_ namespace is already reserved for roles:
> 
> * CreateRole() rejects any pg_-prefixed name via IsReservedName() (src/backend/commands/user.c),
>   so a CREATEROLE user cannot forge it;
> * RenameRole() rejects renaming any role to a pg_ name, so it cannot be smuggled in via ALTER
>   ROLE ... RENAME either;
> 
> One caveat: DROP is not name-protected

We can't use a pg_ name because creating it requires changes in pg_authid.dat (a bootstrap role), which bumps CATALOG_VERSION_NO and breaks in-place minor upgrades.

postgres=# CREATE ROLE pg_manage_resgroup;
ERROR:  role name "pg_manage_resgroup" is reserved
DETAIL:  Role names starting with "pg_" are reserved.

I added in the last commit a check in user.c that restricts CREATE/ALTER/RENAME/DROP/GRANT/REVOKE of the mdb_admin and mdb_superuser roles to superusers only, so an ordinary CREATEROLE user cannot hijack the by-name privilege gate (create, rename, reset the password of, or self-grant the role) and thereby escalate privileges.

CREATE ROLE rg_attacker CREATEROLE;
SET ROLE rg_attacker;
GRANT mdb_admin TO rg_attacker;                    -- self-grant membership
DROP ROLE mdb_admin;                               -- drop the gate
ALTER ROLE mdb_admin RENAME TO rg_attacker_admin;  -- move the real role aside
ALTER ROLE mdb_admin PASSWORD 'hijack';            -- take it over via password reset
CREATE ROLE mdb_superuser;                         -- create a protected role
RESET ROLE;
DROP ROLE rg_attacker;

CREATE ROLE
SET
ERROR:  must be superuser to grant membership in role "mdb_admin"
ERROR:  must be superuser to drop role "mdb_admin"
ERROR:  must be superuser to rename role "mdb_admin"
ERROR:  must be superuser to alter role "mdb_admin"
ERROR:  must be superuser to create role "mdb_superuser"
RESET
DROP ROLE

Alena0704:
Reworked the patch following the pg_aux commit - moved the mdb_admin role into a contrib extension, pg_aux_catalog: it's created via pg_create_mdb_admin_role() (fixed OID 8067, superuser-only). Resource-group operations are gated on mdb_admin membership, identified by that fixed OID.

Because the role is created at a single fixed OID and only by a superuser, it cannot be re-created or taken over by another user - a CREATEROLE user can neither establish mdb_admin nor recreate it at that OID:

```
postgres=# CREATE ROLE rg_attacker CREATEROLE;
CREATE ROLE
postgres=# SET ROLE rg_attacker;
SET
postgres=> ALTER ROLE mdb_admin RENAME TO rg_attacker_admin;
ERROR:  role "mdb_admin" does not exist
postgres=> ALTER ROLE mdb_admin PASSWORD 'hijack';
ERROR:  role "mdb_admin" does not exist
```
Tests were moved into the extension as well.

my-ship-it:
> Reworked the patch following the pg_aux commit - moved the mdb_admin role into a contrib extension, pg_aux_catalog: it's created via pg_create_mdb_admin_role() (fixed OID 8067, superuser-only). Resource-group operations are gated on mdb_admin membership, identified by that fixed OID.
> 
> Because the role is created at a single fixed OID and only by a superuser, it cannot be re-created or taken over by another user - a CREATEROLE user can neither establish mdb_admin nor recreate it at that OID:
> 
> ```
> postgres=# CREATE ROLE rg_attacker CREATEROLE;
> CREATE ROLE
> postgres=# SET ROLE rg_attacker;
> SET
> postgres=> ALTER ROLE mdb_admin RENAME TO rg_attacker_admin;
> ERROR:  role "mdb_admin" does not exist
> postgres=> ALTER ROLE mdb_admin PASSWORD 'hijack';
> ERROR:  role "mdb_admin" does not exist
> ```
> 
> Tests were moved into the extension as well.

Thanks for the rework — moving to a fixed-OID role provisioned through contrib/pg_aux_catalog is the right call,
and I think it closes the escalation path we were discussing. A few notes:

Security model — LGTM. The original concern (a CREATEROLE user forging a role named mdb_admin and
self-granting it) is now moot, because the gate checks membership in OID 8067 rather than the name:

- A plain CREATE ROLE mdb_admin gets an ordinary OID (≥ FirstNormalObjectId), never 8067, so a forged role is inert.
- next_aux_pg_authid_oid is only set inside the superuser-only function, and 8067 sits in the reserved range that normal OID allocation never hands out.
- I confirmed is_member_of_role() short-circuits for superusers (src/backend/utils/adt/acl.c), so when the extension isn't installed (OID 8067 absent) the behavior cleanly degrades to the previous superuser-only semantics, and superusers never lose access. Nice fallback.

Alena0704:
> > Reworked the patch following the pg_aux commit - moved the mdb_admin role into a contrib extension, pg_aux_catalog: it's created via pg_create_mdb_admin_role() (fixed OID 8067, superuser-only). Resource-group operations are gated on mdb_admin membership, identified by that fixed OID.
> > Because the role is created at a single fixed OID and only by a superuser, it cannot be re-created or taken over by another user - a CREATEROLE user can neither establish mdb_admin nor recreate it at that OID:
> > ```
> > postgres=# CREATE ROLE rg_attacker CREATEROLE;
> > CREATE ROLE
> > postgres=# SET ROLE rg_attacker;
> > SET
> > postgres=> ALTER ROLE mdb_admin RENAME TO rg_attacker_admin;
> > ERROR:  role "mdb_admin" does not exist
> > postgres=> ALTER ROLE mdb_admin PASSWORD 'hijack';
> > ERROR:  role "mdb_admin" does not exist
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Tests were moved into the extension as well.
> 
> Thanks for the rework — moving to a fixed-OID role provisioned through contrib/pg_aux_catalog is the right call, and I think it closes the escalation path we were discussing. A few notes:
> 
> Security model — LGTM. The original concern (a CREATEROLE user forging a role named mdb_admin and self-granting it) is now moot, because the gate checks membership in OID 8067 rather than the name:
> 
> * A plain CREATE ROLE mdb_admin gets an ordinary OID (≥ FirstNormalObjectId), never 8067, so a forged role is inert.
> * next_aux_pg_authid_oid is only set inside the superuser-only function, and 8067 sits in the reserved range that normal OID allocation never hands out.
> * I confirmed is_member_of_role() short-circuits for superusers (src/backend/utils/adt/acl.c), so when the extension isn't installed (OID 8067 absent) the behavior cleanly degrades to the previous superuser-only semantics, and superusers never lose access. Nice fallback.

Thanks! Yes, that's the intent - the check is on membership in OID 8067, and only a superuser can create the role, so a plain CREATE ROLE mdb_admin is harmless. It's all in the current commit (ae95005).

leborchuk (approved):
LGTM, checked on internal Cloudberry clusters

fairyfar:
It seems unnecessary to add the new extension `pg_aux_catalog`.
Perhaps the newly added interface SQL can be added to `gp_toolkit`.

Alena0704:
> It seems unnecessary to add the new extension `pg_aux_catalog`. Perhaps the newly added interface SQL can be added to `gp_toolkit`.

Agree. I moved it there and updated commit message

Smyatkin-Maxim (approved):

PR #1823, opened 2026-06-18
Automatically generate node support functions in cloudberry

chenjinbao1989:
When merging Postgres 16, automatic generation of node support
functions was not enabled. This commit enables the automatic
generation of write, read, copy, and equal functions.

In future code, developers will no longer need to manually create 
NodeTags and their corresponding write/read/copy/equal functions 
when creating or modifying structures.


src/include/commands/explain_gp.h
```diff
@@ -0,0 +1,81 @@
+/*-------------------------------------------------------------------------
+*
+ * explain_gp.h
+ *
+ * Portions Copyright (c) 2026, Cloudberry, Inc.
```
tuhaihe: Hi, there is no `Cloudberry, Inc`. For the new files created by the Cloudberry community members, we should add the standard ASF license headers. Can take the following as a reference:

https://github.com/apache/cloudberry/blob/3a64dbce00df4ef7801aad253a1a9cbe49b23626/contrib/pax_storage/src/cpp/pax_gbench.cc#L1-L26

src/backend/executor/execAmi.c
```diff
@@ -226,7 +226,6 @@ ExecReScan(PlanState *node)
 			break;
 
 		case T_DynamicIndexScanState:
-		case T_DynamicIndexOnlyScanState:
```
yjhjstz, line 229: why remove `T_DynamicIndexOnlyScanState` ?

src/include/commands/explain_gp.h
chenjinbao1989: Got it.

src/backend/executor/execAmi.c
chenjinbao1989: In fact, there is no such struct named DynamicIndexOnlyScanState. I suppose this is a legacy issue, and we can just take this opportunity to remove it by the auto-generated NodeTag feature.

tuhaihe:
The new file now includes the standard ASF license header and has passed the RAT scan. It looks fine to me. Thanks!

tuhaihe:
Hi @chenjinbao1989 welcome to add the issue description to your commit message body for a better context.

You can take the [.gitmessage](https://github.com/apache/cloudberry/blob/main/.gitmessage) as a commit template.

chenjinbao1989:
> Hi @chenjinbao1989 welcome to add the issue description to your commit message body for a better context.
> 
> You can take the [.gitmessage](https://github.com/apache/cloudberry/blob/main/.gitmessage) as a commit template.

This pr does not come from a issue.  It is a feature from pg16.

tuhaihe:
> > Hi @chenjinbao1989 welcome to add the issue description to your commit message body for a better context.
> > You can take the [.gitmessage](https://github.com/apache/cloudberry/blob/main/.gitmessage) as a commit template.
> 
> This pr does not come from a issue. It is a feature from pg16.

Thanks for the clarification!

That's totally fine. My point wasn't necessarily about having an issue, but about providing enough context in the commit message.

For PostgreSQL merge work, adding the upstream background or the related PostgreSQL commit in the message body would make the history much easier to understand later.

tuhaihe:
If no more comments, could we start to merge this PR for PG16 FIXME?

leborchuk (approved):
LGTM

leborchuk:
We could copy description to commit message while squashing

tuhaihe (approved):
Not waiting for a response. Let's just merge it. If it's not OK, we can always revert it.

PR #1826, opened 2026-06-18
Fix gpexpand 'SyncPackages object has no attribute ret': enqueue OperationWorkerPool operations via items=

### LINKED ISSUE #1825: gpexpand first stage fails: 'SyncPackages' object has no attribute 'ret' (extension sync never runs; regression)
### Summary

During the first stage of `gpexpand` (initialization), `gpexpand` fails while "Syncing Apache Cloudberry extensions" with:

```
AttributeError: 'SyncPackages' object has no attribute 'ret'
```

The exception is squashed into a `WARNING`, so the expansion appears to continue, but **two things go wrong**:

1. The error traceback is printed and an `[ERROR]` line is logged.
2. More importantly, the package/extension synchronization to the new segment hosts **silently never runs at all** — even on a single-host expansion.

### Version / Environment

- Reproduced on a build based on current `apache/cloudberry` `main`.
- The buggy code is present on `main` today (see Root cause). Observed on a downstream `4.5.0-rc.3` build, but the defect is in shared `gpMgmt` code, not downstream-only.

### Steps to reproduce

```bash
gpexpand -i /tmp/expand_input.txt -v 2>&1 | tee /tmp/gpexpand_init.log
```

### Actual output (abridged, `-v`)

```
gpexpand:...-[INFO]:-Syncing Apache Cloudberry extensions
gpexpand:...-[DEBUG]:-Starting ParallelOperation
gpexpand:...-[DEBUG]:-WorkerPool() initialized with 1 workers
gpexpand:...-[DEBUG]:-WorkerPool haltWork()
gpexpand:...-[DEBUG]:-[worker0] haltWork
gpexpand:...-[DEBUG]:-Ending ParallelOperation
gpexpand:...-[DEBUG]:-[worker0] got a halt cmd
gpexpand:...-[ERROR]:-Syncing of Apache Cloudberry extensions has failed.
Traceback (most recent call last):
  File ".../bin/gpexpand", line 1935, in sync_packages
    operation.get_ret()
  File ".../lib/python/gppylib/operations/__init__.py", line 64, in get_ret
    if isinstance(self.ret, Exception):
AttributeError: 'SyncPackages' object has no attribute 'ret'
gpexpand:...-[WARNING]:-Please run gppkg --clean after successful expansion.
```

Note the telltale sequence: the worker pool is created, then `haltWork()` is called and the worker only ever receives a *halt* command — there is no `Starting SyncPackages` / `[worker0] got cmd` line. The submitted `SyncPackages` operation is **never executed**.

### Root cause

`WorkerPool.__init__` signature (`gpMgmt/bin/gppylib/commands/base.py`):

```python
def __init__(self, numWorkers=16, should_stop=False, items=None, daemonize=False, logger=...):
    ...
    if items is not None:
        for item in items:
            self.addCommand(item)
```

The second **positional** parameter is `should_stop`; the work items are the third parameter `items`.

`OperationWorkerPool.__init__` (same file) passes the operations **positionally**:

```python
class OperationWorkerPool(WorkerPool):
    def __init__(self, numWorkers=16, operations=None):
        if operations is not None:
            for operation in operations:
                self._spoof_operation(operation)
        super(OperationWorkerPool, self).__init__(numWorkers, operations)   # <-- BUG
```

So `operations` is bound to `should_stop` (a truthy list), and `items` stays `None`. Consequences:

1. `items is None` → **no operation is ever added to the work queue**.
2. `should_stop` is set to a truthy value.

`ParallelOperation.execute()` then `join()`s an empty queue (returns immediately) and `haltWork()`s the pool, so the worker only ever pops the halt command. `SyncPackages.run()` / `execute()` is never invoked, so `Operation.run()` never assigns `self.ret`. (`SyncPackages.__init__` does not call `super().__init__()`, so `ret` is not pre-initialized either.) Back in `gpexpand.sync_packages()`:

```python
for operation in operations:
    operation.get_ret()   # SyncPackages has no attribute 'ret' -> AttributeError
```

→ `AttributeError: 'SyncPackages' object has no attribute 'ret'`.

### This is a regression

This exact bug was already fixed once and then re-introduced by a merge:

- `cd3c88f6e1e` "Fix gppkg error: 'SyncPackages' object has no attribute 'ret'." changed the call to `super(OperationWorkerPool, self).__init__(numWorkers, items=operations)`.
- `0f4cf8d5068` "Merge tag 'REL_16_9' into Cloudberry" reverted that line back to `super(OperationWorkerPool, self).__init__(numWorkers, operations)`.

On current `main`, `grep -c "items=operations" gpMgmt/bin/gppylib/commands/base.py` returns `0` — the fix is gone again.

### Suggested fix

Pass the operations as the keyword argument `items` (re-applying the lost fix):

```python
super(OperationWorkerPool, self).__init__(numWorkers, items=operations)
```

This restores enqueueing of the operations and keeps `should_stop=False`.

### Impact

- `gpexpand` first stage prints an error/traceback even on a healthy cluster.
- Extensions/gppkg packages are **not** synchronized to new segment hosts during expansion (the operation never runs), which can leave new segments missing extensions until a manual `gppkg` sync. The squashing of the exception into a WARNING hides this.

talmacschen-arch:
## Summary

Fixes #1825.

`OperationWorkerPool.__init__` passed its `operations` **positionally** to `WorkerPool.__init__`:

```python
super(OperationWorkerPool, self).__init__(numWorkers, operations)
```

But `WorkerPool.__init__`'s second positional parameter is `should_stop`, and the work items are the third parameter `items`:

```python
def __init__(self, numWorkers=16, should_stop=False, items=None, ...):
    ...
    if items is not None:
        for item in items:
            self.addCommand(item)
```

So `operations` bound to `should_stop` (a truthy list) and `items` stayed `None`. Consequently **no operation was ever added to the work queue**, `should_stop` was set truthy, the worker only ever received the halt command, and the submitted operations never ran. `Operation.run()` therefore never assigned `self.ret`, and callers hit:

```
AttributeError: 'SyncPackages' object has no attribute 'ret'
```

This breaks the first stage of `gpexpand` ("Syncing Apache Cloudberry extensions"): the exception is squashed into a `WARNING`, and — more importantly — package/extension synchronization to new segment hosts **silently never runs**.

## Fix

Pass the operations as the `items` keyword argument so they are enqueued and `should_stop` stays `False`:

```python
super(OperationWorkerPool, self).__init__(numWorkers, items=operations)
```

This re-applies commit `cd3c88f6e1e` ("Fix gppkg error: 'SyncPackages' object has no attribute 'ret'."), which was inadvertently reverted by the `Merge tag 'REL_16_9' into Cloudberry` merge (`0f4cf8d5068`). On current `main`, `grep -c "items=operations" gpMgmt/bin/gppylib/commands/base.py` returns `0`.

## Testing

- Single-host `gpexpand -i <input> -v` first stage no longer raises the `AttributeError`; the `SyncPackages` operation now actually executes (the `-v` log shows the worker receiving the operation instead of only a halt command).

See #1825 for the full root-cause analysis and the regression history.

### DIFF
--- a/gpMgmt/bin/gppylib/commands/base.py
+++ b/gpMgmt/bin/gppylib/commands/base.py
@@ -230,7 +230,7 @@ class OperationWorkerPool(WorkerPool):
         if operations is not None:
             for operation in operations:
                 self._spoof_operation(operation)
-        super(OperationWorkerPool, self).__init__(numWorkers, operations)
+        super(OperationWorkerPool, self).__init__(numWorkers, items=operations)
 
     def check_results(self):
         raise NotImplementedError("OperationWorkerPool has no means of verifying success.")


leborchuk:
We have lost fix https://github.com/apache/cloudberry/pull/1716/changes while rebasing. Maybe we should just cherry-pick fix from PG14_ARCHIVE to main branch?

tuhaihe:
> We have lost fix https://github.com/apache/cloudberry/pull/1716/changes while rebasing. Maybe we should just cherry-pick fix from PG14_ARCHIVE to main branch?

+1.

xinzweb (approved):
Verified this against current `main` rather than taking the description at face value:

- `WorkerPool.__init__(self, numWorkers=16, should_stop=False, items=None, ...)` — confirmed by reading `gpMgmt/bin/gppylib/commands/base.py` directly; `items` is the 3rd positional param.
- `OperationWorkerPool.__init__` on `main` (line 233) currently calls `super().__init__(numWorkers, operations)` — positional, so `operations` binds to `should_stop` and `items` stays `None`. This exactly reproduces the reported bug: nothing gets enqueued, `should_stop` becomes a truthy list, and the worker only ever receives the halt command.
- The "this was already fixed once and silently reverted" claim checks out: commit `cd3c88f6e1e` (2026-05-02) made this exact same `items=operations` change, and it's absent from `main` after the `Merge tag 'REL_16_9' into Cloudberry` merge (`0f4cf8d5068`, 2026-05-28) — confirmed both commits' SHAs, dates, and the merge's parents.
- The fix (`items=operations` as a keyword arg) is correct against the actual `WorkerPool` signature.

Non-blocking note: there's no unit test covering `OperationWorkerPool` (no `test_base.py` under `gpMgmt/bin/gppylib/test/unit`), and this exact bug has now silently regressed once already via a merge with nothing to catch it. Might be worth a small follow-up test asserting the work queue actually gets populated — not blocking this fix, which is minimal and correct as-is.

CI is green across the board (build/RPM/DEB/rat-check/ic-*). LGTM.

PR #1829, opened 2026-06-24
Remove redundant return statements in UDP interconnect code

MutableFire:
In C, reaching the closing brace `}` of a `void` function implicitly returns control to the caller. Explicit `return;` statements in these positions are unnecessary noise that clutters the code. Removing them aligns the code with common C style guidelines and improves readability.


tuhaihe:
> In C, reaching the closing brace `}` of a `void` function implicitly returns control to the caller. Explicit `return;` statements in these positions are unnecessary noise that clutters the code. Removing them aligns the code with common C style guidelines and improves readability.

Hi @MutableFire welcome to add these description to the commit message body. 

Can take the [.gitmessage](https://github.com/apache/cloudberry/blob/main/.gitmessage) as a commit template, FYI.

MutableFire:
> > In C, reaching the closing brace `}` of a `void` function implicitly returns control to the caller. Explicit `return;` statements in these positions are unnecessary noise that clutters the code. Removing them aligns the code with common C style guidelines and improves readability.
> 
> Hi @MutableFire welcome to add these description to the commit message body.
> 
> Can take the [.gitmessage](https://github.com/apache/cloudberry/blob/main/.gitmessage) as a commit template, FYI.

Done.

contrib/interconnect/udp/ic_udpifc.c
```diff
@@ -8208,7 +8192,6 @@ SendDummyPacket(void)
 		pg_freeaddrinfo_all(hint.ai_family, addrs);
 	if (sockfd != -1)
 		closesocket(sockfd);
-	return;
```
yjhjstz, line 8211: any gcc warning ? I think no need if no warning.

MutableFire: No, there were no gcc warnings.
My main motivation, as described in the PR, is simply code clarity. I believe we should avoid unnecessary lines that don't add logic, as they make the code harder to read.
While I agree we should minimize diffs from upstream for standard PostgreSQL code, this is our own custom code. Therefore, I think it's perfectly fine to refine its style and clean it up.
Could you clarify your concern? Is there a specific reason you prefer keeping these return statements?

yjhjstz: It's generally recommended to keep the final return; when a function already contains multiple early return; statements. This keeps the control flow consistent and makes the code easier to read.
```c++
static void
foo(...)
{
    if (a == NULL)
        return;

    if (b == NULL)
        return;

    ...

    return;
}
```

MutableFire: Done.

yjhjstz (approved):
LGTM

leborchuk (approved):
LGTM

PR #1842, opened 2026-07-06
Feature: Apache Iceberg lake tables as an extension (datalake_fdw skeleton)

MisterRaindrop:
> [!IMPORTANT]
> **This PR has been repurposed, in place, from kernel syntax to an extension.**
>
> Review here converged on doing lake tables without kernel changes, and
> @tuhaihe [asked](https://github.com/apache/cloudberry/pull/1842#issuecomment-5248374467)
> for the new commits to land in this PR so the review context stays in one
> place. So rather than opening a third PR, the branch now carries:
>
> 1. the original 26 kernel commits, **untouched**, so the 64 inline review
>    comments stay anchored to the code they were written about;
> 2. one commit that reverts all of them (the tree afterwards is identical to
>    the merge base, verified by comparing tree object ids);
> 3. the extension itself, moved over from #1881.
>
> Nothing was force-pushed. The kernel work is therefore still recoverable two
> ways: revert the single revert commit, or check out
> `archive/iceberg-ddl-kernel-pr1842` on the author's fork.
>
> **The diff is now 100% different from what was reviewed and approved earlier
> in this thread, so the existing approval no longer applies to it. Please
> re-review from scratch.** The original description is kept, struck through, at
> the bottom.

### What does this PR do?

Adds `contrib/datalake_fdw`: Apache Iceberg lake tables as a table access
method, with no kernel changes. A lake table is an ordinary
`CREATE TABLE ... USING iceberg WITH (catalog = '...', volume = '...')`, where
`catalog` and `volume` name foreign servers created through two foreign-data
wrappers this extension registers.

**This is deliberately a skeleton, published to get the shape reviewed before
the implementation follows.** `CREATE TABLE` and `DROP TABLE` work end to end
against a stub metadata engine -- no catalog service, object store, Arrow or JVM
involved -- and everything that would touch data reports a clean
`iceberg: <operation> is not supported yet`. The interfaces the later work plugs
into ship whole so they can be argued about now: the metadata-engine vtable with
a capability bitmap the registry validates and dispatches through, the
reader/writer format interfaces, and the storage facade.

The question this change is really asking is whether the **seam** is in the
right place -- see Additional Context.

```sql
CREATE SERVER cat FOREIGN DATA WRAPPER iceberg_catalog_fdw
  OPTIONS (type 'rest', uri 'https://catalog.example:443', warehouse 's3://bucket/wh');
CREATE USER MAPPING FOR CURRENT_USER SERVER cat
  OPTIONS (client_id '...', client_secret '...');   -- credentials only ever live here

CREATE SERVER vol FOREIGN DATA WRAPPER iceberg_volume_fdw
  OPTIONS (base_path 's3://bucket/prefix', region 'us-east-1');
CREATE USER MAPPING FOR CURRENT_USER SERVER vol
  OPTIONS (access_key_id '...', secret_access_key '...');

CREATE TABLE t (a int, b text) USING iceberg
  WITH (catalog = 'cat', volume = 'vol', purge_on_drop = false);
```

### Type of Change
- [ ] Documentation update

### Test Plan

### Impact

**Performance:** none. Nothing outside the module changes behaviour, and no
statement gains work: the utility and object-access hooks match on the access
method and return immediately for every other relation.

**User-facing changes:** a new extension, plus two lines of registration
(`contrib/Makefile`, and a CI matrix entry so the suite actually runs). The
extension installs process-wide hooks, so it must be in
`shared_preload_libraries`; `_PG_init` refuses to load any other way rather than
half-initialising. One consequence worth documenting when this ships: because
`gpconfig -c shared_preload_libraries` *replaces* the value rather than
appending to it, overwriting that GUC on a cluster where the extension is in use
will break the databases that have it installed.

**Dependencies:** none. The module builds with no new external dependency -- the
C++ translation units are stubs today, and the object-store SDK and Arrow arrive
with the changes that need them.

### Checklist

### Additional Context

**This is the upstream half of an implementation that already exists.** A
datalake_fdw with the same shape -- the same two hooks, the same table access
method, largely the same file names -- runs in production elsewhere, and the
intent is that this becomes the single source for it rather than a second
lookalike. Two consequences a reviewer should see explicitly:

*The mapping accessor is deliberately shaped to that implementation.* A lake
table's mapping to its catalog and volume servers is read only through
`pg_iceberg_get_table_info()`, whose signature and result types
(`IcebergTableInfo`, `IcebergTableOptions`) match it field for field. That
implementation keeps the same mapping in a system catalog of its own, which an
extension cannot add, so here the function reads reloptions instead -- but the
difference stops inside the function body, and the layers above it can be the
same code on both sides.

*The SQL vocabulary follows the Iceberg specification, not the released
implementation.* Catalog servers take `uri` and `warehouse`, and the REST catalog
type is `rest`. An earlier revision of this change used the other tree's `url` /
`warehouse_location_prefix` / `polaris` on the reasoning that one vocabulary
across both trees was worth more than either name; review pointed out that the
REST catalog specification exists precisely so that these keys are the same
across implementations, and that argument wins. `polaris` is kept as an accepted
alias for `rest` so existing SQL keeps working. Convergence with the other tree
is unaffected in substance: the macro names, struct names, field names and
parse-function split are unchanged, so only the string a user types differs.
`hadoop` and `s3` are in the vocabulary but refused until something implements
them.

### Semantics that are decided (and why they are in the SQL surface)

- **`DROP TABLE` does not delete lake data** unless the table was created with
  `purge_on_drop = true`. It is a table option rather than a GUC or a separate
  function so that it travels with the table, cannot turn one `DROP` into a
  destructive one through session state, and is carried out by a dump.
- **Table identity is the `(catalog, namespace, table)` name triple**, matching
  the existing implementation.
- **Concurrency is Iceberg's**: compare-and-swap on the metadata pointer, with
  the external catalog owning the metadata. Nothing here takes distributed locks.
- **A schema that has diverged from the lake is an error**, not something to
  adapt to silently. The external table is the truth.

### What was and was not verified

Stated plainly, because a skeleton is worth less if you have to guess:

**Verified on a three-segment cluster:** `make installcheck` 3/3, and
order-independent (each file alone and in reverse order, on a fresh database);
zero build warnings; exactly 8 exported symbols; a schema-level `pg_dump`/restore
round trip; Apache RAT clean. Separately, an 84-statement smoke pass over
ordinary objects -- heap/AO/partitioned/temp/unlogged tables, CTAS, matviews,
inherited and typed tables, views, sequences, indexes, functions, another
wrapper's servers and user mappings, every ALTER form, COPY, TRUNCATE, VACUUM
FULL, CLUSTER, REINDEX -- with the library preloaded, both with and without the
extension installed, plus two concurrency probes. Zero errors: the hooks are
inert for everything that is not a lake table.

**Verified by CI** (this was listed as unverified when the change was first
published, and has since run): the `ic-datalake-fdw` entry added here passes on
Rocky 8, 9 and 10, and `ic-cbdb-parallel` passes on every platform.

**Still not verified:** the error-detail channel has no regression coverage, for
the reason given below.

### Why this branch also touches `contrib/interconnect`

The second commit changes two lines in another module's test, which would
otherwise look like unrelated scope. It is here because that test made this PR's
CI red and the failure is worth fixing rather than retrying:
`contrib/interconnect/sql/interconnect.sql` restarts the cluster with
`gpstop -raiq`. An immediate shutdown skips the shutdown checkpoint, so the next
startup runs crash recovery; while the postmaster is in `PM_RECOVERY` it rejects
the connection `gpstart` makes to read the segment configuration ("the database
system is not accepting connections"), `gpstop -r` exits CRITICAL, psql gives up
at the following `\c`, and every statement in the file is skipped -- leaving the
cluster with only the coordinator running, so the suites that run after it in the
same job fail too. A fast shutdown checkpoints on the way out, so no recovery
runs and the state that produces that error is never entered. Measured before
and after with recovery deliberately slowed: `-raiq` failed 2/2, `-rafq` passed
3/3. Fast is also what the rest of the tree already uses; this file was the only
`-raiq`. Happy to split it into its own PR if that is preferred.

### Known limits, stated rather than left to be found

- The mapping stores server *names*, so `ALTER SERVER ... RENAME` on a referenced
  server is refused. Storing OIDs instead would allow it, at the cost of owning
  dump/restore translation; the accessor above is the only thing that would
  change.
- A full `pg_dump` of a database containing a lake table fails, because dumping
  table *contents* means scanning. A schema-level dump round-trips. The data
  belongs to the lake, so it arguably should not be in the dump at all -- but
  `pg_dump` decides from relkind, which an extension cannot influence, so that
  needs a separate conversation.
- Every `ALTER TABLE` form except `OWNER TO` is refused while the access method
  is unfinished. `purge_on_drop` is therefore fixed at CREATE time.
- The error-detail channel has no regression coverage, because no statement can
  make the stub engine fail.

### Naming

`datalake_fdw` is inherited from the existing implementation and is a poor fit:
what this installs is a table access method plus two wrappers, not one FDW. A
better name is welcome while it is still cheap to change.

Discussion: #1683 &nbsp;&nbsp;·&nbsp;&nbsp; Superseded PR: #1881

---

<details>
<summary><b>Original description — kernel-side DDL scaffolding (superseded; kept for the review history)</b></summary>

<br>

**~~Summary~~**

~~Kernel-side DDL scaffolding for Iceberg lake tables, per the design proposal in #1683 (this PR is the "core syntax" milestone of the roadmap posted there). It adds the system catalogs, parse nodes, grammar, commands, and client-tool awareness that a datalake provider extension builds on. **No table AM implementation or provider is included** — those are later milestones (agent / FDW phases in the roadmap).~~

```sql
CREATE FOREIGN CATALOG hive_cat SERVER hive_cat_srv OPTIONS (type 'hive', uri 'thrift://...');
CREATE FOREIGN VOLUME s3_vol SERVER s3_srv OPTIONS (path 's3://bucket/prefix');
CREATE ICEBERG TABLE t (a int, b text) CATALOG hive_cat VOLUME s3_vol OPTIONS (fileformat 'parquet');
DROP CATALOG hive_cat;  DROP VOLUME s3_vol;
```

**~~What's included (structured for commit-by-commit review)~~**

1. ~~**catalog**: `pg_foreign_catalog`, `pg_foreign_volume` (metadata service / storage location handles, hanging off `pg_foreign_server`), `pg_lake_table` (per-relation lake metadata keyed by relid). All three have TOAST tables; name indexes are single-column unique (see "Design decisions").~~
2. ~~**nodes**: `CreateLakeTableStmt` (embeds `CreateStmt`), `CreateForeignCatalogStmt`, `CreateForeignVolumeStmt` + hand-maintained node-support functions.~~
3. ~~**parser**: `CREATE ICEBERG TABLE ... CATALOG ... VOLUME ... OPTIONS (...)`, `CREATE FOREIGN CATALOG|VOLUME ... SERVER ...`, `DROP CATALOG|VOLUME` via `drop_type_name`.~~
4. ~~**commands (foreign catalog/volume)**: create/drop with IF [NOT] EXISTS, dependencies, ownership, object addressing / COMMENT / event-trigger integration, syscaches, QD→QE dispatch with synchronized OIDs.~~
5. ~~**commands (CREATE ICEBERG TABLE)**: resolves catalog/volume (explicit clause or `iceberg_default_catalog`/`iceberg_default_volume` GUCs), validates the statement uses the `iceberg` AM, creates the relation + TOAST + `pg_lake_table` row + dependencies; guards `ALTER TABLE SET ACCESS METHOD` / `SET DISTRIBUTED BY` both directions; lake tables are forced `DISTRIBUTED RANDOMLY`.~~
6. ~~**bin**: `pg_dump` skips iceberg tables (by AM name, with a warning — same pattern as PAX); psql tab completion.~~
7. ~~**tests**: `lake_table` regression in `greenplum_schedule` (DDL lifecycle, duplicate/USING rejection, TOAST wide rows, ownership, dependency/CASCADE, QD/QE dispatch checks) + catalog-expected refresh across `regress`, `singlenode_regress`, and pax copies.~~

**~~Design decisions to review~~**

- ~~**Provider decoupling**: the kernel resolves the AM strictly by name (`get_table_am_oid("iceberg")`); no hardcoded AM OID and no extension-name checks. Everything degrades gracefully (with a hint) when no provider is installed.~~
- ~~**Provider integration contract**: `CreateLakeTable()` runs after `DefineRelation()` and ends with `CommandCounterIncrement()` + `InvokeObjectPostCreateHook(LakeTableRelationId, ...)` — a provider performs remote Iceberg creation from that object-access hook (on QD and QEs), where catalog/volume/options metadata is already visible. `relation_set_new_filelocator` remains local-storage-only.~~
- ~~**Catalog/volume names are database-global** (single-column unique index), like `pg_foreign_server` — every reference syntax (`DROP CATALOG x`, the `CATALOG x` clause, the GUCs) identifies them by bare name, so the uniqueness scope matches what the syntax can express. (The alternative — per-server names like user mappings — would require server-qualified reference syntax everywhere.)~~
- ~~`CREATE ICEBERG TABLE` **rejects a `USING` clause naming any other AM**; without it the statement implies `USING iceberg`.~~

**~~Known limitations / open questions (deliberately out of scope here)~~**

- ~~**pg_dump / pg_upgrade**: FOREIGN CATALOG/VOLUME objects and pg_lake_table metadata are not dumped yet; iceberg tables are skipped with a warning (PAX precedent). Restore/upgrade support is planned as a follow-up.~~
- ~~**Permission model**: creating a catalog/volume requires `USAGE` on its server. Referencing one from `CREATE ICEBERG TABLE` currently checks existence only — whether that should require `USAGE` on the underlying server, or dedicated ACLs (`GRANT USAGE ON CATALOG`), is an open question we'd like reviewer input on.~~
- ~~The GUC defaults are synced to QEs and behave like `default_tablespace` w.r.t. objects dropped after `SET`.~~

~~Discussion: #1683~~

</details>


src/backend/commands/laketablecmds.c
```diff
@@ -0,0 +1,469 @@
...
+ *	  lake table creation/manipulation commands
+ *
+ * Portions Copyright (c) 1996-2021, PostgreSQL Global Development Group
+ *
+ *
+ * IDENTIFICATION
+ *	  src/backend/commands/laketablecmds.c
+ *
+ *-------------------------------------------------------------------------
+ */
```
tuhaihe, line 27: For the new files created by the Cloudberry community, we can use the standard ASF license header:

https://github.com/apache/cloudberry/blob/a4ece882be877e23ae3e1ea3981b39e43788d40e/contrib/pax_storage/src/cpp/pax_gbench.cc#L1-L26

If this PR includes the third-party files, feel free to let me know.

MisterRaindrop: Ok

contrib/pax_storage/src/test/regress/expected/oidjoins.out
```diff
@@ -235,6 +235,10 @@ NOTICE:  checking pg_foreign_data_wrapper {fdwhandler} => pg_proc {oid}
 NOTICE:  checking pg_foreign_data_wrapper {fdwvalidator} => pg_proc {oid}
 NOTICE:  checking pg_foreign_server {srvowner} => pg_authid {oid}
 NOTICE:  checking pg_foreign_server {srvfdw} => pg_foreign_data_wrapper {oid}
+NOTICE:  checking pg_foreign_catalog {fcowner} => pg_authid {oid}
+NOTICE:  checking pg_foreign_catalog {fcserver} => pg_foreign_server {oid}
+NOTICE:  checking pg_foreign_volume {fvowner} => pg_authid {oid}
+NOTICE:  checking pg_foreign_volume {fvserver} => pg_foreign_server {oid}
```
Vlasdislav, line 241: Maybe you will use `SET client_min_messages TO error;` to ignore `NOTICE` in whole file

src/backend/commands/foreigncmds.c
```diff
@@ -1026,6 +1028,266 @@ CreateForeignServer(CreateForeignServerStmt *stmt)
...
+					 errmsg("foreign catalog \"%s\" already exists, skipping",
+							stmt->catalogname)));
+			table_close(rel, RowExclusiveLock);
+			return InvalidObjectAddress;
+		}
+		else
+			ereport(ERROR,
+					(errcode(ERRCODE_DUPLICATE_OBJECT),
+					 errmsg("foreign catalog \"%s\" already exists",
+							stmt->catalogname)));
```
andr-sokolov: I suggest to invert the condition to reduce indentation  and get rid of `else`. Here and in `CreateForeignVolume` too.
```suggestion
		if (!stmt->if_not_exists)
			ereport(ERROR,
					(errcode(ERRCODE_DUPLICATE_OBJECT),
					 errmsg("foreign catalog \"%s\" already exists",
							stmt->catalogname)));

		/*
		 * If we are in an extension script, insist that the pre-existing
		 * object be a member of the extension, to avoid security risks.
		 */
		ObjectAddressSet(myself, ForeignCatalogRelationId, catalogId);
		checkMembershipInCurrentExtension(&myself);

		/* OK to skip */
		ereport(NOTICE,
				(errcode(ERRCODE_DUPLICATE_OBJECT),
				 errmsg("foreign catalog \"%s\" already exists, skipping",
						stmt->catalogname)));
		table_close(rel, RowExclusiveLock);
		return InvalidObjectAddress;
```

src/test/regress/sql/lake_table.sql
```diff
@@ -0,0 +1,142 @@
...
+\d+ pg_foreign_volume
+\d+ pg_lake_table
+
+-- Setup: foreign servers for the catalogs and volumes to hang off
+CREATE FOREIGN DATA WRAPPER lake_test_fdw;
+CREATE SERVER lake_test_srv FOREIGN DATA WRAPPER lake_test_fdw;
+CREATE SERVER lake_test_srv2 FOREIGN DATA WRAPPER lake_test_fdw;
+
+-- CREATE FOREIGN CATALOG
+CREATE FOREIGN CATALOG lake_test_cat SERVER lake_test_srv OPTIONS (type 'hive', uri 'thrift://localhost:9083');
```
andr-sokolov: What about making `type` as a required parameter to make more strict checks? For example
```
CREATE FOREIGN CATALOG lake_test_cat SERVER lake_test_srv TYPE 'hive' URI 'thrift://localhost:9083';
```

When I pass options I can pass nonexistent type and invalid option name, and the query is executed without errors
```
postgres=# CREATE FOREIGN CATALOG lake_test_cat SERVER lake_test_srv OPTIONS (type 'nonexistent_type', uriii 'thrift://localhost:9083');
CREATE FOREIGN CATALOG
postgres=# table pg_foreign_catalog;
  oid  |    fcname     | fcowner | fcserver |                       fcoptions                       
-------+---------------+---------+----------+-------------------------------------------------------
 17022 | lake_test_cat |      10 |    17019 | {type=nonexistent_type,uriii=thrift://localhost:9083}
(1 row)

postgres=# 
```

src/test/regress/sql/lake_table.sql
```diff
@@ -0,0 +1,142 @@
...
+-- DROP CATALOG ... CASCADE takes the remaining tables with it
+DROP CATALOG lake_test_cat CASCADE;
+SELECT count(*) FROM pg_lake_table lt JOIN pg_class c ON c.oid = lt.ltrelid
+ WHERE c.relname LIKE 'lake\_test%';
+
+-- DROP variants
+DROP CATALOG lake_test_cat;		-- fail, already gone
+DROP CATALOG IF EXISTS lake_test_cat;	-- skip with notice
+DROP VOLUME lake_test_vol;
+DROP VOLUME lake_test_vol;		-- fail, already gone
```
andr-sokolov: When I have created foreign table, I cannot drop it using `drop table`, I should use `DROP FOREIGN TABLE`. But in the case of foreign catalog I should drop it using `drop catalog` without the `foreign` keyword. So I suggest to replace `DROP CATALOG` with `DROP FOREIGN CATALOG` and `DROP VOLUME` with `DROP FOREIGN VOLUME` for unification with existing syntax

```
postgres=# create foreign table ttt(i int, j int) server lake_test_srv;
CREATE FOREIGN TABLE
postgres=# drop table ttt;
ERROR:  "ttt" is not a table
HINT:  Use DROP FOREIGN TABLE to remove a foreign table.
postgres=# DROP FOREIGN CATALOG lake_test_cat;
ERROR:  syntax error at or near "CATALOG"
LINE 1: DROP FOREIGN CATALOG lake_test_cat;
                     ^
postgres=# 
```

contrib/pax_storage/src/test/regress/expected/oidjoins.out
andr-sokolov: This is not related to Iceberg DDL. We shouldn't change client_min_messages in this PR.

src/test/regress/sql/lake_table.sql
MisterRaindrop: I fully agree. However, I haven’t implemented validation at this stage because the syntax is still incomplete and subject to frequent changes. Given this, would you recommend we add validation now, or hold off until the entire workflow is fully functional before enforcing validation rules?

andr-sokolov: I recommend to make syntax without the OPTIONS keyword, if possible, because in this case validation will be implemented in grammar. So the validation will be an integral part of the syntax implementation.

MisterRaindrop: That might be difficult. We need to support various external catalogs with numerous distinct parameter configurations. Therefore, using SQL options would be the optimal approach.

andr-sokolov: May be. But I think every catalog has its type. So type is not an option, but a required property, which should be in `pg_foreign_catalog` as a column and in grammar

src/test/regress/sql/lake_table.sql
andr-sokolov: I suggest to add `DROP ICEBERG TABLE` as a pair for `CREATE ICEBERG TABLE` too.
```
postgres=# CREATE ICEBERG TABLE lake_test_t0 (a int) CATALOG lake_test_cat VOLUME lake_test_vol;        
CREATE LAKE TABLE
postgres=# DROP ICEBERG TABLE lake_test_t0;
ERROR:  syntax error at or near "ICEBERG"
LINE 1: DROP ICEBERG TABLE lake_test_t0;
             ^
postgres=# 
```

src/backend/commands/laketablecmds.c
```diff
@@ -0,0 +1,483 @@
...
+ *
+ * Look up the OID of the iceberg table access method, which is provided by
+ * a datalake extension rather than the kernel.  Returns InvalidOid if the
+ * access method is not installed and missing_ok is true.
+ */
+Oid
+GetIcebergTableAmOid(bool missing_ok)
+{
+	return get_table_am_oid(ICEBERG_TABLE_AM_NAME, missing_ok);
+}
```
andr-sokolov, line 195: In the b4704be776f814b46a5a59eb543df80fad32a864 commit description I read " The iceberg table access method is resolved strictly by name (get_table_am_oid) and is expected to be provided by a datalake extension; without it, CREATE ICEBERG TABLE fails up front with a hint. The kernel does not hardcode any extension name or AM OID."

In my opinion, it is strange to have queries (like `CREATE ICEBERG TABLE`) in the core which are useless without extensions. What about default iceberg AM which can be replaced (using `USING`) with AM from some extension? Without default AM we cannot test the code properly.

src/backend/commands/laketablecmds.c
```diff
@@ -0,0 +1,483 @@
...
+
+/*
+ * RelationIsIcebergTable
+ *
+ * True iff the relation uses the iceberg table access method.  Resolved by
+ * access method name so the kernel does not depend on any particular
+ * extension's OID assignments.
+ */
+bool
+RelationIsIcebergTable(Relation rel)
```
andr-sokolov: What about adding `RELKIND_ICEBERG` constant like `RELKIND_FOREIGN_TABLE` and checking `rel->rd_rel->relkind` here to make it possible to use more that one extension with AM?

src/test/regress/sql/lake_table.sql
MisterRaindrop: The currently supported types are hive, hdfs, polaris, s3 and builtin. 
```
CREATE FOREIGN CATALOG lake_test_cat SERVER lake_test_srv TYPE 'builtin';
CREATE FOREIGN CATALOG lake_test_cat SERVER lake_test_srv TYPE 's3';
CREATE FOREIGN CATALOG lake_test_cat SERVER lake_test_srv TYPE 'hive';
CREATE FOREIGN CATALOG lake_test_cat SERVER lake_test_srv TYPE 'polaris';
CREATE FOREIGN CATALOG lake_test_cat SERVER lake_test_srv TYPE 'hdfs';
```
Is it correct like this?

src/backend/commands/laketablecmds.c
MisterRaindrop: 1. Does this AM refer to the minimal Iceberg AM implementation that does not support scans or inserts? Is that correct?
2. One more question: CREATE ICEBERG TABLE USING <am>
Is this syntax designed to support extensions via other <am> frameworks? From my understanding, there should be no need to specify USING am if we're already using CREATE ICEBERG TABLE, right?

src/test/regress/sql/lake_table.sql
andr-sokolov: Yes, it is

andr-sokolov: What about type name without quotes around?
```
CREATE FOREIGN CATALOG lake_test_cat SERVER lake_test_srv TYPE builtin;
```

src/backend/commands/laketablecmds.c
MisterRaindrop: Adding RELKIND_ICEBERG requires modifications across all reference points, which will be a substantial change. Do we plan to support multiple AMs in the future? From my understanding, CREATE ICEBERG statements natively only create Iceberg-related tables. you mean like hudi, Paimon ?，Could you share your thoughts on this?

MisterRaindrop: If that’s the case, a syntax like CREATE LAKEHOUSE xxx USING xxx would be more appropriate.

src/test/regress/sql/lake_table.sql
MisterRaindrop: I don't recommend this approach, as it would result in too many dedicated catalog creation syntaxes. We may also support JDBC catalog and other user-defined catalog types down the line. Creating a separate syntax for each type would clearly be redundant.

andr-sokolov: Ok

src/backend/commands/laketablecmds.c
andr-sokolov: If we can add AM using an extension, then we can have more than one such extensions. User may want to install several extensions. I'm not sure if several iceberg extensions installation is really useful. But your idea about possible support of other formats (`CREATE LAKEHOUSE xxx USING xxx`) seems to be very interesting

src/backend/catalog/heap.c
```diff
@@ -2322,6 +2323,10 @@ heap_drop_with_catalog(Oid relid)
 	 */
 	CheckTableForSerializableConflictIn(rel);
 
+	/* If this is a lake table, remove its pg_lake_table entry */
+	if (RelationIsIcebergTable(rel))
+		RemoveLakeTableEntry(relid);
```
yjhjstz, line 2328: can RemoveLakeTableEntry put into extension via object_access_hook ?

MisterRaindrop: I can migrate this logic to object_access_hook, but the issue is that the AM layer hasn’t been implemented yet on my end, which blocks the migration.
So do I need to implement a default plugin for the AM layer first? Then replace this default AM plugin later once the official datalake plugin is fully rolled out?

src/backend/commands/laketablecmds.c
MisterRaindrop: At this stage, I believe the regression testing coverage is sufficient, test stub is enough. Introducing the AM layer at this point would only add unnecessary complexity with no practical benefit, as we still cannot run any real data workloads with it.

src/test/regress/sql/lake_table.sql
MisterRaindrop: I added a type category to the grammar.

src/test/regress/sql/lake_table.sql
MisterRaindrop: Create and delete now share the same syntax.

src/backend/commands/laketablecmds.c
MisterRaindrop: At present, I do not expect multiple Iceberg AM plugins to exist, hence adding RELKIND_ICEBERG is unnecessary.

src/backend/foreign/foreign.c
```diff
@@ -982,6 +984,98 @@ get_foreign_server_oid(const char *servername, bool missing_ok)
...
+
+	/* Extract the volume options */
+	datum = SysCacheGetAttr(FOREIGNVOLUMENAME,
+							tp,
+							Anum_pg_foreign_volume_fvoptions,
+							&isnull);
+	if (isnull)
+		volume->options = NIL;
+	else
+		volume->options = untransformRelOptions(datum);
```
andr-sokolov: I suggest using ternary operator to make the code more compact and to explicitly write that for any value of the condition, we change one variable
```suggestion
	volume->options = (isnull) ? NIL : untransformRelOptions(datum);
```

src/test/regress/sql/lake_table.sql
```diff
@@ -0,0 +1,150 @@
...
+CREATE FOREIGN CATALOG lake_test_cat SERVER lake_test_srv TYPE 'hive';			-- fail, duplicate
+CREATE FOREIGN CATALOG IF NOT EXISTS lake_test_cat SERVER lake_test_srv TYPE 'hive';	-- skip with notice
+-- catalog names are global: the same name on another server is still a duplicate
+CREATE FOREIGN CATALOG lake_test_cat SERVER lake_test_srv2 TYPE 'hive';			-- fail, duplicate
+CREATE FOREIGN CATALOG IF NOT EXISTS lake_test_cat SERVER lake_test_srv2 TYPE 'hive';	-- skip with notice
+CREATE FOREIGN CATALOG lake_test_bad SERVER no_such_server TYPE 'hive';		-- fail, no server
+SELECT fcname, fctype, fcoptions FROM pg_foreign_catalog WHERE fcname LIKE 'lake\_test%' ORDER BY 1;
+
+-- CREATE FOREIGN VOLUME
+CREATE FOREIGN VOLUME lake_test_vol SERVER lake_test_srv OPTIONS (path 's3://bucket/prefix');
```
andr-sokolov: The proposal says that volume is "Where data files physically live". Is it possible that a volume can be without path? What about moving path from options to  required first-class properties?

src/bin/psql/tab-complete.c
```diff
@@ -3819,7 +3854,15 @@ psql_completion(const char *text, int start, int end)
...
-		COMPLETE_WITH("DATA WRAPPER", "TABLE");
+		COMPLETE_WITH("CATALOG", "DATA WRAPPER", "TABLE", "VOLUME");
+	else if (Matches("DROP", "FOREIGN", "CATALOG"))
+		COMPLETE_WITH_QUERY(Query_for_list_of_foreign_catalogs);
+	else if (Matches("DROP", "FOREIGN", "VOLUME"))
+		COMPLETE_WITH_QUERY(Query_for_list_of_foreign_volumes);
+	else if (Matches("DROP", "ICEBERG"))
+		COMPLETE_WITH("TABLE");
+	else if (Matches("DROP", "ICEBERG", "TABLE"))
+		COMPLETE_WITH_SCHEMA_QUERY(Query_for_list_of_tables);
```
andr-sokolov: Is it possible to write a query to get here iceberg tables only?

src/include/commands/laketablecmds.h
```diff
@@ -0,0 +1,61 @@
...
+extern bool check_iceberg_default_volume(char **newval, void **extra, GucSource source);
+
+/* Functions to get default values */
+extern const char *GetDefaultIcebergCatalog(void);
+extern const char *GetDefaultIcebergVolume(void);
+
+/* Lake table management */
+extern Oid	GetIcebergTableAmOid(bool missing_ok);
+extern bool RelationIsIcebergTable(Relation rel);
+extern void ValidateLakeTableOptions(CreateLakeTableStmt *stmt);
```
andr-sokolov: I think the `Options` word in the function name is not correct because `CreateLakeTableStmt` contains the `options` field and reading the function name I get a false expectation that the function validates the `options` field only. I suggest to rename the function. For example, `ValidateLakeTableStmt`, because it validates all fields of the structure
```suggestion
extern void ValidateLakeTableStmt(CreateLakeTableStmt *stmt);
```

src/test/regress/sql/lake_table.sql
MisterRaindrop: During design, this path functions as a prefix and is optional for users to specify. If a bucket is available, you may fully specify the path within the Iceberg table configuration. like name is base_path in Discussion: https://github.com/apache/cloudberry/discussions/1683

src/backend/catalog/heap.c
yjhjstz: if you can provide a draft version, that's good.

andr-sokolov:
When a foreign table has been created and I press tab after `drop table`, there is no foreign table names in the suggestion. I cannot use `drop table` to remove a foreign table.
```
postgres=# CREATE FOREIGN TABLE ft (i int) SERVER lake_test_srv;
CREATE FOREIGN TABLE
postgres=# drop table 
gp_toolkit.          information_schema.  public.              t                    t2                   
postgres=# drop table ft;
ERROR:  "ft" is not a table
HINT:  Use DROP FOREIGN TABLE to remove a foreign table.
postgres=# 
```

Iceberg table name is displayed  in the suggestion after `drop table`. I can use `drop table` to remove an iceberg table.
```
postgres=# CREATE ICEBERG TABLE lake_test_t1 (a int, b text) CATALOG lake_test_cat VOLUME lake_test_vol OPTIONS (fileformat 'parquet');
CREATE LAKE TABLE
postgres=# drop table 
gp_toolkit.          information_schema.  lake_test_t1         public.              t                    t2                   
postgres=# drop table lake_test_t1;
DROP TABLE
postgres=# 
```

Is it correct? I expect that iceberg table is something like foreign table and should be processed using a similar syntax

andr-sokolov:
What about adding new syntax in psql help (`\h`)?

MisterRaindrop:
> When a foreign table has been created and I press tab after `drop table`, there is no foreign table names in the suggestion. I cannot use `drop table` to remove a foreign table.
> 
> ```
> postgres=# CREATE FOREIGN TABLE ft (i int) SERVER lake_test_srv;
> CREATE FOREIGN TABLE
> postgres=# drop table 
> gp_toolkit.          information_schema.  public.              t                    t2                   
> postgres=# drop table ft;
> ERROR:  "ft" is not a table
> HINT:  Use DROP FOREIGN TABLE to remove a foreign table.
> postgres=# 
> ```
> 
> Iceberg table name is displayed in the suggestion after `drop table`. I can use `drop table` to remove an iceberg table.
> 
> ```
> postgres=# CREATE ICEBERG TABLE lake_test_t1 (a int, b text) CATALOG lake_test_cat VOLUME lake_test_vol OPTIONS (fileformat 'parquet');
> CREATE LAKE TABLE
> postgres=# drop table 
> gp_toolkit.          information_schema.  lake_test_t1         public.              t                    t2                   
> postgres=# drop table lake_test_t1;
> DROP TABLE
> postgres=# 
> ```
> 
> Is it correct? I expect that iceberg table is something like foreign table and should be processed using a similar syntax

It's been updated. Tables are now dropped using the DROP ICEBERG TABLE syntax.

src/backend/catalog/heap.c
MisterRaindrop: I’m still unsure how best to provide this. Could you share some existing examples for reference?

src/test/regress/sql/lake_table.sql
```diff
@@ -0,0 +1,155 @@
...
+-- CREATE FOREIGN CATALOG: TYPE is a required first-class property
+CREATE FOREIGN CATALOG lake_test_cat SERVER lake_test_srv TYPE 'hive' OPTIONS (uri 'thrift://localhost:9083');
+CREATE FOREIGN CATALOG lake_test_notype SERVER lake_test_srv;		-- fail, TYPE is required
+CREATE FOREIGN CATALOG lake_test_cat SERVER lake_test_srv TYPE 'hive';			-- fail, duplicate
+CREATE FOREIGN CATALOG IF NOT EXISTS lake_test_cat SERVER lake_test_srv TYPE 'hive';	-- skip with notice
+-- catalog names are global: the same name on another server is still a duplicate
+CREATE FOREIGN CATALOG lake_test_cat SERVER lake_test_srv2 TYPE 'hive';			-- fail, duplicate
+CREATE FOREIGN CATALOG IF NOT EXISTS lake_test_cat SERVER lake_test_srv2 TYPE 'hive';	-- skip with notice
+CREATE FOREIGN CATALOG lake_test_bad SERVER no_such_server TYPE 'hive';		-- fail, no server
+SELECT fcname, fctype, fcoptions FROM pg_foreign_catalog WHERE fcname LIKE 'lake\_test%' ORDER BY 1;
```
andr-sokolov: Why do we need `ORDER BY 1` here and below? Tuples order doesn't matter for regression tests

doc/src/sgml/ref/create_iceberg_table.sgml
```diff
@@ -0,0 +1,193 @@
...
+
+ <refsynopsisdiv>
+<synopsis>
+CREATE ICEBERG TABLE [ IF NOT EXISTS ] <replaceable class="parameter">table_name</replaceable> ( [
+    <replaceable class="parameter">column_name</replaceable> <replaceable class="parameter">data_type</replaceable> [, ... ]
+  ] )
+    [ CATALOG <replaceable class="parameter">catalog_name</replaceable> ]
+    [ VOLUME <replaceable class="parameter">volume_name</replaceable> ]
+    [ OPTIONS ( <replaceable class="parameter">option</replaceable> '<replaceable class="parameter">value</replaceable>' [, ...] ) ]
+    [ DISTRIBUTED BY ( <replaceable class="parameter">column</replaceable> [, ... ] ) | DISTRIBUTED RANDOMLY | DISTRIBUTED REPLICATED ]
```
andr-sokolov: It is written in the test comment: "Lake tables are always DISTRIBUTED RANDOMLY". Can we remove this line from syntax? This line seems meaningless.

src/include/tcop/cmdtaglist.h
```diff
@@ -176,10 +179,13 @@ PG_CMDTAG(CMDTAG_DROP_DOMAIN, "DROP DOMAIN", true, false, false)
 PG_CMDTAG(CMDTAG_DROP_DYNAMIC_TABLE, "DROP DYNAMIC TABLE", true, false, false)
 PG_CMDTAG(CMDTAG_DROP_EVENT_TRIGGER, "DROP EVENT TRIGGER", false, false, false)
 PG_CMDTAG(CMDTAG_DROP_EXTENSION, "DROP EXTENSION", true, false, false)
+PG_CMDTAG(CMDTAG_DROP_FOREIGN_CATALOG, "DROP FOREIGN CATALOG", true, false, false)
 PG_CMDTAG(CMDTAG_DROP_FOREIGN_DATA_WRAPPER, "DROP FOREIGN DATA WRAPPER", true, false, false)
 PG_CMDTAG(CMDTAG_DROP_FOREIGN_TABLE, "DROP FOREIGN TABLE", true, false, false)
+PG_CMDTAG(CMDTAG_DROP_FOREIGN_VOLUME, "DROP FOREIGN VOLUME", true, false, false)
 PG_CMDTAG(CMDTAG_DROP_FUNCTION, "DROP FUNCTION", true, false, false)
 PG_CMDTAG(CMDTAG_DROP_INDEX, "DROP INDEX", true, false, false)
+PG_CMDTAG(CMDTAG_DROP_LAKE_TABLE, "DROP LAKE TABLE", true, false, false)
```
andr-sokolov, line 188: When I run a `CREATE ICEBERG TABLE` query, I get output `CREATE LAKE TABLE`. The same replacement `ICEBERG` with `LAKE` happens in the case of `DROP ICEBERG TABLE`.
```
postgres=# CREATE ICEBERG TABLE lake_test_t1 (a int, b text) CATALOG lake_test_cat VOLUME lake_test_vol OPTIONS (fileformat 'parquet');
CREATE LAKE TABLE
postgres=# DROP ICEBERG TABLE IF EXISTS lake_test_t1;
DROP LAKE TABLE
postgres=# 
```
Usually the output is the beginning of CREATE or DROP query. May be it means that we should replace `CREATE ICEBERG TABLE` with `CREATE LAKE TABLE ... TYPE 'ICEBERG'` or `CREATE LAKE TABLE ... USING ICEBERG`? We have the laketablecmds.c and laketablecmds.h files, the `T_CreateLakeTableStmt` node with the `table_type` field and the `pg_lake_table` catalog table with `lttable_type` column. I think it will be cool if the syntax will correspond to data structures and names in the source code.

MisterRaindrop: This is a good question. I suggest other reviewers join the discussion: which approach is better, `CREATE ICEBERG`, or `CREATE LAKE TABLE ... USING ICEBERG`?

`CREATE LAKE TABLE ... USING ICEBERG`  This offers the benefit of decoupling the data lake from the kernel. The AM plugin can be controlled via registration.

@andr-sokolov @yjhjstz @leborchuk @my-ship-it What do you think?

leborchuk: Hi! I vote for `CREATE LAKE TABLE ... USING ICEBERG` - add space for the other formats )

andr-sokolov: @MisterRaindrop I vote for `CREATE LAKE TABLE ... USING ICEBERG` too

MisterRaindrop:
@andr-sokolov It is ready for review

src/test/regress/expected/lake_table.out
```diff
@@ -0,0 +1,318 @@
...
+
+\d+ pg_lake_table
+                                  Table "pg_catalog.pg_lake_table"
+      Column       |  Type  | Collation | Nullable | Default | Storage  | Stats target | Description 
+-------------------+--------+-----------+----------+---------+----------+--------------+-------------
+ ltrelid           | oid    |           | not null |         | plain    |              | 
+ ltforeign_catalog | oid    |           | not null |         | plain    |              | 
+ ltforeign_volume  | oid    |           | not null |         | plain    |              | 
+ lttable_type      | text   | C         |          |         | extended |              | 
+ ltoptions         | text[] | C         |          |         | extended |              |
```
andr-sokolov: Can we use `pg_class.relam` instead of `lttable_type` and `pg_class.reloptions` instead of `ltoptions`?

When a lake table is created, `lttable_type` contains the same string as `amname` does but in upper case, and the `reloptions` column is unused.
```
postgres=# CREATE LAKE TABLE lake_test_t1 (a int, b text) USING ICEBERG CATALOG lake_test_cat VOLUME lake_test_vol OPTIONS (fileformat 'parquet');
CREATE LAKE TABLE
postgres=# select lttable_type, ltoptions from pg_lake_table;
 lttable_type |      ltoptions       
--------------+----------------------
 ICEBERG      | {fileformat=parquet}
(1 row)

postgres=# select amname, reloptions from pg_class, pg_am where relname='lake_test_t1' and relam=pg_am.oid;
 amname  | reloptions 
---------+------------
 iceberg | 
(1 row)

postgres=# 

```

src/test/regress/sql/lake_table.sql
```diff
@@ -0,0 +1,154 @@
...
+SELECT policytype, distkey FROM gp_distribution_policy WHERE localoid = 'lake_test_t1'::regclass;
+
+-- The (heap-backed) table is usable
+INSERT INTO lake_test_t1 VALUES (1, 'x'), (2, 'y');
+SELECT count(*) FROM lake_test_t1;
+
+-- Lake tables get a TOAST table like plain tables, so wide values work
+SELECT reltoastrelid <> 0 AS has_toast FROM pg_class WHERE relname = 'lake_test_t1';
+INSERT INTO lake_test_t1 VALUES (3, repeat('x', 500000));
+SELECT a, length(b) FROM lake_test_t1 WHERE a = 3;
```
andr-sokolov: TOAST table is needed to save data in it, but lake table doesn't save its data in database files. Do we really need this part of the test?

src/test/regress/sql/lake_table.sql
```diff
@@ -0,0 +1,154 @@
...
+
+-- ... unless the GUCs provide defaults
+SET iceberg_default_catalog = 'lake_test_cat';
+SET iceberg_default_volume = 'lake_test_vol';
+CREATE LAKE TABLE lake_test_t2 (a int) USING ICEBERG;
+RESET iceberg_default_catalog;
+RESET iceberg_default_volume;
+
+-- A DISTRIBUTED clause is ignored with a warning
+CREATE LAKE TABLE lake_test_t3 (a int) USING ICEBERG CATALOG lake_test_cat VOLUME lake_test_vol DISTRIBUTED BY (a);
```
andr-sokolov: I suggest to show an error message instead of warning, because `DISTRIBUTED BY` is not supported for the `CREATE LAKE TABLE` query
```
postgres=# \h CREATE LAKE TABLE
Command:     CREATE LAKE TABLE
Description: define a new lake table
Syntax:
CREATE LAKE TABLE [ IF NOT EXISTS ] table_name ( [
    column_name data_type [, ... ]
  ] )
    USING format
    [ CATALOG catalog_name ]
    [ VOLUME volume_name ]
    [ OPTIONS ( option 'value' [, ...] ) ]

URL: https://www.postgresql.org/docs/16/sql-createlaketable.html

postgres=# 

```

src/backend/parser/gram.y
```diff
@@ -9093,6 +9098,185 @@ CreateDirectoryTableStmt:
...
+ * the table format (currently only ICEBERG); the format also determines the
+ * table access method the relation is created with.
+ *
+ *****************************************************************************/
+
+CreateLakeTableStmt:
+			CREATE LAKE TABLE qualified_name '(' OptTableElementList ')'
+			USING name
+			OptForeignCatalog OptForeignVolume create_generic_options
+			OptDistributedBy
```
andr-sokolov: I suggest to remove this line because  DISTRIBUTED clause is not supported for lake tables

src/backend/commands/laketablecmds.c
```diff
@@ -0,0 +1,448 @@
...
+ */
+static void
+validate_table_type(const char *table_type)
+{
+	if (!table_type)
+		ereport(ERROR,
+				(errcode(ERRCODE_INVALID_PARAMETER_VALUE),
+				 errmsg("lake table format cannot be NULL")));
+
+	if (strcmp(table_type, "ICEBERG") != 0)
```
andr-sokolov: What about using ICEBERG_TABLE_AM_NAME here?
```suggestion
	if (strcmp(table_type, ICEBERG_TABLE_AM_NAME) != 0)
```

src/backend/commands/laketablecmds.c
```diff
@@ -0,0 +1,448 @@
...
+	if (!OidIsValid(GetIcebergTableAmOid(true)))
+		ereport(ERROR,
+				(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+				 errmsg("table access method \"%s\" does not exist",
+						ICEBERG_TABLE_AM_NAME),
+				 errhint("CREATE LAKE TABLE ... USING ICEBERG requires an extension that provides the \"%s\" table access method.",
+						 ICEBERG_TABLE_AM_NAME)));
+
+	/* Validate the table format named in the USING clause */
+	validate_table_type(stmt->table_type);
```
andr-sokolov: Let's validate table type before checking for AM existence to make it possible to pass any table type in the future
```suggestion
	/* Validate the table format named in the USING clause */
	validate_table_type(stmt->table_type);

	/*
	 * Lake tables are unusable without an extension providing
	 * a table access method; check it first so the install hint takes
	 * precedence over catalog/volume resolution errors.
	 */
	if (!OidIsValid(get_table_am_oid(stmt->table_type, true)))
		ereport(ERROR,
				(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
				 errmsg("table access method \"%s\" does not exist",
						stmt->table_type),
				 errhint("CREATE LAKE TABLE ... USING %s requires an extension that provides the \"%s\" table access method.",
						 stmt->table_type, stmt->table_type)));
```

src/backend/commands/laketablecmds.c
```diff
@@ -0,0 +1,448 @@
...
+ * access method is not installed and missing_ok is true.
+ */
+Oid
+GetIcebergTableAmOid(bool missing_ok)
+{
+	return get_table_am_oid(ICEBERG_TABLE_AM_NAME, missing_ok);
+}
+
+/*
+ * RelationIsIcebergTable
```
andr-sokolov: I suggest to rename this function to `RelationIsLakeTable` and check for the table oid existence in `pg_lake_table`. It seems that we call this function to check whether relation is lake table, but not iceberg table only.

src/backend/commands/laketablecmds.c
```diff
@@ -0,0 +1,458 @@
...
+	 * check it here (after the format) so the install hint takes precedence
+	 * over catalog/volume resolution errors.
+	 */
+	if (!OidIsValid(get_table_am_oid(stmt->table_type, true)))
+		ereport(ERROR,
+				(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+				 errmsg("table access method \"%s\" does not exist",
+						stmt->table_type),
+				 errhint("CREATE LAKE TABLE ... USING ICEBERG requires an extension that provides the \"%s\" table access method.",
+						 stmt->table_type)));
```
andr-sokolov: There is ICEBERG word in the hint. I suggest to replace it with stmt->table_type because table_type may be not iceberg only in the future
```suggestion
				 errhint("CREATE LAKE TABLE ... USING \"%s\" requires an extension that provides the \"%s\" table access method.",
						 stmt->table_type, stmt->table_type)));
```

src/test/regress/sql/lake_table.sql
```diff
@@ -0,0 +1,157 @@
...
+CREATE LAKE TABLE lake_test_optbad (a int) USING ICEBERG CATALOG lake_test_cat VOLUME lake_test_vol OPTIONS (bogus_opt 'x');	-- fail, AM rejects unknown option
+DROP LAKE TABLE lake_test_opt;
+
+-- Catalog and volume are both required
+CREATE LAKE TABLE lake_test_t2 (a int) USING ICEBERG VOLUME lake_test_vol;	-- fail, no catalog
+CREATE LAKE TABLE lake_test_t2 (a int) USING ICEBERG CATALOG lake_test_cat;	-- fail, no volume
+
+-- ... unless the GUCs provide defaults
+SET iceberg_default_catalog = 'lake_test_cat';
+SET iceberg_default_volume = 'lake_test_vol';
```
andr-sokolov, line 88: What about renaming GUCs
```suggestion
SET lake_default_catalog = 'lake_test_cat';
SET lake_default_volume = 'lake_test_vol';
```
and adding a new GUC with default lake table AM? The new GUC with default value "iceberg" will be useful to don't write `USING ICEBERG` in every `CREATE LAKE TABLE` query

andr-sokolov: Catalogs and volumes can potentially be specified for any format. But the low-level implementations for various formats are distinct, so we should specify format when catalogs and volumes are created.
For example
```
CREATE FOREIGN CATALOG ... FORMAT 'iceberg' TYPE 'hive';
CREATE FOREIGN VOLUME ...  FORMAT 'iceberg' SERVER ...;
```

When FORMAT is not specified we can get it from the new GUC

MisterRaindrop: @andr-sokolov 
1. about renaming GUCs
The current design supports multiple formats. If we rename iceberg_default_volume to a generic GUC parameter like lake_default_volume, we will not be able to determine whether a lake table uses Iceberg or another format. In my opinion, there should be one default configuration per format, such as iceberg_default_catalog and hudi_default_catalog. This eliminates ambiguity, at the cost of introducing additional GUC variables.
  
2. about CREATE FOREIGN CATALOG ... FORMAT 'xxx' 
```
CREATE FOREIGN CATALOG ... FORMAT 'iceberg' TYPE 'hive';
CREATE FOREIGN VOLUME ... FORMAT 'iceberg' SERVER ...;
```
This binds catalogs and volumes to the Iceberg format. I am unsure whether catalogs and volumes will be reused across multiple formats in the future. Nevertheless, such binding lacks flexibility. Besides, we already have corresponding GUC parameters, making the explicit FORMAT clause redundant. When creating lake tables, users will clearly specify the catalog and volume to use. Therefore, I suggest not binding catalogs or volumes to specific formats.

src/test/regress/sql/lake_table.sql
```diff
@@ -0,0 +1,157 @@
...
+RESET ROLE;
+
+-- Dependencies: the server holds the catalog/volume, which hold the tables
+DROP SERVER lake_test_srv;		-- fail, catalog and volume depend on it
+DROP FOREIGN CATALOG lake_test_cat;	-- fail, tables depend on it
+
+-- Dropping a lake table removes its pg_lake_table entry
+DROP LAKE TABLE lake_test_t1;
+SELECT count(*) FROM pg_lake_table lt JOIN pg_class c ON c.oid = lt.ltrelid
+ WHERE c.relname = 'lake_test_t1';
```
andr-sokolov: This query returns 0 when a lake table entry is deleted from pg_class only. It doesn't check that `DROP LAKE TABLE` deletes entry from pg_lake_table. We should save the table oid before `DROP LAKE TABLE` and find the oid in pg_lake_table after the query.

src/backend/commands/tablecmds.c
```diff
@@ -17059,13 +17110,32 @@ static void
...
+		ereport(ERROR,
+				(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+				 errmsg("cannot change access method of table \"%s\" to \"%s\"",
+						RelationGetRelationName(rel), ICEBERG_TABLE_AM_NAME),
+				 errhint("Use CREATE LAKE TABLE ... USING ICEBERG to create a lake table.")));
+	if (OidIsValid(iceberg_amoid) && rel->rd_rel->relam == iceberg_amoid)
+		ereport(ERROR,
+				(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+				 errmsg("cannot change access method of lake table \"%s\"",
+						RelationGetRelationName(rel))));
```
andr-sokolov: Let's check `OidIsValid(iceberg_amoid)` one time.
ICEBERG_TABLE_AM_NAME is a constant, but not variable, we can concatenate it at compile time
```suggestion
	if (OidIsValid(iceberg_amoid) {
		if (amoid == iceberg_amoid)
			ereport(ERROR,
					(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
					 errmsg("cannot change access method of table \"%s\" to \"" ICEBERG_TABLE_AM_NAME "\"",
							RelationGetRelationName(rel)),
					 errhint("Use CREATE LAKE TABLE ... USING ICEBERG to create a lake table.")));
		if (rel->rd_rel->relam == iceberg_amoid)
			ereport(ERROR,
					(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
					 errmsg("cannot change access method of lake table \"%s\"",
							RelationGetRelationName(rel))));
	}
```

andr-sokolov (approved):

src/test/regress/sql/lake_table.sql
```diff
@@ -0,0 +1,158 @@
...
+  FROM pg_foreign_volume WHERE fvname = 'lake_test_vol';
+
+-- Catalog and volume rows are dispatched to all segments
+SELECT count(DISTINCT gp_segment_id) > 1 AS on_all_segments
+  FROM gp_dist_random('pg_foreign_catalog') WHERE fcname = 'lake_test_cat';
+SELECT count(DISTINCT gp_segment_id) > 1 AS on_all_segments
+  FROM gp_dist_random('pg_foreign_volume') WHERE fvname = 'lake_test_vol';
+
+-- Without a provider extension there is no iceberg table AM
+CREATE LAKE TABLE lake_test_t0 (a int) USING ICEBERG CATALOG lake_test_cat VOLUME lake_test_vol;	-- fail with hint
```
yjhjstz, line 49: We can align syntax with crunchydata.com

[Building a Postgres Data Warehouse using Iceberg.pdf](https://github.com/user-attachments/files/30297764/Building.a.Postgres.Data.Warehouse.using.Iceberg.pdf)

```sql
> Iceberg as a Postgres table format Capture queries, writes, & schema changes to provide a transactional table experience for Iceberg in S3.

create table chats (
message_id bigserial not null,
thread_id bigint not null,
…
) using iceberg;
```
@leborchuk  @andr-sokolov

andr-sokolov: @yjhjstz How does crunchydata specify catalog? I didn't find this syntax in their presentation

leborchuk: > Building a Postgres Data Warehouse using Iceberg.pdf

Yes, looks good. But how we setup catalog and lakehouse format? There should exist some definition ... It's good for pre-configured database instances, but your own database needs a more advanced syntax.

src/backend/catalog/heap.c
yjhjstz: just add a fixme .

src/test/regress/sql/lake_table.sql
yjhjstz: Em,  currently how about change `CREATE LAKE TABLE lake_test_t0 ` -> `CREATE TABLE lake_test_t0`, no need `LAKE` ?

MisterRaindrop: @andr-sokolov @leborchuk @yjhjstz I looked into how Crunchy Data Warehouse (now open-sourced by Snowflake as [pg_lake](https://github.com/Snowflake-Labs/pg_lake)) handles the catalog — hope this helps answer the question above.

**TL;DR: Crunchy doesn't specify a catalog at CREATE time, because Postgres itself acts as the catalog.**

- Managed Iceberg tables are created with plain `CREATE TABLE ... USING iceberg`. The Iceberg metadata lives inside Postgres and is exposed through an `iceberg_tables` view compatible with the Iceberg JDBC/SQL catalog protocol, so Spark / pyiceberg / iceberg-rust connect **to Postgres as the catalog** (JDBC with Postgres credentials; `catalog_name` = the database name). ([Crunchy docs](https://docs.crunchybridge.com/warehouse/iceberg))
- I couldn't find any syntax for attaching an external catalog (Glue / HMS / REST) — the product doesn't seem to have that concept, which would explain why it doesn't appear in the presentation.
- Storage location is a GUC default plus an optional per-table reloption: `SET crunchy_iceberg.default_location_prefix TO 's3://...'`, or `CREATE TABLE ... USING iceberg WITH (location = 's3://...')`. With neither set, the cluster's bundled managed storage is used — so in the demo nothing needs to be specified at all. @leborchuk's reading ("good for pre-configured database instances") matches what I found.

One more data point that may be relevant: pg_lake is currently adding external Iceberg REST catalog support as a **client** ([pg_lake#94](https://github.com/Snowflake-Labs/pg_lake/issues/94)), and that roadmap mentions handling multiple endpoints/credentials "likely using a Postgres `SERVER`-like concept" — quite close to what `CREATE FOREIGN CATALOG ... SERVER` does in this PR. So the two designs may be converging from different starting points.

For reference, the two syntax shapes side by side:

- Plain `CREATE TABLE ... USING iceberg`: the catalog/volume bindings move into reloptions, e.g. `WITH (catalog 'c', volume 'v')`, with GUC defaults; no new grammar, and plain `DROP TABLE` applies (the Spark / pg_lake convention).
- `CREATE LAKE TABLE ... USING <format> CATALOG c VOLUME v`: catalog/volume are first-class clauses with dedicated `CREATE/DROP LAKE TABLE` commands (the current PR).

yjhjstz: if we use `CREATE TABLE ... USING iceberg WITH (catalog 'c', volume 'v')`, we need not change kernel code, iceberg will be totally an extension .

MisterRaindrop: @yjhjstz Agreed — this looks technically feasible, so I sketched what the fully-extension shape could look like.

Since Cloudberry's table AM API already supports per-AM reloptions validation (`TableAmRoutine->amoptions`, unlike vanilla PG16), `WITH (catalog ..., volume ...)` can be validated by the iceberg AM itself with no kernel changes. The catalog/volume objects could reuse the existing FDW infrastructure (`SERVER` + `USER MAPPING`), which gives ACLs, pg_dump support, QD/QE dispatch and dependency tracking for free:

```sql
CREATE EXTENSION pg_iceberg;   -- provides the table AM, two FDWs, and hooks

CREATE SERVER hive_cat FOREIGN DATA WRAPPER iceberg_catalog
    OPTIONS (type 'hive', uri 'thrift://metastore:9083');
CREATE USER MAPPING FOR alice SERVER hive_cat
    OPTIONS (user '...', password '...');

CREATE SERVER s3_vol FOREIGN DATA WRAPPER iceberg_volume
    OPTIONS (base_path 's3://bucket/prefix');

CREATE TABLE t (a int, b text) USING iceberg
    WITH (catalog 'hive_cat', volume 's3_vol');

DROP TABLE t;   -- plain DROP TABLE applies
```

A few things the extension would need to own, instead of kernel guards:

- registering table metadata and `pg_depend` entries (table → server) via `object_access_hook`, so `DROP SERVER` stays protected and CASCADE works;
- intercepting paths that don't make sense for iceberg (CTAS, partition children, `ALTER TABLE ... SET ACCESS METHOD`) via `ProcessUtility_hook`, and forcing random distribution;
- the library would need to be in `shared_preload_libraries` so the hooks are always installed.

This is essentially the pg_lake architecture, and `USER MAPPING` would also give a cleaner credentials story than embedding keys in options. If we go this way, the kernel scaffolding in this PR (the grammar, the three catalogs, the guards) would no longer be needed — the work would move into the datalake extension instead. WDYT?
@leborchuk @andr-sokolov

andr-sokolov: @MisterRaindrop If possible, let's not change the kernel to reduce conflicts when updating to new versions of PostgreSQL. I think it's better to use an extension.

MisterRaindrop:
@andr-sokolov @yjhjstz 
I open this draft PR about lake table ext (create table xxx using iceberg xxx), please give some suggest and review it.
https://github.com/apache/cloudberry/pull/1881

tuhaihe:
> @andr-sokolov @yjhjstz I open this draft PR about lake table ext (create table xxx using iceberg xxx), please give some suggest and review it. #1881

Hi, is there any way to introduce your new commits to this place, where we have a context for review? FYI. @MisterRaindrop

MisterRaindrop:
@tuhaihe as you asked, the extension work now lives in this PR instead of a separate one, so the review context stays in one place. @andr-sokolov @yjhjstz please note what changed below before looking at the diff again.

**How it was moved, and why this way.** The branch was *not* force-pushed. It now has, in order:

1. the original 26 kernel commits, untouched -- so your 64 inline comments stay anchored to the code they were written about;
2. `Revert the kernel-side lake-table DDL scaffolding` -- one commit that undoes all of them. The tree after it is identical to the merge base, verified by comparing tree object ids rather than by reading a diff;
3. the two extension commits, moved over from #1881. The resulting diff is identical to what #1881 showed -- I compared the two patches line by line; the only differences are blob hashes and one hunk offset, because the two branches had different merge bases.

Rewriting the branch would have been shorter, but it would have orphaned the review history, and it would have made going back expensive. This way the kernel scaffolding is one `git revert` away if the extension approach turns out to be wrong. The pre-revert tip is also kept as `archive/iceberg-ddl-kernel-pr1842` on my fork.

**Please re-review from scratch.** The diff is now 100% different from what was reviewed and approved earlier in this thread, so the earlier approval does not apply to what is here now. I would rather say that explicitly than have it merged on the strength of it.

**One thing that will look out of place:** the second commit changes two lines in `contrib/interconnect/sql/interconnect.sql`. That test restarts the cluster with `gpstop -raiq`; an immediate shutdown skips the shutdown checkpoint, so the next startup runs crash recovery, and while the postmaster is in `PM_RECOVERY` it rejects the connection `gpstart` makes to read the segment configuration. `gpstop -r` then exits CRITICAL, psql gives up at the `\c` that follows, the whole file is skipped, and the cluster is left with only the coordinator running -- which is why `pg_trgm` and `indexscan` failed after it in the same job. A fast shutdown checkpoints on the way out, so that state is never entered. Measured with recovery deliberately slowed: `-raiq` failed 2/2, `-rafq` passed 3/3. Fast is what the rest of the tree already uses; this file was the only `-raiq`. Happy to split it into its own PR if you would rather keep this one to the extension.

The original description is kept at the bottom of the PR body, struck through. #1881 is superseded by this PR.

contrib/Makefile
```diff
@@ -25,6 +25,7 @@ SUBDIRS = \
 		btree_gin	\
 		btree_gist	\
 		citext		\
+		datalake_fdw	\
```
tuhaihe: Is `datalake_fdw` enabled by default? We can disable it by default and enable it using the `--enable-datalake_fdw` configure option, the way like PAX.

MisterRaindrop: Good call -- done in 6f40dd0f7db, following the PAX mechanism: `PGAC_ARG_BOOL` defaulting to `no`, substituted into `Makefile.global`, and a conditional in `contrib/Makefile` that moves the directory to `ALWAYS_SUBDIRS` when disabled so the clean targets still reach it.

Three things I decided while doing it, all cheap to reverse if you disagree:

- **The option is spelled `--enable-datalake-fdw`, with dashes.** Every option in this tree uses dashes (`--enable-tap-tests`, `--enable-ic-udp2`, `--enable-catalog-ext`), and autoconf maps the dashes to `enable_datalake_fdw` for the Makefile variable, so the directory name still matches. Happy to use the underscore form if you would rather it read exactly like the directory.
- **No `AC_DEFINE`.** PAX and ic-udp2 define a macro because C code tests it; nothing here does, and adding an unread macro would mean touching `src/include/pg_config.h.in` for no reader. `--enable-pxf` and `--enable-orafce` already use this shorter four-argument form.
- **`--enable-datalake-fdw` is also added to `devops/build/automation/cloudberry/scripts/configure-cloudberry.sh`**, next to `--enable-pax`. Without it the `ic-datalake-fdw` CI job would install nothing and silently test nothing. I deliberately did *not* add it to `coverity.yml` and `sonarqube.yml`, which also pass `--enable-pax` -- say the word if you want the module in the weekly scans too.

**One thing you should know about `configure`.** I regenerated it by hand rather than running autoconf over the whole file, because the committed `configure` and `configure.ac` are currently out of sync in *both* directions:

- `configure.ac` has a restructured PAX liburing check (the `case $host_os in linux*)` block with the macOS fallback comment) that was never regenerated into `configure`;
- `configure` has a Darwin python shared-library lookup (trying `.dylib` in addition to `DLSUFFIX`) that does not exist in `configure.ac`.

A full `autoconf` run therefore produced twelve hunks, only four of which were mine; taking all of them would have swept those two unrelated changes into this PR and, in the python case, reverted working behaviour. So I applied only the four hunks for this option and left the drift alone. It is worth someone reconciling those two separately -- I did not want to do it inside this PR.

Verified by configuring a real tree both ways, not by inspection:

| | configure says | `Makefile.global` | `SUBDIRS` | `ALWAYS_SUBDIRS` |
|---|---|---|---|---|
| `--enable-datalake-fdw` | `... yes` | `enable_datalake_fdw = yes` | has `datalake_fdw` | -- |
| default | `... no` | `enable_datalake_fdw = no` | -- | has `datalake_fdw` |

andr-sokolov:
What about moving the extension to gpcontrib to minimize conflicts with PostgreSQL?

contrib/interconnect/sql/interconnect.sql
```diff
@@ -83,5 +89,5 @@ DROP EXTENSION interconnect;
 
 -- start_ignore
 \! gpconfig -r shared_preload_libraries
-\! gpstop -raiq
+\! gpstop -rafq
```
andr-sokolov: Why are the changes in another extension tests relate to datalake_fdw?

contrib/datalake_fdw/src/iceberg_volume_fdw/iceberg_volume_fdw.c
```diff
@@ -0,0 +1,157 @@
...
+}
+
+Datum
+iceberg_volume_fdw_validator(PG_FUNCTION_ARGS)
+{
+	List	   *options = untransformRelOptions(PG_GETARG_DATUM(0));
+	Oid			catalog = PG_GETARG_OID(1);
+	ListCell   *lc;
+	IcebergVolumeServerOptions server_options;
+	IcebergForeignVolumeOptions volume_options;
```
andr-sokolov: What about using `{0}` instead of `memset`? The code will be simpler and some potential errors, which compiler cannot detect, will be avoided
```suggestion
	IcebergVolumeServerOptions server_options = {0};
	IcebergForeignVolumeOptions volume_options = {0};
```

tuhaihe:
> What about moving the extension to gpcontrib to minimize conflicts with PostgreSQL?

I think it's OK to place it in the `contrib` folder, just like `pax` which was created by the Cloudberry community.

leborchuk (approved):
LGTM

contrib/interconnect/sql/interconnect.sql
MisterRaindrop: Fair objection — split out into **#1919** and reverted here in a720cf69798, so this PR no longer touches `contrib/interconnect` at all.

It was here because that race is what made this PR's `ic-contrib` job fail, and the failure is destructive rather than local: `gpstop -raiq` skips the shutdown checkpoint, the next startup runs crash recovery, and while the postmaster is in `PM_RECOVERY` it rejects the connection `gpstart` makes to read the segment configuration. `gpstop -r` then exits CRITICAL, `psql` gives up at the following `\c`, the whole file is skipped, and the cluster is left with only the coordinator running — which is why `pg_trgm` and `indexscan` also failed after it in the same job. #1919 has the full chain and the before/after measurements.

Reverted rather than dropped by rewriting the branch, so this thread stays anchored to the change it is about.

**One consequence to expect:** `ic-contrib` here may go red again until #1919 lands. That failure is in `contrib/interconnect` and has nothing to do with this module — the library is not in `shared_preload_libraries` for that job, so it is never loaded there.

andr-sokolov (approved):

.github/workflows/build-cloudberry.yml
```diff
@@ -326,6 +326,10 @@ jobs:
                                "contrib/pg_buffercache:installcheck",
                                "contrib/sslinfo:installcheck"]
               },
+              {"test":"ic-datalake-fdw",
+               "make_configs":["contrib/datalake_fdw:installcheck"],
+               "shared_preload_libraries":"datalake_fdw"
```
yjhjstz, line 332: test datalake_fdw as a true extension on a separated CI ?

tuhaihe: I think it looks good now. Same with the PAX way.

tuhaihe (approved):
+1 from ASF Compliance and Brand.

.github/workflows/build-cloudberry.yml
MisterRaindrop: Done in a58aec79f14 — datalake_fdw is now built and tested the way anyone outside the project would consume it, rather than compiled into the server we ship.

- `configure-cloudberry.sh` no longer passes `--enable-datalake-fdw`, so the RPM every test job installs is unchanged by this module. `configure-cloudberry.sh` has therefore dropped out of this PR's diff entirely.
- The test job gained a step that, for a matrix entry naming a `pgxs_extension`, builds and installs it with PGXS against the server that was just installed. It has to run before the demo cluster is created: a module named in `shared_preload_libraries` must exist by the time the cluster starts.
- `--enable-datalake-fdw` itself stays, for people building from source who do want it in-tree.

I validated it on a real run before bringing it here (fork, `workflow_dispatch`, `ic-datalake-fdw` only):

- the build job's log shows `checking whether to build with datalake_fdw support ... no`, and zero compilation of the module — the RPM genuinely does not contain it;
- the new step compiled and installed it in 2.5s (`datalake_fdw.so`, `.control`, `--1.0.sql`);
- the cluster started with it preloaded and the regression passed 3/3.

One thing to weigh, since it is a real cost rather than a detail: **the in-tree path now has no CI coverage at all.** The `contrib/Makefile` conditional and the configure option feeding it are no longer built by anything automated — building from source by hand is the only thing that exercises them. That was previously verified precisely because `ic-datalake-fdw` was green on the RPM build.

@tuhaihe you had said the previous shape looked good and matched PAX, which it did — every optional module here (`pax`, `gpcloud`, `orafce`, `mapreduce`, `ic-proxy`) is `no` by default in `configure.ac` and switched on in `configure-cloudberry.sh`. This change deliberately steps away from that pattern in favour of exercising the extension path instead. Happy to go back to the PAX shape if you'd rather keep the in-tree wiring covered — the previous commit is still in this branch's history.

tuhaihe: Makes sense. Since the extension is still in the skeleton stage, proceeding with the separated CI / PGXS extension path works well for now. Once the implementation matures, we can revisit and switch it back to the `configure.ac` option if needed to align with other modules.

doc/src/sgml/ref/create_iceberg_table.sgml
MisterRaindrop: You were right, and the point outlived the design it was made against.

That file is gone — the kernel DDL it documented was reverted when this PR became an extension, so there is no `CREATE ICEBERG TABLE` grammar and no `DISTRIBUTED` clause in the syntax to remove. `doc/` no longer appears in this PR's diff at all.

The underlying judgement was adopted rather than dropped. A lake table's rows live outside PostgreSQL, so no local key can describe where they are; the module injects `DISTRIBUTED RANDOMLY` itself and does not ask the user for it. `DISTRIBUTED BY` and `DISTRIBUTED REPLICATED` are both refused.

There is one deliberate exception, and it is the only reason any clause is accepted: `pg_dump` writes `DISTRIBUTED RANDOMLY` into the `CREATE TABLE` it emits, so refusing every clause would mean refusing to restore a dump this module produced. An explicit clause is therefore accepted only when it asks for exactly what would have been injected anyway (`pg_iceberg_extensible.c`, in `CREATE TABLE` handling).

Resolving this one, since the file it points at no longer exists.

PR #1846, opened 2026-07-10
ORCA: restore non-ASCII column aliases for all target entry kinds

yjhjstz:
ORCA represents names as wide characters; when the database LC_CTYPE cannot decode a multibyte name (e.g. LC_CTYPE='C' with a UTF-8 alias), clib::Vswprintf substitutes the generic "UNKNOWN" string and the DXL-to-PlStmt translator restores the original name from the query tree (commit 6b19c44262b). That restore only ran for Var target entries, so with the optimizer enabled

    SELECT '한글' AS "한글";

returned a column named UNKNOWN: constants, aggregates, set operation and set-returning-function outputs never took the repair path.

The old repair walked the entire query tree with
update_unknown_locale_walker matching on (resorigtbl, resno). That match key is ambiguous: TargetEntries nested inside Aggref arguments or SubLink subqueries can collide with it, either clobbering an already-restored name (empty column header for the first of two aggregates) or restoring a name from the wrong query level.

Replace the walker with restore_unknown_locale_resname, which scans only the top-level query targetList for the non-junk entry with the same resno. Only the topmost plan node is translated with a context that carries the original query, and its projection list produces the query output columns in order, so the positional top-level match is exact. A legitimate alias literally named "UNKNOWN" self-matches and the restore is a no-op. Call it for every entry translated by TranslateDXLProjList, for the inlined Append targetlist (UNION ALL), and for the ProjectSet targetlists built by SetupAliasParameter (set-returning functions).

Extend gp_locale with alias cases for constants, expressions, aggregates, UNION ALL, set-returning functions, a subquery containing a same-position column, and a legitimate "UNKNOWN" alias.

Fixes #ISSUE_Number

### Type of Change
- [ ] Documentation update

### Test Plan

### Impact

**Performance:**

**User-facing changes:**

**Dependencies:**

### Checklist

### CI Skip Instructions

---

### DIFF
--- a/src/backend/gpopt/translate/CTranslatorDXLToPlStmt.cpp
+++ b/src/backend/gpopt/translate/CTranslatorDXLToPlStmt.cpp
@@ -4031,6 +4031,49 @@ CTranslatorDXLToPlStmt::CreateProjectSetNodeTree(const CDXLNode *result_dxlnode,
 	return project_set_parent_plan;
 }
 
+//---------------------------------------------------------------------------
+//	@function:
+//		restore_unknown_locale_resname
+//
+//	@doc:
+//		ORCA represents strings using wide characters. Converting a multibyte
+//		name to wide format uses vswprintf(), which depends on the database
+//		LC_CTYPE. When that locale cannot interpret the name (e.g. LC_CTYPE=C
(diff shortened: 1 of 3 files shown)


my-ship-it (approved):

zhangwenchao-123 (approved):
LGTM

PR #1848, opened 2026-07-13
Fix gpexpand to automatically set data directory permissions

MutableFire:
Injects logic into gpexpand to automatically correct segment data directory permissions during cluster expansion.


leborchuk (approved):
LGTM

gpMgmt/bin/gpexpand
```diff
@@ -846,6 +846,24 @@ class SegmentTemplate:
         # and mirror information
         update_pg_hba_on_segments(self.gparray, self.isHbaHostnames, self.batch_size, expanded_host_content)
 
+        new_segments = self.gparray.getExpansionSegDbList()
+        chmod_commands_by_host = defaultdict(list)
+
+        for seg in new_segments:
+            if seg.isSegmentMirror():
```
Smyatkin-Maxim, line 853: I'm curious why mirrors are handled differently here? (not saying they shouldn't, just want to make sure)

MutableFire: Mirrors are excluded because they are not started or manipulated during this phase of gpexpand. They are initialized as passive copies and are only brought up later during the gprecoverseg synchronization step. The permission fix here is targeted at the active primaries that we are about to start, clean, and stop. The mirror permissions will be corrected when they are first synchronized

Smyatkin-Maxim (approved):

PR #1852, opened 2026-07-15
Fix ORCA direct dispatch for multi-value quals

### LINKED ISSUE #1839: [Bug] ORCA misses direct dispatch for multi-value predicates (IN / OR of equalities) on the distribution key
### Apache Cloudberry version

All versions

### What happened

  For a predicate that restricts the distribution key to a small set of
  constants, e.g. `WHERE dist_key IN (c1, c2)`, GPORCA only performs direct
  dispatch when ALL values happen to hash to the SAME segment. If the values
  hash to different segments, ORCA gives up entirely and dispatches the slice
  to every segment, while the Postgres planner correctly direct-dispatches to
  the union of the target segments.

  Results are correct — this is a performance issue (missed direct dispatch),
  not a wrong-results bug. I'm filing it as a bug rather than a feature request
  because the optimizer side already emits the complete multi-value dispatch
  info into the DXL plan, and the DXL-to-PlannedStmt translator silently drops
  it (details in "Anything else").

  ORCA should dispatch to the union of the segments the constants hash to,
  like the Postgres planner does (see `MergeDirectDispatchCalculationInfo` in
  `src/backend/cdb/cdbmutate.c` / `directdispatch.c` lineage). For the
  reproduction below, the planner produces:

      Gather Motion 2:1  (slice1; segments: 2)
      INFO:  (slice 1) Dispatch command to PARTIAL contents: 1 2

  while ORCA produces:

      INFO:  (slice 1) Dispatch command to ALL contents: 0 1 2

  The infrastructure already supports this: `PlanSlice->directDispatch.contentIds`
  is a List, and the dispatcher handles multiple content ids today (the planner
  populates it with more than one; so does ORCA's own raw-values path for
  `gp_segment_id IN (...)` predicates on randomly distributed tables).

### How to reproduce

  On a 3-segment demo cluster:

  ```sql
  create table t_dd (a int, b text) distributed by (a);
  insert into t_dd select i, 'x' from generate_series(1, 10) i;
  -- on my cluster: a=1 -> seg1, a=2,3 -> seg0, a=5 -> seg2
  -- (cdbhash of int is deterministic, so a 3-segment cluster gets the same mapping)

  set gp_test_print_direct_dispatch_info = on;

  set optimizer = on;
  explain (costs off) select * from t_dd where a in (1, 5);
  --  Gather Motion 3:1  (slice1; segments: 3)      <-- all segments
  select * from t_dd where a in (1, 5);
  --  INFO:  (slice 1) Dispatch command to ALL contents: 0 1 2

  explain (costs off) select * from t_dd where a in (2, 3);
  --  Gather Motion 1:1  (slice1; segments: 1)      <-- works, but only because
  --                                                    both values hash to seg0
  set optimizer = off;
  explain (costs off) select * from t_dd where a in (1, 5);
  --  Gather Motion 2:1  (slice1; segments: 2)      <-- planner dispatches to
  select * from t_dd where a in (1, 5);             --    the union {seg1, seg2}
  --  INFO:  (slice 1) Dispatch command to PARTIAL contents: 1 2
  ```
The explicit OR form (`a = 1 or a = 5`) behaves the same way.

### Operating System

Rocky Linux 9.5

### Anything else

_No response_

### Code of Conduct

zhangwenchao-123:
For a predicate restricting the distribution key to several constants, e.g. WHERE dist_key IN (c1, c2), the Postgres planner direct-dispatches the slice to the union of the segments the constants hash to. GPORCA gave up instead: the optimizer side already emits one datum array per value into CDXLDirectDispatchInfo, but TranslateDXLDirectDispatchInfo required every value to hash to the same segment and returned NIL otherwise, so the query was dispatched to all segments.

Collect the distinct content ids of all datum combinations instead, mirroring the planner's MergeDirectDispatchCalculationInfo behavior. A value set that spans every segment still returns NIL, since direct dispatch buys nothing there. PlanSlice->directDispatch.contentIds is already a list and the dispatcher handles multiple content ids, so no dispatcher change is needed. Add a gpdb::LAppendUniqueInt wrapper for list_append_unique_int.

On a 3-segment cluster, "select * from t1_varchar where col1_varchar in ('a','b')" now dispatches to 2 segments (PARTIAL contents) instead of all 3, matching the planner. Single-value and same-segment cases are unchanged, as is the raw-values gp_segment_id path. Update the affected expected outputs: the direct_dispatch disjunction scenarios, the bfv_dd dispatch INFO lines, and the plans in create_index, inherit, partition, partition_append, partition_prune, partition_prune_append and qp_misc_jiras whose Gather Motion drops from 3:1 to 2:1, in src/test/regress as well as in the copies of these tests carried by the pax_storage and vectorization suites.

Fixes #1839

### Type of Change
- [ ] Documentation update

### Test Plan

### Impact

**Performance:**

**User-facing changes:**

**Dependencies:**

### Checklist

### CI Skip Instructions

---

### DIFF
--- a/src/backend/gpopt/translate/CTranslatorDXLToPlStmt.cpp
+++ b/src/backend/gpopt/translate/CTranslatorDXLToPlStmt.cpp
@@ -5530,9 +5530,6 @@ CTranslatorDXLToPlStmt::TranslateDXLDirectDispatchInfo(
 		return NIL;
 	}
 
-	CDXLDatumArray *dxl_datum_array = (*dispatch_identifier_datum_arrays)[0];
-	GPOS_ASSERT(0 < dxl_datum_array->Size());
-
 	const ULONG length = dispatch_identifier_datum_arrays->Size();
 
 	if (dxl_direct_dispatch_info->FContainsRawValues())
@@ -5568,23 +5565,29 @@ CTranslatorDXLToPlStmt::TranslateDXLDirectDispatchInfo(
 		return segids_list;
 	}
 
-	ULONG hash_code = GetDXLDatumGPDBHash(dxl_datum_array, pRTEHashFuncCal);
+	// Each datum array is one value combination of the distribution key
+	// (e.g. one array per IN-list constant). Dispatch to the union of the
+	// segments they hash to, like the planner does.
+	List *segids_list = NIL;
 	for (ULONG ul = 0; ul < length; ul++)
 	{
 		CDXLDatumArray *dispatch_identifier_datum_array =
 			(*dispatch_identifier_datum_arrays)[ul];
 		GPOS_ASSERT(0 < dispatch_identifier_datum_array->Size());
-		ULONG hash_code_new = GetDXLDatumGPDBHash(
-			dispatch_identifier_datum_array, pRTEHashFuncCal);
+		ULONG hash_code = GetDXLDatumGPDBHash(dispatch_identifier_datum_array,
+											  pRTEHashFuncCal);
 
-		if (hash_code != hash_code_new)
-		{
-			// values don't hash to the same segment
-			return NIL;
-		}
+		segids_list = gpdb::LAppendUniqueInt(segids_list, hash_code);
 	}
 
-	List *segids_list = gpdb::LAppendInt(NIL, hash_code);
+	// a value set that spans every segment is equivalent to not direct
+	// dispatching at all
+	if (gpdb::ListLength(segids_list) == m_num_of_segments)
+	{
+		gpdb::ListFree(segids_list);
+		return NIL;
+	}
+
 	return segids_list;
 }
 
--- a/contrib/pax_storage/src/test/regress/expected/bfv_dd_optimizer.out
+++ b/contrib/pax_storage/src/test/regress/expected/bfv_dd_optimizer.out
@@ -88,7 +88,7 @@ INFO:  (slice 1) Dispatch command to SINGLE content
 
 -- single column distr key
 select * from dd_singlecol_1 where a in (10,11,12);
-INFO:  (slice 1) Dispatch command to ALL contents: 0 1 2
+INFO:  (slice 1) Dispatch command to PARTIAL contents: 2 1
  a  | b  
 ----+----
  10 | 10
@@ -97,7 +97,7 @@ INFO:  (slice 1) Dispatch command to ALL contents: 0 1 2
 (3 rows)
 
 select * from dd_singlecol_1 where a=10 or a=11 or a=12;
-INFO:  (slice 1) Dispatch command to ALL contents: 0 1 2
+INFO:  (slice 1) Dispatch command to PARTIAL contents: 2 1
  a  | b  
 ----+----
  10 | 10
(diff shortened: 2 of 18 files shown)


my-ship-it (approved):
LGTM, thanks

leborchuk:
It looks like we see some issues in tests, see:

```
diff -I HINT: -I CONTEXT: -I GP_IGNORE: -U3 /__w/cloudberry/cloudberry/src/test/regress/expected/inherit_optimizer.out /__w/cloudberry/cloudberry/src/test/regress/results/inherit.out
--- /__w/cloudberry/cloudberry/src/test/regress/expected/inherit_optimizer.out	2026-07-15 01:31:24.515754942 -0700
+++ /__w/cloudberry/cloudberry/src/test/regress/results/inherit.out	2026-07-15 01:31:24.599755165 -0700
@@ -2500,7 +2500,7 @@
 explain (costs off) select * from mclparted where a in(3,4,5) order by a;
 QUERY PLAN
 ___________
- Gather Motion 3:1  (slice1; segments: 3)
+ Gather Motion 2:1  (slice1; segments: 2)
    Merge Key: a
    ->  Sort
          Sort Key: a
```

zhangwenchao-123:
> It looks like we see some issues in tests, see:
> 
> ```
> diff -I HINT: -I CONTEXT: -I GP_IGNORE: -U3 /__w/cloudberry/cloudberry/src/test/regress/expected/inherit_optimizer.out /__w/cloudberry/cloudberry/src/test/regress/results/inherit.out
> --- /__w/cloudberry/cloudberry/src/test/regress/expected/inherit_optimizer.out	2026-07-15 01:31:24.515754942 -0700
> +++ /__w/cloudberry/cloudberry/src/test/regress/results/inherit.out	2026-07-15 01:31:24.599755165 -0700
> @@ -2500,7 +2500,7 @@
>  explain (costs off) select * from mclparted where a in(3,4,5) order by a;
>  QUERY PLAN
>  ___________
> - Gather Motion 3:1  (slice1; segments: 3)
> + Gather Motion 2:1  (slice1; segments: 2)
>     Merge Key: a
>     ->  Sort
>           Sort Key: a
> ```

Got it, will fix.

src/backend/gpopt/translate/CTranslatorDXLToPlStmt.cpp
```diff
@@ -5568,23 +5565,29 @@ CTranslatorDXLToPlStmt::TranslateDXLDirectDispatchInfo(
...
-			dispatch_identifier_datum_array, pRTEHashFuncCal);
+		ULONG hash_code = GetDXLDatumGPDBHash(dispatch_identifier_datum_array,
+											  pRTEHashFuncCal);
 
-		if (hash_code != hash_code_new)
-		{
-			// values don't hash to the same segment
-			return NIL;
-		}
+		segids_list = gpdb::LAppendUniqueInt(segids_list, hash_code);
```
yjhjstz, line 5580: if (gpdb::ListLength(segids_list) == m_num_of_segments) , early exit ?

zhangwenchao-123: Maybe now is okay, as if we move `if (gpdb::ListLength(segids_list) == m_num_of_segments)` here, we will also add overhead in every iteration of the foreach iteration, as a result, in some cases moving the judgement here are better while are worse in other cases.

yjhjstz (approved):
LGTM

PR #1853, opened 2026-07-15
ORCA: Fix use-after-free in CXformSplitWindowFunc

zhangwenchao-123:
CXformSplitWindowFunc::Transform() builds a local and a global CLogicalSequenceProject, and each constructor takes ownership of one reference to the window's distribution spec (pds), order specs (pdrgpos) and frames (pdrgpwf). Only a single AddRef() was issued for each object, so once both SequenceProjects and the original are released the reference count underflows and the objects are freed while still in use.

The dangling memory later surfaces as a corrupt (unaligned) scalar DXL node while translating a table scan filter, crashing the coordinator with SIGSEGV at CTranslatorDXLToScalar.cpp:111 during DXL-to-PlStmt translation. This reproduces on window-function queries (e.g. TPC-DS query 44) when optimizer_force_split_window_function is on; the transform is only enabled under that GUC, which is why the crash disappears when it is off.

Add the missing AddRef() for pds, pdrgpos and pdrgpwf before building the global SequenceProject, and for the Select operator and its scalar comparison before reusing them in the global Select, so every owner holds its own reference.

Fixes #ISSUE_Number

### Type of Change
- [ ] Documentation update

### Test Plan

### Impact

**Performance:**

**User-facing changes:**

**Dependencies:**

### Checklist

### CI Skip Instructions

---

### DIFF
--- a/src/backend/gporca/libgpopt/src/xforms/CXformSplitWindowFunc.cpp
+++ b/src/backend/gporca/libgpopt/src/xforms/CXformSplitWindowFunc.cpp
@@ -375,12 +375,25 @@ CXformSplitWindowFunc::Transform(CXformContext *pxfctxt, CXformResult *pxfres,
 	CExpression *pexprLocalSelect =
 		GPOS_NEW(mp) CExpression(mp, pSelectCopy, pexprLocal, pexprScalarCmp);
 
+	// the global SequenceProject below takes ownership of one reference each to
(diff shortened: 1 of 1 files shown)


my-ship-it (approved):
LGTM, thanks

leborchuk (approved):
LGTM

yjhjstz (approved):

PR #1863, opened 2026-07-26
Fix inverted cpuset assignment

### LINKED ISSUE #1862: [Bug] Incorrect cpuset role assignment in getCpuSetByRole()
### Apache Cloudberry version

2.1.0-incubating

### What happened

Location: `src/backend/commands/resgroupcmds.c`
Function `getCpuSetByRole()`
Lines: 1618–1649.

The `getCpuSetByRole()` function in `src/backend/commands/resgroupcmds.c` was returning cpuset values for the wrong roles.

```c
 * Seperate cpuset by coordinator and segment
 * Return as splitcpuset
 *
 * ex:
 * cpuset = "1;4"
 * then we should assign '1' to corrdinator and '4' to segment
 * 
 * cpuset = "1"
 * assign '1' to both coordinator and segment
 */
extern char *
getCpuSetByRole(const char *cpuset)
{
	char *splitcpuset = NULL;

	if (cpuset == NULL)
	{
		ereport(ERROR,
				(errcode(ERRCODE_SYNTAX_ERROR),
				errmsg("Unexpected cpuset invalid in getCpuSetByRole")));
	}

	char *first = strchr(cpuset, ';');
	if (first == NULL)
		splitcpuset = (char *)cpuset;
	else
	{
		char *scpu = first + 1;

		/* Get result cpuset by IS_QUERY_DISPATCHER(), on master or segment */
		if (IS_QUERY_DISPATCHER())
			splitcpuset = scpu;
		else
		{
			char *mcpu = (char *)palloc0(sizeof(char) * MaxCpuSetLength);
			strncpy(mcpu, cpuset, first - cpuset);
			splitcpuset = mcpu;
		}
	}

	return splitcpuset;
}
```

This bug manifests in heterogeneous Cloudberry clusters where the coordinator and segment nodes have different numbers of CPU cores.
The issue becomes visible only when the core counts differ; on homogeneous clusters the swapped values would be identical and the bug would go unnoticed.

When the function receives a cpuset string formatted as "coordinator_cpus;segment_cpus" (e.g., "0-7;0-15" where coordinator has 8 cores and segments have 16 cores), it should return:

- coordinator cpuset (first part) when called from the coordinator (dispatcher),
- segment cpuset (second part) when called from segments.

However, the logic was inverted – it returned the opposite value for each role, causing incorrect CPU affinity assignment.

Currently, the assignments are inverted. The fix is to swap the logic so that the coordinator gets `mcpu `and the segment gets `scpu`:

```c
 * Seperate cpuset by coordinator and segment
 * Return as splitcpuset
 *
 * ex:
 * cpuset = "1;4"
 * then we should assign '1' to corrdinator and '4' to segment
 * 
 * cpuset = "1"
 * assign '1' to both coordinator and segment
 */
extern char *
getCpuSetByRole(const char *cpuset)
{
	char *splitcpuset = NULL;

	if (cpuset == NULL)
	{
		ereport(ERROR,
				(errcode(ERRCODE_SYNTAX_ERROR),
				errmsg("Unexpected cpuset invalid in getCpuSetByRole")));
	}

	char *first = strchr(cpuset, ';');
	if (first == NULL)
		splitcpuset = (char *)cpuset;
	else
	{
		char *scpu = first + 1;

		/* Get result cpuset by IS_QUERY_DISPATCHER(), on master or segment */
		if (IS_QUERY_DISPATCHER())
		{
			char *mcpu = (char *)palloc0(sizeof(char) * MaxCpuSetLength);
			strncpy(mcpu, cpuset, first - cpuset);
			splitcpuset = mcpu;
		}
		else
            splitcpuset = scpu;
	}

	return splitcpuset;
}
```

### How to reproduce

**Prerequisites**:
A heterogeneous Cloudberry cluster with:

- 1 coordinator node – 8 vCPUs (cores 0–7)
- 2 segment nodes – 18 vCPUs each (cores 0–17)

**Steps**:

1. Connect to the coordinator node via psql.
2. Create a resource group (or alter an existing one) with a cpuset that includes both coordinator and segment CPU ranges, separated by a semicolon: `ALTER RESOURCE GROUP rgroup1 SET CPUSET '4-7;9-17';`
3. The command will fail with an error similar to: `cpu cores %s are unavailable on the system`

I'm currently trying to fix it and rebuilding the code locally to verify the changes.

I hope the details I've provided above (root cause, reproduction steps, and proposed fix) will be helpful

### Operating System

Ubuntu 22.04

### Anything else

_No response_

### Code of Conduct

nix-oss:
Fixes #1862

### What does this PR do?
Fix inverted cpuset assignment in `getCpuSetByRole()`. The function returned the wrong cpuset for coordinator and segment roles when the cpuset string contains a semicolon separator (e.g., "0-7;0-15"). This caused incorrect CPU affinity settings on clusters. 

Additionally, this PR addresses the suggestion from the @my-ship-it to return a palloc'd copy instead of a pointer into the caller's string. 

A unit test for `getCpuSetByRole()` is also added to guard against future regressions.

### Type of Change
- [ ] Documentation update

### Test Plan

### Impact

**Performance:**
This fix ensures that CPU cores are assigned to the coordinator and segments exactly as configured by the administrator. It prevents unintended usage of cores that may be reserved for other critical system tasks.

**User-facing changes:**
Users can now successfully set cpuset strings with a semicolon separator (e.g., `ALTER RESOURCE GROUP ... SET CPUSET '4-7;9-17'`) without receiving a "cpu cores unavailable" error.

### Checklist

---

### DIFF
--- a/src/backend/commands/test/resgroupcmds_test.c
+++ b/src/backend/commands/test/resgroupcmds_test.c
@@ -1,0 +1,135 @@
+/*-------------------------------------------------------------------------
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ *
+ * resgroupcmds_test.c
+ *
+ * IDENTIFICATION
+ *	  src/backend/commands/test/resgroupcmds_test.c
+ *
+ *-------------------------------------------------------------------------
+ */
+
+#include <stdarg.h>
+#include <stddef.h>
+#include <setjmp.h>
+#include "cmockery.h"
+
+#include "../resgroupcmds.c"
+
+/*
+ * Helper: simulate running on the coordinator (dispatcher).
+ */
+static void
+set_role_coordinator(void)
+{
+	GpIdentity.segindex = MASTER_CONTENT_ID;
+}
+
+/*
+ * Helper: simulate running on a segment.
+ */
+static void
+set_role_segment(void)
+{
+	GpIdentity.segindex = 0;
+}
+
+/*
+ * NULL input must raise an ERROR.
+ */
+static void
+test__getCpuSetByRole_null_input(void **state)
+{
+	PG_TRY();
+	{
+		getCpuSetByRole(NULL);
+		fail_msg("expected ereport(ERROR) for NULL cpuset");
+	}
+	PG_CATCH();
+	{
+		FlushErrorState();
+	}
+	PG_END_TRY();
+}
+
+/*
+ * cpuset without a separator: same value must be returned
+ * regardless of the role.
+ */
+static void
+test__getCpuSetByRole_no_separator(void **state)
+{
+	const char *input = "0-7";
+	char *result;
+
+	set_role_coordinator();
+	result = getCpuSetByRole(input);
+	assert_string_equal(result, "0-7");
+
+	set_role_segment();
+	result = getCpuSetByRole(input);
+	assert_string_equal(result, "0-7");
+}
+
+/*
+ * cpuset with a separator, called as coordinator:
+ * must return the part BEFORE the ';'.
+ */
+static void
+test__getCpuSetByRole_with_separator_as_coordinator(void **state)
+{
+	const char *input = "0-7;0-15";
+	char *result;
+
+	set_role_coordinator();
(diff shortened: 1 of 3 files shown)


src/backend/commands/test/resgroupcmds_test.c
tuhaihe, line 1: For the newly created files, we need to add the standard Apache license header to them.

nix-oss: Thank you for pointing this out. I apologize for the oversight. The license header has been added to resgroupcmds_test.c and the changes are pushed.

zhangwenchao-123:
LGTM, but it's better to squash to one commit and add more commit message. What's more, this change may influence compatibility for greenplum or older cloudberry versions which maybe should be token into consideration.

nix-oss:
Hi @zhangwenchao-123.
Thanks for the approval and the comments! I've squashed the commits into one.

Regarding compatibility — this is indeed an important point. It seems this CPU management mechanism is either not heavily used, or limited to homogeneous environments, which is why this issue went unnoticed for so long.

zhangwenchao-123:
> Hi @zhangwenchao-123. Thanks for the approval and the comments! I've squashed the commits into one.
> 
> Regarding compatibility — this is indeed an important point. It seems this CPU management mechanism is either not heavily used, or limited to homogeneous environments, which is why this issue went unnoticed for so long.

Thanks for your contribution, LGTM.

zhangwenchao-123 (approved):

reshke (approved):

reshke:
IMHO this should be back-batched to rel 2

PR #1864, opened 2026-07-27
PAX: fix SIGSEGV in SUM stats merge during DELETE on minmax columns

### LINKED ISSUE #1767: [Bug] PAX DELETE may crash in MergeRawInfo when merging SUM stats for minmax_columns
### Apache Cloudberry version

2.1

### What happened

A PAX table may crash a segment during frequent `DELETE` / `INSERT` operations when `minmax_columns` contains columns that also have SUM statistics support, such as `int`, `bigint`, `numeric`, or similar numeric types.

The crash happens in the PAX DELETE visibility-map statistics refresh path.

example log : 

```

PANIC XX000 Unexpected internal error: Segment process received signal SIGSEGV

libpostgres.so datumCopy
pax.so cbdb::datumCopy
pax.so pax::MicroPartitionStats::MergeRawInfo
pax.so pax::MicroPartitionStatsUpdater::Update
pax.so pax::TableDeleter::UpdateStatsInAuxTable
pax.so pax::TableDeleter::DeleteWithVisibilityMap
```

The DELETE and INSERT statements should complete successfully.

PAX micro-partition statistics should be refreshed without crashing the segment.

### How to reproduce

```
DROP TABLE IF EXISTS pax_minmax_sum_dml_crash;

CREATE TABLE pax_minmax_sum_dml_crash (
    dist_key int,
    id bigint,
    amount bigint,
    k int,
    payload text
)
USING pax
WITH (minmax_columns = 'amount')
DISTRIBUTED BY (dist_key);

-- Put all rows on one segment to make the repro more likely to hit
-- multiple PAX internal groups in the same segment.
INSERT INTO pax_minmax_sum_dml_crash
SELECT
    1 AS dist_key,
    i::bigint AS id,
    i::bigint AS amount,
    (i % 1000)::int AS k,
    md5(i::text) AS payload
FROM generate_series(1, 800000) AS s(i);

-- Repeated small DELETE + INSERT operations.
-- On affected builds, one of the DELETE statements may crash a segment
-- during PAX micro-partition statistics refresh.
DO $$
DECLARE
    v_iter int;
    v_lo bigint;
BEGIN
    FOR v_iter IN 1..200 LOOP
        v_lo := ((v_iter - 1) * 1000 + 1)::bigint;

        DELETE FROM pax_minmax_sum_dml_crash
        WHERE id BETWEEN v_lo AND v_lo + 99;

        INSERT INTO pax_minmax_sum_dml_crash
        SELECT
            1 AS dist_key,
            (1000000000 + v_iter * 10000 + i)::bigint AS id,
            (1000000000 + v_iter * 10000 + i)::bigint AS amount,
            (i % 1000)::int AS k,
            md5((1000000000 + v_iter * 10000 + i)::text) AS payload
        FROM generate_series(1, 100) AS s(i);

        RAISE NOTICE 'finished iteration %', v_iter;
    END LOOP;
END $$;

SELECT count(*) FROM pax_minmax_sum_dml_crash;
```

### Operating System

rocky 9.6

### Anything else

Suspected root cause from CODEX:
```
PAX minmax_columns appear to maintain not only min/max stats but also SUM stats for supported column types.

During DELETE visibility-map refresh, MicroPartitionStatsUpdater::Update() may merge existing group-level raw stats through:

MicroPartitionStats::MergeRawInfo()
In MergeRawInfo(), serialized SUM stats appear to be deserialized using the physical column type metadata:

FromValue(..., typlen, typbyval, column_index)
However, the serialized SUM value should be interpreted using the aggregate return type metadata, for example:

sum(bigint) returns numeric
So a SUM result for a bigint column may be serialized as numeric, but later decoded as bigint. This can produce an invalid Datum and eventually crash in datumCopy().

There are also suspicious datumCopy() calls where typByVal and typLen appear to be passed in the wrong order, for example:

cbdb::datumCopy(newval, sum_stat->rettyplen, sum_stat->rettypbyval);
while the wrapper signature is:

datumCopy(Datum value, bool typByVal, int typLen)
This should likely be:

cbdb::datumCopy(newval, sum_stat->rettypbyval, sum_stat->rettyplen);
Similar calls may also exist in MicroPartitionStats::MergeTo().
```

### Code of Conduct

gongxun0928:
Fixes #1767

### What does this PR do?

Fixes a segment crash (SIGSEGV) that occurs when PAX tables with `minmax_columns` containing numeric types (int, bigint, numeric) undergo repeated DELETE operations. The crash happens in `datumCopy()` during the visibility-map statistics refresh path (`MicroPartitionStats::MergeRawInfo()`).

Two bugs are fixed in `contrib/pax_storage/src/cpp/storage/micro_partition_stats.cc`:

1. **Wrong type metadata in `FromValue()`**: The serialized SUM value was deserialized using the column's physical type (`typlen`/`typbyval`) instead of the SUM aggregate's return type (`rettyplen`/`rettypbyval`). For example, `sum(bigint)` returns `numeric`, so the stored SUM datum must be interpreted as `numeric`, not as `int8`. This mismatch produced an invalid `Datum` that crashed in `datumCopy()`.

2. **Swapped arguments in `datumCopy()`**: The call passed `(value, typlen, typbyval)` but the wrapper signature is `datumCopy(value, typByVal, typLen)`. This swapped pass-by-value/pass-by-reference flags, leading to memory corruption.

### Type of Change

### Breaking Changes
None.

### Test Plan
- [x] Unit tests added/updated — new regression test `delete_sum_stats` covering DELETE + INSERT + DELETE on int, bigint, and numeric columns with `minmax_columns` enabled

### Impact

**Performance:**
None — fix only affects correctness of stats merging, not the hot path.

**User-facing changes:**
Fixes a crash that users could hit with DELETE on PAX tables with minmax_columns.

**Dependencies:**
None.

### Checklist
- [x] Added/updated documentation (regression test included)

### Additional Context

Root cause analysis and reproduction steps are documented in issue #1767. The crash stack trace shows the failure path:

```
libpostgres.so datumCopy
pax.so cbdb::datumCopy
pax.so pax::MicroPartitionStats::MergeRawInfo
pax.so pax::MicroPartitionStatsUpdater::Update
pax.so pax::TableDeleter::UpdateStatsInAuxTable
pax.so pax::TableDeleter::DeleteWithVisibilityMap
```

### DIFF
--- a/contrib/pax_storage/expected/delete_sum_stats.out
+++ b/contrib/pax_storage/expected/delete_sum_stats.out
@@ -1,0 +1,53 @@
+-- Verify that DELETE statistics refresh uses the SUM result type when
+-- deserializing and merging stats. In particular, sum(bigint) returns numeric.
+set pax.max_tuples_per_group = 25;
+create table pax_delete_sum_stats (
+  dist_key int,
+  id bigint,
+  int_amount int,
+  bigint_amount bigint,
+  numeric_amount numeric
+)
+using pax
+with (minmax_columns = 'int_amount,bigint_amount,numeric_amount')
+distributed by (dist_key);
+insert into pax_delete_sum_stats
+select 1, i, i, i, i::numeric * 2
+from generate_series(1, 1000) i;
+select count(*) as rows,
+       sum(int_amount) as int_sum,
+       sum(bigint_amount) as bigint_sum,
+       sum(numeric_amount) as numeric_sum
+from pax_delete_sum_stats;
+ rows | int_sum | bigint_sum | numeric_sum 
+------+---------+------------+-------------
+ 1000 |  500500 |     500500 |     1001000
+(1 row)
+
+delete from pax_delete_sum_stats where id between 1 and 100;
+select count(*) as rows,
+       sum(int_amount) as int_sum,
+       sum(bigint_amount) as bigint_sum,
+       sum(numeric_amount) as numeric_sum
+from pax_delete_sum_stats;
+ rows | int_sum | bigint_sum | numeric_sum 
+------+---------+------------+-------------
+  900 |  495450 |     495450 |      990900
+(1 row)
+
+insert into pax_delete_sum_stats
+select 1, i, i, i, i::numeric * 2
+from generate_series(1001, 1100) i;
+delete from pax_delete_sum_stats where id between 301 and 400;
+select count(*) as rows,
+       sum(int_amount) as int_sum,
+       sum(bigint_amount) as bigint_sum,
+       sum(numeric_amount) as numeric_sum
+from pax_delete_sum_stats;
+ rows | int_sum | bigint_sum | numeric_sum 
+------+---------+------------+-------------
+  900 |  565450 |     565450 |     1130900
+(1 row)
+
+drop table pax_delete_sum_stats;
+reset pax.max_tuples_per_group;
--- a/contrib/pax_storage/pax_schedule
+++ b/contrib/pax_storage/pax_schedule
@@ -10,6 +10,7 @@ test: statistics_bloom_filter
 test: filter_tree filter_tree_arithmetic
 test: filter_tree_root_quals
 test: statistics/statistics 
+test: delete_sum_stats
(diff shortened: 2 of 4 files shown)


jiaqizho (approved):
lgtm

my-ship-it (approved):
LGTM, thanks for taking care of this.

PR #1865, opened 2026-07-27
fix: ExplainNode guard against NULL planstate on QE

roaldm153:
Fix segfault in ExplainNode() on a QE
A receiving Motion has no child PlanState when its child slice runs on
another gang.  ExplainNode() recursed into it anyway and crashed.
Return early when planstate is NULL.

Fixes #ISSUE_Number

### What does this PR do?
This change fixes a crash that occurs on the Query Executor (QE) when a subtree lives in a different slice.
The original code assumed that planstate is always non‑NULL and directly dereferenced planstate->plan.
When the subtree is not instantiated in the current slice, planstate can be NULL, leading to a segmentation fault.
The patch adds:

A NULL guard for planstate at the start of ExplainNode.
Moves the assignment plan = planstate->plan after the guard.
Additional guard when recursing into the outer plan to avoid calling ExplainNode with a NULL outerPlanState.
These changes make EXPLAIN robust across slice boundaries without altering any existing functionality.

### Type of Change
- [ ] Documentation update

### Breaking Changes
No breaking changes. The public API and plan execution behavior remain unchanged; only additional safety checks were added.

### Test Plan
Unit tests: No new unit tests were added because the fix is defensive and does not change the logical output of EXPLAIN.
Integration tests: Ran the full regression test suite, including parallel tests, and verified that the previously failing scenario no longer crashes.

### Impact
**Performance:**
Negligible impact – the added NULL checks are inexpensive branch predictions and do not affect the typical execution path.

**User-facing changes:**
ExplainPrintPlan now works on QE

**Dependencies:**
Nope

### Checklist

### Additional Context
ExplainNode was called from the QE side with a planstate that had not been instantiated because the subtree resides entirely in another slice.

### DIFF
--- a/src/backend/commands/explain.c
+++ b/src/backend/commands/explain.c
@@ -1557,13 +1557,12 @@ ExplainNode(PlanState *planstate, List *ancestors,
 	int			motion_recv;
 	int			motion_snd;
 	ExecSlice  *parentSlice = NULL;
-	/* 
-	 * Guard if subtree on QE lives in another slice 
-	 * and not instantiated here.
-	*/
+	/*
+	 * Guard against the case where a subtree on the QE lives in another slice
+	 * and is not instantiated in this slice.
+	 */
 	if (planstate == NULL)
 		return;
-	
 	plan = planstate->plan;
 
 	/* Remember who called us. */
--- a/src/backend/commands/explain.c
+++ b/src/backend/commands/explain.c
@@ -1540,7 +1540,7 @@ ExplainNode(PlanState *planstate, List *ancestors,
 			const char *relationship, const char *plan_name,
 			ExplainState *es)
 {
-	Plan	   *plan = planstate->plan;
+	Plan	   *plan;
 	PlanState  *parentplanstate;
 	ExecSlice  *save_currentSlice = es->currentSlice;    /* save */
 	const char *pname;			/* node type name for text output */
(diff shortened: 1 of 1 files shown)


Alena0704:
LGTM. I think you should rephrase the commit message. Maybe like that
```
Fix segfault in ExplainNode() on a QE

A receiving Motion has no child PlanState when its child slice runs on
another gang.  ExplainNode() recursed into it anyway and crashed.
Return early when planstate is NULL.
```
As far as I see this code is unreachable in-tree. The only callers of ExplainPrintPlan() are ExplainOnePlan(), which runs on the QD (for example, in auto_explain - it returns from _PG_init when Gp_role != GP_ROLE_DISPATCH (contrib/auto_explain/auto_explain.c:101)) - so it never installs its hooks on a segment. Only an out-of-tree extension loaded on the segments can hit this path or debuggers today.

That's fine as a hardening fix, but I think you should state it explicitly in the commit message.

src/backend/commands/explain.c
```diff
@@ -1557,13 +1557,12 @@ ExplainNode(PlanState *planstate, List *ancestors,
 	int			motion_recv;
 	int			motion_snd;
 	ExecSlice  *parentSlice = NULL;
-	/* 
-	 * Guard if subtree on QE lives in another slice 
-	 * and not instantiated here.
-	*/
+	/*
+	 * Guard against the case where a subtree on the QE lives in another slice
+	 * and is not instantiated in this slice.
```
Alena0704, line 1562: The comment doesn't say where the decision not to instantiate the subtree is made, so a reader has no starting point. I think you should add a pointer to ExecInitMotion here:

	/*
     * Guard against the case where a subtree on the QE lives in another slice
	 * and is not instantiated in this slice (see ExecInitMotion).
	 */

Alena0704 (approved):

zhangwenchao-123:
Squash to one commit?

Alena0704:
> Squash to one commit?

We could squash it when merging. As I remember, there’s an option for that in the GitHub UI. I think it makes sense to update the PR description based on the final commit message that will be used.

CC: @roaldm153

yjhjstz (approved):
LGTM

leborchuk (approved):
LGTM

PR #1868, opened 2026-07-28
[ORCA] Support hash partitioning in ORCA

zhangwenchao-123:
GPORCA previously rejected any query touching a hash-partitioned table and fell back to the Postgres planner. Enable GPORCA to plan such queries, with static partition pruning for equality predicates and dynamic (join-driven) partition elimination.

How:
- CTranslatorRelcacheToDXL: stop raising "hash partitioning" for a single-column, single-level hash partition key and let the strategy flow into GPORCA. Composite keys, partitioning by expression and multi-level partitioning still fall back to the planner.
- IMDRelation: add ErelpartitionHash ('h') to the partition-type enum.
- CExpressionPreprocessor: a hash leaf's qual is a satisfies_hash_partition() call with no btree-interval form, so PcnstrFromChildPartition now returns NULL instead of asserting. Add FHashPartitionPruned(): substitute the query's equality constants for the partition-key columns into that call and evaluate it with the constant-expression evaluator; a leaf whose call folds to false cannot hold a matching row and is pruned. This reuses PostgreSQL's exact hashing (seed, per-column proc, combine), so selection is identical.
- CConstExprEvaluatorDXL: allow folding any column-free immutable expression (e.g. satisfies_hash_partition over constants), not only (const cmp const). Volatile/stable and column-referencing expressions are still rejected.

Dynamic partition elimination needs no hash-specific code: once hash tables become CLogicalDynamicGet the existing CPhysicalPartitionSelector path is partition-type agnostic, and CPartPruneStepsBuilder resolves the strategy from the partition's own hash opfamily.

ORCA vs Postgres planner behavior:
- Partition selection and results are identical. For the supported single-column, single-level hash key, an equality predicate prunes to exactly the same surviving partition(s) as the Postgres planner and returns the same rows row-for-row. When a single partition survives, both narrow dispatch to that one segment (Gather Motion 1:1).
- Unchanged: composite keys, partitioning by expression and multi-level partitioning still fall back to the Postgres planner, and range/list pruning is unaffected.

Authored-by: Zhang Wenchao <zhangwenchao@apache.org>

Fixes #ISSUE_Number

### Type of Change
- [ ] Documentation update

### Test Plan

### Impact

**Performance:**

**User-facing changes:**

**Dependencies:**

### Checklist

### CI Skip Instructions

---

### DIFF
--- a/src/backend/gporca/libgpopt/src/operators/CExpressionPreprocessor.cpp
+++ b/src/backend/gporca/libgpopt/src/operators/CExpressionPreprocessor.cpp
@@ -22,6 +22,7 @@
 #include "gpopt/base/CConstraintInterval.h"
 #include "gpopt/base/COptCtxt.h"
 #include "gpopt/base/CUtils.h"
+#include "gpopt/eval/IConstExprEvaluator.h"
 #include "gpopt/exception.h"
 #include "gpopt/mdcache/CMDAccessor.h"
 #include "gpopt/operators/CDedupSupersetPreprocessor.h"
@@ -49,6 +50,9 @@
 #include "gpopt/operators/COrderedAggPreprocessor.h"
 #include "gpopt/operators/CPredicateUtils.h"
 #include "gpopt/operators/CScalarCmp.h"
+#include "gpopt/operators/CScalarConst.h"
+#include "gpopt/operators/CScalarFunc.h"
+#include "gpopt/operators/CScalarIdent.h"
 #include "gpopt/operators/CScalarNAryJoinPredList.h"
 #include "gpopt/operators/CScalarProjectElement.h"
 #include "gpopt/operators/CScalarProjectList.h"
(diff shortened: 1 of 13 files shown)


jiaqizho (approved):
Great implementation, lgtm

src/backend/gporca/libgpopt/src/operators/CExpressionPreprocessor.cpp
```diff
@@ -2995,25 +3034,216 @@ CExpressionPreprocessor::PcnstrFromChildPartition(
...
+	// the whole query: reset the error and conservatively keep the partition
+	// (pexprResult stays null, so we fall through to return false).
+	CExpression *pexprResult = nullptr;
+	GPOS_TRY
+	{
+		pexprResult = pceeval->PexprEval(pexprTest);
+	}
+	GPOS_CATCH_EX(ex)
+	{
+		GPOS_RESET_EX;
```
yjhjstz, line 3223: ignore exceptions ？

yjhjstz (approved):

PR #1871, opened 2026-07-30
diskquota: avoid flaky gpstop -ari in max_monitored_databases test

gongxun0928:
## Summary
- Replace `gpstop -ari` with `gpstop -arf` in `test_max_monitored_databases` (and redirect noisy output), matching other diskquota restart tests.
- Immediate restart forces crash recovery; with `hot_standby=off` the pidfile briefly reports `PM_STATUS_STANDBY`, so `pg_ctl -w` can return early and `gpstart` fails with `Hot standby mode is disabled`. Fast restart avoids that race.

## Test plan
- [x] Local: 15/15 PASS on `test_max_monitored_databases` (30× `gpstop -arf` restarts, 0 failures)
- [ ] CI: diskquota / related installcheck on this PR

### DIFF
--- a/gpcontrib/diskquota/tests/regress/expected/test_max_monitored_databases.out
+++ b/gpcontrib/diskquota/tests/regress/expected/test_max_monitored_databases.out
@@ -1,7 +1,6 @@
 --start_ignore
-\! gpconfig -c diskquota.max_monitored_databases -v 3
-20230905:12:39:55:332748 gpconfig:zhrt:zhrt-[INFO]:-completed successfully with parameters '-c diskquota.max_monitored_databases -v 3'
-\! gpstop -ari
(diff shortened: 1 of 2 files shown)


my-ship-it (approved):
LGTM — the diagnosis holds up:

The fix is right, and it eliminates the race rather than just narrowing it:
fast shutdown writes a shutdown checkpoint, so the restart is a clean start with
no recovery phase and no PM_STATUS_STANDBY window at all. diskquota.max_monitored_databases is PGC_POSTMASTER, so a normal restart
is sufficient — -i was never needed. It also matches tests/isolation2/sql/test_temporary_table.sql:17, which already uses gpstop -afr.

jiaqizho (approved):
lgtm

PR #1883, opened 2026-08-05
ORCA: fix CTE column-pruning misalignment across consumers

yjhjstz:
When a CTE has multiple consumers requiring different column subsets, the producer's shared-scan output is pruned to the union of all consumers' required columns (CTranslatorDXLToExpr::PruneCTEs). But each consumer independently decided its own output columns in CPhysicalCTEConsumer from its own per-column GetUsage(). A consumer that considers all of its columns used (e.g. SELECT a.*) kept every column with an identity index map, while the producer emitted only the pruned union. The consumer then read the shared tuple by stale positions, producing wrong results (a join key read from the wrong slot -> LEFT JOIN yields NULLs) and "invalid attnum N for relation shareX_refY" during EXPLAIN.

Drive the consumer's kept columns from the producer's finalized used mask (CLogicalCTEProducer::UsedMask) -- the single source of truth -- so every consumer exposes exactly the producer's surviving columns. When the producer was not pruned the mask is NULL and behavior is unchanged.

Add a regression test (cte_prune_multi_consumer) covering multi-consumer CTEs where a SELECT * consumer is referenced directly with a join key. It asserts the EXPLAIN no longer errors and returns correct results; the ORCA path is exercised with shared scans and matches the Postgres planner.

Fixes #ISSUE_Number

### Type of Change
- [X] Bug fix (non-breaking change)
- [ ] Documentation update

### Test Plan

### Impact

**Performance:**

**User-facing changes:**

**Dependencies:**

### Checklist

### CI Skip Instructions

---

### DIFF
--- a/src/backend/gporca/libgpopt/src/operators/CPhysicalCTEConsumer.cpp
+++ b/src/backend/gporca/libgpopt/src/operators/CPhysicalCTEConsumer.cpp
@@ -13,6 +13,7 @@
 
 #include "gpos/base.h"
 
+#include "gpopt/base/CCTEInfo.h"
 #include "gpopt/base/CCTEMap.h"
 #include "gpopt/base/COptCtxt.h"
 #include "gpopt/operators/CExpression.h"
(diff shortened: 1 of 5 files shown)


jiaqizho (approved):
lgtm

src/backend/gporca/libgpopt/src/operators/CPhysicalCTEConsumer.cpp
```diff
@@ -45,12 +46,30 @@ CPhysicalCTEConsumer::CPhysicalCTEConsumer(CMemoryPool *mp, ULONG id,
...
+	// CTranslatorDXLToExpr::PruneCTEs). Therefore every consumer must expose
+	// exactly the producer's surviving columns: it can neither read a column the
+	// producer pruned (that would run past the shared tuple) nor decide what to
+	// keep from its own per-column usage. Drive the consumer's kept set from the
+	// producer's used mask, which is the single source of truth. When the
+	// producer was not pruned (mask is NULL) fall back to the previous behavior.
+	CCTEInfo *pcteinfo = COptCtxt::PoctxtFromTLS()->Pcteinfo();
+	CLogicalCTEProducer *popProducer =
+		CLogicalCTEProducer::PopConvert(pcteinfo->PexprCTEProducer(m_id)->Pop());
+	BOOL *producer_umask = popProducer->UsedMask();
```
jiaqizho, line 60: nit: Since `producer_umask` represents the finalized layout and is only read here, declared as const BOOL *?

my-ship-it (approved):

PR #1884, opened 2026-08-05
Fix distributed snapshot race during commit

### LINKED ISSUE #1465: [Bug] Concurrent update cause select data error
### Apache Cloudberry version

```bash
gpadmin=# select version();
                                                                                                    version
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 PostgreSQL 14.4 (Apache Cloudberry 2.0.0-incubating build 1) on x86_64-pc-linux-gnu, compiled by gcc (GCC) 11.5.0 20240719 (Red Hat 11.5.0-11), 64-bit compiled on Dec  2 2025 19:35:44 (with assert checking)
(1 row)

[gpadmin@cdw ~]$ pg_config
BINDIR = /usr/local/cloudberry-db/bin
DOCDIR = /usr/local/cloudberry-db/share/doc/postgresql
HTMLDIR = /usr/local/cloudberry-db/share/doc/postgresql
INCLUDEDIR = /usr/local/cloudberry-db/include
PKGINCLUDEDIR = /usr/local/cloudberry-db/include/postgresql
INCLUDEDIR-SERVER = /usr/local/cloudberry-db/include/postgresql/server
LIBDIR = /usr/local/cloudberry-db/lib
PKGLIBDIR = /usr/local/cloudberry-db/lib/postgresql
LOCALEDIR = /usr/local/cloudberry-db/share/locale
MANDIR = /usr/local/cloudberry-db/share/man
SHAREDIR = /usr/local/cloudberry-db/share/postgresql
SYSCONFDIR = /usr/local/cloudberry-db/etc/postgresql
PGXS = /usr/local/cloudberry-db/lib/postgresql/pgxs/src/makefiles/pgxs.mk
CONFIGURE =  '--prefix=/usr/local/cloudberry-db' '--enable-cassert' '--enable-debug-extensions' '--enable-ic-proxy' '--enable-mapreduce' '--enable-orafce' '--enable-orca' '--enable-pxf' '--enable-tap-tests' '--with-gssapi' '--with-ldap' '--with-libxml' '--with-openssl' '--with-pam' '--with-perl' '--with-pgport=5432' '--with-python' '--with-pythonsrc-ext'
CC = gcc
CPPFLAGS = -D_GNU_SOURCE -I/usr/include/libxml2
CFLAGS = -Wall -Wmissing-prototypes -Wpointer-arith -Werror=vla -Wendif-labels -Wmissing-format-attribute -Wimplicit-fallthrough=3 -Wcast-function-type -Wformat-security -fno-strict-aliasing -fwrapv -fexcess-precision=standard -Wno-unused-but-set-variable -Werror=implicit-fallthrough=3 -Wno-format-truncation -Wno-stringop-truncation -O3 -fPIC  -DUSE_INTERNAL_FTS=1  -Werror=uninitialized -Werror=implicit-function-declaration -Werror
CFLAGS_SL = -fPIC
LDFLAGS = -Wl,--as-needed -Wl,-rpath,'/usr/local/cloudberry-db/lib',--enable-new-dtags -Werror
LDFLAGS_EX =
LDFLAGS_SL =
LIBS = -lpgcommon -lpgport -lxerces-c -lbz2 -lxml2 -lpam -lrt -lssl -lcrypto -lgssapi_krb5 -luv -lz -lreadline -lm  -lcurl -lzstd
VERSION = PostgreSQL 14.4

```

### What happened

select data error

```bash
## real result（multi rows）
gpadmin=*# select * from public.pgbench_branches where bid = 71;
 bid | bbalance | filler
-----+----------+--------
  71 |   868076 |
  71 |   785976 |
(2 rows)

gpadmin=*# \d public.pgbench_branches
s              Table "public.pgbench_branches"
  Column  |     Type      | Collation | Nullable | Default
----------+---------------+-----------+----------+---------
 bid      | integer       |           | not null |
 bbalance | integer       |           |          |
 filler   | character(88) |           |          |
Distributed by: (bid)

```

Image

select only 1 rows data
```bash
## expect result（only 1 rows）
gpadmin=*# select * from public.pgbench_branches where bid = 71;
 bid | bbalance | filler
-----+----------+--------
  71 |   xxx |
(1 rows)
```

### 1、set param
```bash
gpconfig -c gp_enable_global_deadlock_detector -v on
gpconfig -c max_connections -v 4096 -m 2048
gpconfig -c gp_vmem_protect_limit -v 16384
gpstop -r -M fast -a
```

### 2、prepare sql
- error_snapshot.sql
```bash
begin;
SET TRANSACTION ISOLATION LEVEL repeatable read;
SELECT pg_catalog.pg_export_snapshot();
\copy public.pgbench_branches to '/home/gpadmin/pgbench_branches.csv';
select count(1) from public.pgbench_branches;

DO $$
DECLARE
    user_count INTEGER;
    max_users INTEGER := 1000;
BEGIN
    select count(1) into user_count from public.pgbench_branches;

    IF user_count > max_users THEN
        perform pg_sleep(10000000);
    END IF;
END 
$$;

COMMIT;
```

- pgbench.sql
```bash
\set bid random(1, 100 * :scale)
\set delta random(-50000, 50000)
BEGIN;
UPDATE pgbench_branches SET bbalance = bbalance + :delta WHERE bid = :bid;
END;
```

### 3、reproduce
```bash
### init data
pgbench -i -s 1000 -h localhost -p 5432

## concurrent run pgbench and error_snapshot.sql
pgbench -M prepared -h localhost -p 5432 -T 30000 --rate 100000 -j 32 -c 512 -P 1 -f pgbench.sql
for i in $(seq 1 100000); do psql -f error_snapshot.sql; done
```

Image

### Operating System

NAME="Rocky Linux" VERSION="9.7 (Blue Onyx)" ID="rocky" ID_LIKE="rhel centos fedora" VERSION_ID="9.7" PLATFORM_ID="platform:el9" PRETTY_NAME="Rocky Linux 9.7 (Blue Onyx)" ANSI_COLOR="0;32" LOGO="fedora-logo-icon" CPE_NAME="cpe:/o:rocky:rocky:9::baseos" HOME_URL="https://rockylinux.org/" VENDOR_NAME="RESF" VENDOR_URL="https://resf.org/" BUG_REPORT_URL="https://bugs.rockylinux.org/" SUPPORT_END="2032-05-31" ROCKY_SUPPORT_PRODUCT="Rocky-Linux-9" ROCKY_SUPPORT_PRODUCT_VERSION="9.7" REDHAT_SUPPORT_PRODUCT="Rocky Linux" REDHAT_SUPPORT_PRODUCT_VERSION="9.7"

### Anything else

If the problem cannot be reproduced, try increasing the concurrency.

### Code of Conduct

hbswcsyzx:
Fixes #1465

### What does this PR do?

Fixes a distributed snapshot visibility race during distributed transaction
commit.

The coordinator keeps the distributed GXID in the proc array through the
existing transaction cleanup sequence. Immediately before the first QE commit
notification, it marks the transaction as `commitInProgress` while holding
`ProcArrayLock`. When a QD creates a distributed snapshot, it checks these
markers while holding the same lock. If a marked transaction is found, the QD
waits on the existing GXID heavyweight lock, then reacquires the proc-array
lock and builds the snapshot.

This preserves the required ordering:

- A snapshot that acquires the lock before the marker sees the GXID as in
  progress.
- A snapshot that acquires the lock after the marker waits until the QE commit
  notifications and normal QD cleanup have completed.
- `ProcArrayEndGxact()`, `includeInCkpt`, the FORGET record, and crash-recovery
  handling keep their existing lifecycle and ordering.

The one-phase path is marked before `COMMIT ONE PHASE` dispatch, and the
two-phase path is marked before `COMMIT PREPARED` dispatch, including retries.
`WaitForDtxCommit()` also reads the published GXID with `pg_atomic_read_u64()`
to match the existing atomic GXID protocol.

### Type of Change

- [ ] Documentation update

### Breaking Changes

None.

### Test Plan

Targeted validation was performed in Docker using a cluster built from this
branch:

- `issue1465_distributed_snapshot` passed: 2/2 isolation2 tests.
- The baseline reproduced the stale-version result; the fixed build returned
  the committed version for both reads using the same repeatable-read
  snapshot.
- Partial two-phase commit and repeated phase-2 retry tests kept the reader
  blocked until the transaction completed, then returned the committed value.
- One-phase commit success and injected QE failure/rollback paths passed.
- A checkpoint/crash-recovery test recorded the committed GXID, intentionally
  crashed the QD during phase 2, and recovered by re-dispatching `COMMIT
  PREPARED`; committed data was retained and no prepared transactions remained
  on either QE.
- A targeted workload with 16 writers and 8 repeatable-read readers completed
  2,700 writes and 18,948 reader transactions with correct results and no
  remaining prepared transactions.

### Impact

**Performance:**

Snapshot creation can wait during the existing QE commit-notification window.
The wait uses the existing local GXID lock and adds no network round trip.
Snapshots outside that window are unchanged.

**User-facing changes:**

This prevents distributed snapshots from observing mixed committed and
uncommitted versions of the same distributed update. There are no syntax,
catalog, or API changes.

**Dependencies:**

None.

### Checklist

### Additional Context

The earlier approach that removed the GXID from the proc array before QE
notification was superseded. The current patch does not remove or advertise a
GXID early; it adds a commit-in-progress marker and waits at QD distributed
snapshot creation. The code change is limited to the DTX/proc-array paths and
one isolation2 regression test.


zhangwenchao-123:
Hi guy, thank you for your contribution, but the pr looks also have some problems as following:
1: It breaks distributed snapshot atomicity.

The current ordering maintains a critical invariant: if a distributed snapshot considers a gxid completed (gxid < xmax and not in inProgressXidArray), the transaction has already committed locally on every QE. This is what makes a distributed read atomic across segments.

With this patch, a snapshot taken after ProcArrayEndGxact() but during the COMMIT PREPARED broadcast treats the gxid as committed while some QEs have not yet processed COMMIT PREPARED. On a QE, the distributed verdict only applies after the local xid is committed (the DistributedLog mapping is written at local commit); on a lagging QE the tuple falls back to local visibility rules, where the xid is still a prepared, in-progress transaction. Consequences:

  - Cross-segment mixed reads: the same snapshot sees the new tuple version on QEs that already committed and the old version on QEs that haven't. The window is the entire phase-2 broadcast, including the retry loop (ResetAllGangs() + retries bounded by dtx_phase2_retry_second), which can last seconds or longer.
  - Non-repeatable reads within a single snapshot: a REPEATABLE READ transaction reading a lagging QE before the commit is applied sees the old version; re-reading with the same snapshot after the QE applies COMMIT PREPARED sees the new version, because the verdict switches from the local fallback to the distributed "committed" verdict.
  - The one-phase path is worse: doNotifyingOnePhaseCommit() runs from prepareDtxTransaction() before the QD writes its own commit record and before the QE has committed anything, so the gxid disappears from snapshots before the transaction is durable anywhere.
  
2: It breaks checkpoint / crash recovery.

ProcArrayEndGxact() also clears tmGxact->gxid and tmGxact->includeInCkpt. getDtxCheckPointInfo() relies on scanning the proc array for entries with includeInCkpt set to include committed-but-not-yet-forgotten distributed transactions in the checkpoint record, so that crash recovery (recoverTM()) can re-dispatch COMMIT PREPARED to the QEs (see the protocol comment near getDtxCheckPointInfo() in CreateCheckPoint()). Today this flag is cleared only after the FORGET record(doInsertForgetCommitted()).

With this patch, if a checkpoint runs during the broadcast/retry window and the QD then crashes before the FORGET record is written, the committed transaction is missing from the checkpoint's committed-gxid list. Recovery will not re-deliver COMMIT PREPARED, and the prepared transactions on the QEs are treated as in-doubt and rolled back — a durably committed distributed transaction gets partially rolled back on the segments. Committed data is lost.

hbswcsyzx:
Thanks for identifying these issues. They exposed a real flaw in the earlier
approach, which removed the GXID from the proc array before QE commit
notification. That approach has been superseded by the current head
`4dc2724`.

The current fix preserves the original ordering. Before the first QE commit
notification, the QD sets a `commitInProgress` marker under exclusive
`ProcArrayLock`. A QD distributed snapshot either acquires the lock before the
marker and records the GXID as in progress, or acquires it after the marker and
waits on the existing GXID heavyweight lock. The wait finishes only after the
QE notifications return and normal QD cleanup reaches the existing proc-array
and lock-release points.

Therefore `ProcArrayEndGxact()` is not moved, `includeInCkpt` remains set until
the existing FORGET path, and checkpoint/crash recovery retains its original
GXID list behavior. The one-phase path sets the same marker before
`COMMIT ONE PHASE`, and the phase-2 marker covers the complete broadcast and
retry loop.

I also changed the wait path to use `pg_atomic_read_u64()` for `atomic_gxid`.
Targeted Docker validation covered the isolation2 regression, partial and
retrying phase-2 commits, one-phase commit, checkpoint/crash recovery, and a
16-writer/8-reader repeatable-read workload. The isolation2 test passed 2/2;
the recovery test retained committed data and left no prepared transactions.

hbswcsyzx:
@zhangwenchao-123 
Could a maintainer please approve the pending GitHub Actions runs for PR #1884?
Checks still show `action_required` and `Workflow runs completed with no jobs`.

hbswcsyzx:
The previous Debug workflow run failed before any source test was executed:
https://github.com/apache/cloudberry/actions/runs/32447855359

Error when evaluating 'strategy' for job 'test':
matrix must define at least one vector

The current PR head is 00249af after updating from main. It has no .github workflow differences from the current apache/main.

Lines 223-226 define `ALL_TESTS` with an empty `include` list, while line 767 passes the resulting empty value to `strategy.matrix`. The corresponding repository-relative path is https://github.com/apache/cloudberry/blob/main/.github/workflows/build-dbg-cloudberry.yml.

The same empty `ALL_TESTS` configuration exists in the current apache/main base branch. Therefore, this appears to be a pre-existing Debug workflow configuration issue, independent of the Issue #1465 source changes. The new Debug run will likely hit the same error once it runs.

Could a maintainer please advise whether this should be fixed separately in main, or whether a minimal non-empty Debug test-matrix repair is acceptable in this PR?

hbswcsyzx:
@tuhaihe

leborchuk:
I see coredumps in isolation2 tests

```
Thread 2 (Thread 0x7f18d36956c0 (LWP 63051)):
#0  0x00007f18e3180c2f in poll () from /lib64/libc.so.6
#1  0x00007f18e0a56052 in rxThreadFunc () from /usr/local/cloudberry-db-99.0.0/lib/postgresql/interconnect.so
#2  0x00007f18e311d128 in start_thread () from /lib64/libc.so.6
#3  0x00007f18e318dafc in clone3 () from /lib64/libc.so.6

Thread 1 (Thread 0x7f18e0df48c0 (LWP 63050)):
#0  0x00007f18e311ef8c in __pthread_kill_implementation () from /lib64/libc.so.6
#1  0x00007f18e30c8f56 in raise () from /lib64/libc.so.6
#2  0x00007f18e30b08fa in abort () from /lib64/libc.so.6
#3  0x0000000000e1a4cc in errfinish ()
#4  0x00000000007ab4ae in doNotifyingCommitPrepared.cold ()
#5  0x0000000000ed4ee1 in notifyCommittedDtxTransaction ()
#6  0x000000000088e85c in CommitTransaction ()
#7  0x0000000000890125 in CommitTransactionCommand ()
#8  0x0000000000c8a718 in exec_simple_query ()
#9  0x0000000000c8ed5e in PostgresMain ()
#10 0x0000000000bdafc6 in ServerLoop.isra.0 ()
#11 0x0000000000bdc218 in PostmasterMain ()
#12 0x00000000007debd2 in main ()
```

And fails are in `CommitTransaction` section. Not a good sign, most probably something is wrong with the current implementation.

hbswcsyzx:
Hi @leborchuk  and @tuhaihe ,

Thank you for your earlier review and guidance. The current revision has now passed the CI checks.

Whenever you have time, could you please take a look at the current implementation and let me know if you see any remaining correctness, performance, or test-coverage concerns that should be addressed before review approval? I would be happy to make any targeted revisions.

Thank you again for your time.

leborchuk:
> Hi @leborchuk and @tuhaihe ,
> 
> Thank you for your earlier review and guidance. The current revision has now passed the CI checks.
> 
> Whenever you have time, could you please take a look at the current implementation and let me know if you see any remaining correctness, performance, or test-coverage concerns that should be addressed before review approval? I would be happy to make any targeted revisions.
> 
> Thank you again for your time.

Ok, need some time to fully understand the issue and the fixes for it

PR #1890, opened 2026-08-14
Fix spurious global xmin > xmin error on rollback

gfphoenix78:
GetSnapshotData() on a QE computes xmin under ProcArrayLock but reads/advances DistributedLogShared->oldestXmin afterwards, in a different lock domain.  In that window a concurrent backend can advance the shared oldestXmin past this snapshot's xmin: when the transaction whose local xid equals xmin aborts, its distributed-log entry keeps distribXid == 0, so DistributedLog_AdvanceOldestXmin()'s forward scan skips over it instead of stopping, bumping oldestXmin beyond xmin.  The resulting globalxmin > xmin then tripped a hard elog(ERROR) and failed the query (seen intermittently in the partition_prune regression test).

The aborted xid is invisible to this snapshot anyway, and the visibility horizon only needs to be a valid lower bound, so a horizon at our own xmin loses nothing.  Clamp globalxmin to xmin instead of erroring -- always safe, and consistent with the clamp already done in DistributedLog_GetOldestXmin().

Fixes #ISSUE_Number

### Type of Change
- [ ] Documentation update

### Test Plan

### Impact

**Performance:**

**User-facing changes:**

**Dependencies:**

### Checklist

### CI Skip Instructions

---

### DIFF
--- a/src/backend/storage/ipc/procarray.c
+++ b/src/backend/storage/ipc/procarray.c
@@ -3312,9 +3312,29 @@ GetSnapshotData(Snapshot snapshot, DtxContext distributedTransactionContext)
 			globalxmin = DistributedLog_GetOldestXmin(globalxmin);
 	}
 
+	/*
+	 * globalxmin (the distributed oldestXmin read/advanced just above) can end
+	 * up greater than this snapshot's xmin.  xmin was computed under
+	 * ProcArrayLock, but DistributedLogShared->oldestXmin is maintained in a
+	 * different lock domain and is only read here, after ProcArrayLock has been
+	 * released.  In that window another backend can advance it past our xmin:
+	 * if the transaction whose local xid equals our xmin subsequently aborts,
+	 * its distributed-log entry keeps distribXid == 0, so
+	 * DistributedLog_AdvanceOldestXmin()'s forward scan skips over that xid
+	 * (instead of stopping at it) and bumps the shared oldestXmin beyond it.
+	 * A concurrent snapshot that took its xmin from that now-aborted xid then
+	 * reads the advanced value here.
+	 *
+	 * This is a benign, expected transient: the visibility horizon only needs
+	 * to be a valid lower bound (see src/backend/access/transam/README), and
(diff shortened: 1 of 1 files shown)


gfphoenix78:
The problem is:
```
gpadmin=# create temp table ttmp(a int);
ERROR:  global xmin (1076) is higher than transaction xmin (1075) (procarray.c:3298)  (seg0 127.0.0.1:7002 pid=2466718) (procarray.c:3298)
LINE 1: create temp table ttmp(a int);
                          ^
```

repro:
```sql
-- prepare
drop table if exists t1;

create table t1(a int);

-- T1
1: begin;
1: insert into t1 select i from generate_series(1,10)i;
-- to see the local xid on segments
1: select gp_segment_id, xmin, * from t1;

-- T2, will trigger the error
2: set optimizer=off; -- invoke the QE backend on segments
-- gdb attach to one of the QEs and break after release the ProcArrayLock
-- gdb -p xxxx
-- > b procarray.c:3303
-- > c

-- the following sql will be break when running GetSnapshotData
2: create temp table ttmp(a int);

-- rollback T1
1: rollback;

-- running SQL that advance DistributedLogShared.oldestXmin
1: insert into t1 select i from generate_series(1,10)i;

-- gdb: the QE of T2 should be still hanged in GetSnapshotData
-- quit the gdb now
-- > quit

-- now, the above SQL to create temp table will fail
```

gdb break the function GetSnapshotData at:
```c
	/*
	 * GP: In computing Globals, also take distributed snapshots into
	 * account.
	 */
	if (TransactionIdPrecedes(xmin, globalxmin))
		globalxmin = xmin;

	if (!IS_QUERY_DISPATCHER())
```

yjhjstz (approved):

leborchuk (approved):
LGTM, of course getting value not under the lock here means you could get any value

The issue is we do not `globalxmin` after that line.

Here the FIXME -  `	* GPDB_14_MERGE_FIXME: use globalxmin to compute Globals.` We take value, report error, but never use it. Should fix it )

PR #1907, opened 2026-08-19
Add makefile and preprocessing guards for --without-zstd configure

reshke:
$subj, otherwise re-configuring democluster doesn't work good with zstd options


leborchuk (approved):
LGTM

tuhaihe (approved):
LGTM.

I also did a fix on `with-zstd`: https://github.com/apache/cloudberry/pull/1912. PTAL.

PR #1762, opened 2026-05-22
Make PostgreSQL planner surpass ORCA on TPC-DS for the first time.

avamingli:
# Summary

For over a decade, the PostgreSQL planner has been considered inferior to ORCA for analytical workloads in Greenplum and Cloudberry. No one had ever systematically investigated why. This PR changes that.

Through forensic query-by-query analysis of all 99 TPC-DS queries at 1TB scale, I identified 12 fundamental deficiencies in how the PostgreSQL planner handles CTEs, predicate pushdown, parallel execution, cost estimation, and set operations. Each deficiency was addressed with a targeted optimization and validated against the full benchmark suite.
Old PG Planner is based on commit c49e871b02ce8e566455573ae7b074094a079773, New PG Planner is based on last commit of this PR.
**The result: the PostgreSQL planner now surpasses ORCA** on TPC-DS. Validated on both v3 and v4:

| Benchmark | Old PG Planner | ORCA | New PG Planner | New PG + 2 Parallel |
|---|---|---|---|---|
| **TPC-DS v3** | 5,331s | 3,185s | **2,605s** (1.22x faster than ORCA) | **2,325s** (1.37x faster than ORCA) |
| **TPC-DS v4** | 5,819s | 3,697s | **3,020s** (1.22x faster than ORCA) | **2,615s** (1.41x faster than ORCA) |

# Performance Results (TPC-DS v4)

**Environment:** SF=1000 (1TB), AOCO tables (zstd, level 5), 32 segments, single host, SSD. TPC-DS v4 benchmarks run via [cbdb_tpcds](https://github.com/avamingli/cbdb_tpcds) extension.

### Total Execution Time

chart_total_time

### ORCA vs New PG Planner (no parallelism) -- Pure Optimizer Duel

Without parallelism, on equal footing, the new PG planner already beats ORCA: **1.22x faster overall**, winning on 22 queries, tied on 59, slower on only 18.

chart_scatter_orca_vs_newpg

### Per-Query Comparison: Old PG vs ORCA vs New PG (no parallelism)

chart_3way_no_parallel

### ORCA vs New PG + 2 Parallel -- Parallel Bonus

With 2 parallel workers, the advantage widens to **1.41x faster than ORCA**: 67 wins, 24 ties, only 8 losses.

chart_scatter_orca_vs_new2p

### Cross-Benchmark Consistency (v3 + v4)

| Metric | TPC-DS v3 | TPC-DS v4 |
|---|---|---|
[...]

PR #1932, opened 2026-08-27
Materialized CTE (Shared Scan) OR predicate pushdown (#1 rebased from #1762)

Alena0704:
# Materialized CTE (Shared Scan) OR predicate pushdown

This is optimization only **#1** optimization from #1762 — *CTE Predicate Pushdown via OR Collection and CNF
Conversion* — split out on its own and rebased onto current `main` (PG16 kernel).

Original work by @avamingli; reviewed in #1762 by @leborchuk and @yjhjstz.

## What it does

When a CTE is materialized as a Shared Scan, consumer predicates cannot be pushed into the
producer, so the producer materializes every row any consumer might need. Following the ORCA
paper §6.1, this collects the quals of all CTE references, ORs them together, converts the
result to CNF and pushes it into the producer. The original predicates stay on each consumer,
so results are unchanged.

```sql
WITH v AS (SELECT i_brand, i_color FROM item WHERE i_current_price < 50)
SELECT * FROM v v1, v v2
WHERE v1.i_brand = v2.i_brand AND v1.i_color = 'red' AND v2.i_color = 'blue';
```

The producer now also carries `(i_color = 'red' OR i_color = 'blue')`.

CNF conversion is `convert_expr_to_cnf_complete()` in `prepqual.c`; it removes duplicate and
subsumed clauses so the expression does not blow up:

```
(s = 's' AND year = 2001) OR (s = 's' AND year = 2002)
  =>  s = 's' AND (year = 2001 OR year = 2002)
```

## Commits

Commits 1–6 are every commit in #1762 that touches this optimization, kept separate rather
than squashed; 7–11 are new here:

| | upstream commit | subject |
|---|---|---|
| 1 | `7f13fc4fa53` | Materialized CTE (Shared Scan) OR predicate pushdown |
| 2 | `4f341ff07ee` | Fix conflicts from main branch |
| 3 | `b9ae0e452cb` | Fix logical errors in OR-clause simplification |
| 4 | `544e67029cd` | Fix literal-vs-OR subsumption in CNF deduplication |
| 5 | `5e4635efda0` | Rename vague pushdown functions for clarity |
| 6 | `d2bd38d2e82` | Positive side effect (Direct Dispatch) of push down Shared Scan quals |
| 7 | new | Do not push CTE quals down when a consumer needs all rows |
| 8 | new | Compute `sub_total_rows` unconditionally in `set_cte_pathlist()` |
| 9 | new | Do not push CTE quals down into a CTE with volatile functions |
| 10 | new | Update the pax and singlenode copies of `subselect.out` |
| 11 | new | Update `tpcds_q04` expected plan for the CTE qual pushdown |

Rebase to PG16 needed two adjustments: `pushdown_safety_info.unsafeColumns` is now
`unsafeFlags` (bitmask), and `distribute_qual_to_rels()` was rewritten upstream — the
original commit's disabling of `Assert(root->hasLateralRTEs)` is preserved on top of the new
code, since pushed-down CTE quals can land outside their syntactic scope with no LATERAL RTE.

**Commit 7** fixes a wrong-results bug: if one CTE reference has no pushdown-safe qual (it
needs every row), its branch is missing from the disjunction and the producer gets filtered by
the other references' quals. #1762 fixes this in `54930251e4c`, which otherwise belongs to
optimization #2; commit 7 ports just the guard. Before it, a two-reference query where one
reference is unfiltered returned 1000 rows instead of 2666; three references with one
unfiltered returned 0 instead of 334.

**Commit 8** silences a GCC 13 `may be used uninitialized` warning on `sub_total_rows`: the
dummy-rel case is now handled before the loop instead of being skipped inside it. No
behaviour change.

**Commit 9** applies to the shared path the rule the inlined path already had: do not push
quals into a CTE containing volatile functions. Without it,

```sql
WITH x AS (SELECT a, nextval('ts') FROM vol_test) SELECT * FROM x WHERE a = 3
```

evaluated `nextval()` only for the matching row instead of once per row of `vol_test`, so the
value of the CTE's `nextval` column changed.

### Type of Change
- [ ] Documentation update

### Test Plan

### Impact

**Performance:**

**User-facing changes:**

**Dependencies:**

### Checklist

### CI Skip Instructions

---

### DIFF
--- a/contrib/pax_storage/src/test/regress/expected/subselect.out
+++ b/contrib/pax_storage/src/test/regress/expected/subselect.out
@@ -1904,7 +1904,7 @@ with x as materialized (select * from (select f1 from subselect_tbl) ss)
 select * from x where f1 = 1;
                      QUERY PLAN                     
 ----------------------------------------------------
- Gather Motion 3:1  (slice1; segments: 3)
+ Gather Motion 1:1  (slice1; segments: 1)
    Output: x.f1
    ->  Subquery Scan on x
          Output: x.f1
@@ -1913,9 +1913,10 @@ select * from x where f1 = 1;
                Output: share0_ref1.f1
                ->  Seq Scan on public.subselect_tbl
                      Output: subselect_tbl.f1
+                     Filter: (subselect_tbl.f1 = 1)
  Optimizer: Postgres query optimizer
  Settings: gp_cte_sharing=on
-(11 rows)
+(12 rows)
 
 -- Stable functions are safe to inline
 explain (verbose, costs off)
--- a/src/backend/optimizer/path/allpaths.c
+++ b/src/backend/optimizer/path/allpaths.c
@@ -65,6 +65,8 @@
 #include "cdb/cdbutil.h"
 #include "cdb/cdbvars.h"
 
+#include "optimizer/optimizer.h"
+
 // TODO: these planner gucs need to be refactored into PlannerConfig.
 bool		gp_enable_sort_limit = false;
 
@@ -194,6 +196,11 @@ static void bring_to_singleQE(PlannerInfo *root, RelOptInfo *rel);
 static bool is_query_contain_limit_groupby(Query *parse);
 static void handle_gen_seggen_volatile_path(PlannerInfo *root, RelOptInfo *rel);
 
+static List *
+collect_cte_quals(PlannerInfo *root, RelOptInfo *rel,
+				   RangeTblEntry *rte, Index rti, Query *subquery);
+
+static void subquery_push_qual_1(Query *subquery, Relids relids, Node *qual);
 
 /*
  * make_one_rel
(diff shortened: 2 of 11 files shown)


avamingli:
The original PR is still open,  see my comments: https://github.com/apache/cloudberry/pull/1762#issuecomment-5448086971 Suggest holding this PR to avoid divergence.

Alena0704:
> The original PR is still open, see my comments: [#1762 (comment)](https://github.com/apache/cloudberry/pull/1762#issuecomment-5448086971) Suggest holding this PR to avoid divergence.

Fair point on the heads-up - that's on me, I should have pinged you before opening #1932. With no reply on #1762 for weeks, I assumed the PR had been left behind, and I didn't want the work to go to waste. It wasn't meant to take the work over: authorship is yours, the six original commits are kept separate and attributed, changed only where the PG16 rebase required it.

I'm not ignoring your PR - I'm trying to get your ideas committed. Two things prevent that. First, the PR is very large, 98 commits, so it's very difficult to review and check that everything is fine. I only took the first point, and already there I found bugs, which I then fixed, plus a build issue in the form of a complaining compiler. One of those bugs is still present in your branch: quals are pushed into a CTE that contains volatile functions, while the inlined path in the same function refuses to do that. WITH x AS (SELECT a, nextval('ts') FROM vol_test) SELECT * FROM x WHERE a = 3 evaluates nextval() once instead of once per row, so the value of the CTE's column changes. Another one only shows up once point #1 is taken on its own: rows are lost when one CTE reference has no pushdown-safe qual, because the guard for that sits in 54930251e4c, which belongs to point #2. Same story with the expected files - tpcds_q04.out and the singlenode copy of subselect.out are only updated at the very end of the series, in 81d2f57e4d3 and b1f84b8902e, and the pax copy isn't updated anywhere, so point #1 on its own leaves them stale. I've only just noticed some of your own later fixes (described above).

Second, all the commits have fallen behind the main branch and need rebasing. I can't do this in your branch, which is why this PR appeared. Without resolving these issues, we'll never push anything.

As I see you want to rebase and split #1762 yourself, so I close #1932 - you can take these five commits if you want. You should fix somehow problem above.

avamingli:
> > The original PR is still open, see my comments: [#1762 (comment)](https://github.com/apache/cloudberry/pull/1762#issuecomment-5448086971) Suggest holding this PR to avoid divergence.
> 
> Fair point on the heads-up - that's on me, I should have pinged you before opening #1932. With no reply on #1762 for weeks, I assumed the PR had been left behind, and I didn't want the work to go to waste. It wasn't meant to take the work over: authorship is yours, the six original commits are kept separate and attributed, changed only where the PG16 rebase required it.
> 
> I'm not ignoring your PR - I'm trying to get your ideas committed. Two things prevent that. First, the PR is very large, 98 commits, so it's very difficult to review and check that everything is fine. I only took the first point, and already there I found bugs, which I then fixed, plus a build issue in the form of a complaining compiler. One of those bugs is still present in your branch: quals are pushed into a CTE that contains volatile functions, while the inlined path in the same function refuses to do that. WITH x AS (SELECT a, nextval('ts') FROM vol_test) SELECT * FROM x WHERE a = 3 evaluates nextval() once instead of once per row, so the value of the CTE's column changes. Another one only shows up once point #1 is taken on its own: rows are lost when one CTE reference has no pushdown-safe qual, because the guard for that sits in [5493025](https://github.com/apache/cloudberry/commit/54930251e4cc894fe5d29283f3009b3ce8e84f0f), which belongs to point #2. Same story with the expected files - tpcds_q04.out and the singlenode copy of subselect.out are only updated at the very end of the series, in [81d2f57](https://github.com/apache/cloudberry/commit/81d2f57e4d3a51248a99abd0403c3d4b8827962c) and [b1f84b8](https://github.com/apache/cloudberry/commit/b1f84b8902ede6131ba5ccc01327de66bccf7488), and the pax copy isn't updated anywhere, so point #1 on its own leaves them stale. I've only just noticed some of your own later fixes (described above).
> 
> Second, all the commits have fallen behind the main branch and need rebasing. I can't do this in your branch, which is why this PR appeared. Without resolving these issues, we'll never push anything.
> 
> As I see you want to rebase and split #1762 yourself, so I close #1932 - you can take these five commits if you want. You should fix somehow problem above.

 
On the two technical items:

Volatile functions in the CTE OR-pushdown — isn't this the same class of concern @leborchuk raised on the sublink-to-join path? https://github.com/apache/cloudberry/pull/1762#issuecomment-4876762835
> Except maybe IMMUTABLE or STABLE one. My experience tells me they are safe and could be moved.

Row loss when point #1 is taken alone, and stale expected files at intermediate commits — aren't these observations about a partial slice of the branch rather than the branch itself? I addressed the review methodology earlier: https://github.com/apache/cloudberry/pull/1762#issuecomment-4818492376 .
> For reviewing the behavior, I'd suggest evaluating the feature end-to-end rather than per-commit — no single intermediate commit reflects the final result; the incremental commits are the genuine iterative fixes that keep regression correct while improving DS performance.

Later commits in this series are deliberate fixes for issues introduced or surfaced by earlier ones; asking each intermediate to be correct in isolation will keep producing that list. End-to-end (branch HEAD) is the reviewable unit here.

On the rebase behind main: the divergence isn't drift. When #1760 (PG14→16 kernel upgrade, 5700+ planner-touching commits) was landing, I explicitly deferred this PR so the kernel upgrade could go in first 
> A quick update on https://github.com/apache/cloudberry/pull/1760. Jinbao and I talked it through, and we agree the kernel upgrade should take priority — see the background discussion here: https://lists.apache.org/thread/pmwp8v6zg7ds1jg4r9lttkoojhjmxy11

see https://github.com/apache/cloudberry/pull/1762#issuecomment-4539709031.

PR #1941, opened 2026-08-31
Fix ORCA GbAgg drop for nullable UNIQUE keys

roseduan:
CXformSimplifyGbAgg::FDropGbAgg() drops a GROUP BY/DISTINCT's aggregate node when the grouping columns cover a key derived from a UNIQUE or PRIMARY KEY constraint, assuming the key already guarantees one row per value.

That's wrong for a plain UNIQUE column: SQL allows multiple NULLs in it (NULL is never "equal" for uniqueness checks), but dedup must still collapse those NULLs into one row.

Before:
    create table repro (c0 numeric unique);
    insert into repro values (1), (2), (null), (null), (null);
    explain (costs off) select distinct c0 from repro;
        ->  Seq Scan on repro          -- no aggregate, all 5 rows returned

After:
        ->  GroupAggregate
              Group Key: repro.c0
              ->  Seq Scan on repro    -- correctly returns 1, 2, NULL

Fix by also requiring every key column to be provably NOT NULL (CExpression::DeriveNotNullColumns()) before dropping the aggregate. PRIMARY KEY columns are always NOT NULL, so that case is unaffected.

Fixes #ISSUE_Number

### Type of Change
- [ ] Documentation update

### Test Plan

### Impact

**Performance:**

**User-facing changes:**

**Dependencies:**

### Checklist

### CI Skip Instructions

---

### DIFF
--- a/src/test/regress/expected/bfv_aggregate.out
+++ b/src/test/regress/expected/bfv_aggregate.out
@@ -2172,6 +2172,75 @@ commit;
 ERROR:  duplicate key value violates unique constraint "t3_pkey"  (seg1 127.0.0.1:7003 pid=86457)
 DETAIL:  Key (a, b)=(1, 1) already exists.
 drop table  t1, t2, t3, t4, t5, t6;
+-- Test NULLs in a UNIQUE key column. A UNIQUE constraint allows multiple NULLs
+-- (NULL is never "equal" for uniqueness checks), but grouping columns that
+-- cover such a key must still collapse those NULLs into a single row, so
+-- dropping the GbAgg is only safe when every key column is also NOT NULL.
+drop table if exists t7, t8, t9;
+NOTICE:  table "t7" does not exist, skipping
+NOTICE:  table "t8" does not exist, skipping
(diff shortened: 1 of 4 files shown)


src/test/regress/expected/bfv_aggregate.out
```diff
@@ -2172,6 +2172,75 @@ commit;
...
+NOTICE:  table "t7" does not exist, skipping
+NOTICE:  table "t8" does not exist, skipping
+NOTICE:  table "t9" does not exist, skipping
+create table t7 (a int unique, b int);
+create table t8 (a int unique not null, b int);
+create table t9 (a int primary key, b int);
+insert into t7 values (1, 1), (2, 2), (null, 3), (null, 4), (null, 5);
+insert into t8 values (1, 1), (2, 2), (3, 3);
+insert into t9 values (1, 1), (2, 2), (3, 3);
+explain (costs off) select distinct a from t7;
```
yjhjstz, line 2189: try test on partitioned table ?

PR #1946, opened 2026-09-02
Fix IPv6 crash: drop dead HAVE_IPV6 guards left after PG16 merge

leborchuk:
Upstream PostgreSQL commit "Remove configure probe for sockaddr_in6 and require AF_INET6." (bcc8b14) deleted the HAVE_IPV6 probe and stripped the #ifdef HAVE_IPV6 guards from its own code, since AF_INET6 is now always available. Cloudberry-specific code still gated IPv6 handling behind #ifdef HAVE_IPV6, so after the PG16 merge those blocks became dead code: HAVE_IPV6 is never defined on non-Windows builds.

The practical effect: on an IPv6-only cluster, getDnsCachedAddress() never populates its cache entry (the IPv6 branch was compiled out), then returns e->hostinfo with e == NULL -- a bogus non-NULL pointer (offsetof key[]) -- which the caller passes to pstrdup(), crashing in strlen(). This shows up as a coordinator/FtsProbe SIGSEGV:

  #0 __strlen_evex
  #1 MemoryContextStrdup
  #2 getCdbComponentInfo
  #3 cdbcomponent_getCdbComponents
  #4 FtsProbeMain

Remove the leftover #ifdef HAVE_IPV6 guards so the IPv6 paths compile unconditionally, matching what upstream did to its own files. Also guard the cache return against a NULL entry so an unresolvable segment logs a clean "cannot resolve network address" error instead of segfaulting.

Files: cdbutil.c (both getDnsCachedAddress copies), auth.c, and the interconnect listener setup (ic_common.c, ic_tcp.c, ic_udpifc.c).


yjhjstz (approved):
LGTM

Smyatkin-Maxim (approved):

PR #1956, opened 2026-09-03
Disable COPY FROM PROGRAM under separate build-time option

reshke:
$subj. 
Our fork with private patches has some security-related patched. One of them is about very strict security hardening for some deployments. 
I have reimplemented this patch in upstream-compatible way, so no current use case in Cloudberry will be harmed, while we can remove diff in out fork.


leborchuk (approved):
Yep, LGTM, let's compile without option

PR #1960, opened 2026-09-04
Fix join condition lost after pull up sublink to join

Alena0704:
Fix join condition lost after pull up sublink to join
    
After pulling up the sublink to join, the raw join condition may get
lost in the rewritten query, potentially leading to incorrect results.
Within the SubqueryToJoinWalker() function, we address this issue
by adding an 'else' branch to prevent the loss of join clauses and
keep them in their original positions.

(cherry picked from open-gpdb commit c06d16b86316e1d66182f5c75a6c3abe6c30acd4)

### Type of Change
- [ ] Documentation update

### Test Plan

### Impact

**Performance:**

**User-facing changes:**

**Dependencies:**

### Checklist

### CI Skip Instructions

---

### DIFF
--- a/src/backend/cdb/cdbsubselect.c
+++ b/src/backend/cdb/cdbsubselect.c
@@ -451,6 +451,13 @@ SubqueryToJoinWalker(Node *node, ConvertSubqueryToJoinContext *context)
 		 */
 		context->safeToConvert = false;
 	}
+	else
+	{
+		/*
+		 * For other expressions, we should keep them in original place.
+		 */
+		context->innerQual = make_and_qual(context->innerQual, node);
+	}
 
 	return;
 }
--- a/src/test/regress/expected/subselect.out
+++ b/src/test/regress/expected/subselect.out
@@ -2324,3 +2324,85 @@ select (select max((select t.i))) from t;
 (1 row)
 
 drop table t;
+-- Fix join condition expression lost as pull up sublink to join.
+create table tl1(a int, b int, c int, d int) distributed by (a);
+create table tl2(a int, b int, c int, d int) distributed by (a);
+create table tl3(a int, b int, c int, d int) distributed by (a);
+create table tl4(a int, b int, c int, d int) distributed by (a);
+insert into tl1 values (-1, 3, 1, 0);
+insert into tl2 values (2, 1, 1, 0);
+insert into tl2 values (3, 1, 1, 0);
+insert into tl2 values (1, 1, 1, 0);
+insert into tl3 values (9, 9, 1, 9);
+insert into tl4 values (-1, -1, -1, -1);
+explain(costs off, verbose on)
+select * from tl1
+where
+  tl1.b = (
+    select
+      max(tl2.a)
+    from
+      tl2 join tl4
+      on tl4.d = tl2.d
+    where
+      tl2.b = tl1.c
+  );
+                                            QUERY PLAN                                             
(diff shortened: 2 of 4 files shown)


Alena0704:
When a correlated aggregate subquery is pulled up into a join, `SubqueryToJoinWalker()` records only two kinds of quals: non-correlated branches found *inside* an AND `BoolExpr`, and correlated equality `OpExpr`s that become the grouping key. A single non-correlated qual matches neither branch — it falls through to the final `return` and is recorded nowhere. `RemoveInnerJoinQuals()` then clears `je->quals`, and `subselect->jointree->quals` is overwritten with what the walker collected, so the predicate is gone from the rewritten query.

A one-predicate `JOIN ... ON` clause is exactly such a qual, so its condition is silently dropped and the aggregate is computed over an unrestricted join:

```sql
set optimizer = off;

create table o (a int, d int);  insert into o  values (2, 1);
create table i1(a int);         insert into i1 values (1);
create table i2(a int);         insert into i2 values (1), (2);

select * from o where o.a > (select max(i2.a) from i1 join i2 on i2.a = i1.a where i1.a = o.d);
```

```
postgres=# create table o (a int, d int);  insert into o  values (2, 1);
NOTICE:  Table doesn't have 'DISTRIBUTED BY' clause -- Using column named 'a' as the Apache Cloudberry data distribution key for this table.
HINT:  The 'DISTRIBUTED BY' clause determines the distribution of data. Make sure column(s) chosen are the optimal data distribution key to minimize skew.
CREATE TABLE
INSERT 0 1
postgres=# create table i1(a int);         insert into i1 values (1);
NOTICE:  Table doesn't have 'DISTRIBUTED BY' clause -- Using column named 'a' as the Apache Cloudberry data distribution key for this table.
HINT:  The 'DISTRIBUTED BY' clause determines the distribution of data. Make sure column(s) chosen are the optimal data distribution key to minimize skew.
CREATE TABLE
INSERT 0 1
postgres=# create table i2(a int);         insert into i2 values (1), (2);
NOTICE:  Table doesn't have 'DISTRIBUTED BY' clause -- Using column named 'a' as the Apache Cloudberry data distribution key for this table.
HINT:  The 'DISTRIBUTED BY' clause determines the distribution of data. Make sure column(s) chosen are the optimal data distribution key to minimize skew.
CREATE TABLE
INSERT 0 2
postgres=# 
postgres=# -- returns (2,1) as a SubPlan, returns nothing when pulled up
postgres=# select * from o where o.a > (select max(i2.a) from i1 join i2 on i2.a = i1.a where i1.a = o.d);
 a | d 
---+---
 2 | 1
(1 row)

postgres=# set optimizer = off;
SET
postgres=# select * from o where o.a > (select max(i2.a) from i1 join i2 on i2.a = i1.a where i1.a = o.d);
 a | d 
---+---
(0 rows)
```

`i1.a = 1`, so `ON` leaves only `i2.a = 1` and the subquery is `1`; `2 > 1` holds and the row must be returned. Instead the query returns nothing, because a `Nested Loop` with no `Join Cond` producing 2 rows instead of 1, which makes `max` come out as 2.

The fix adds an `else` branch that keeps such quals where they were. Only the Postgres planner is affected (`optimizer=off`, or an ORCA fallback); ORCA does its own decorrelation and preserves the `ON` condition.

leborchuk (approved):
LGTM, let's place it without modification

yjhjstz (approved):
LGTM

Issue #1968, opened 2026-09-07
[Bug] Management utilities print SyntaxWarning: invalid escape sequence on Python 3.12+ (e.g. Rocky Linux 10)


tuhaihe:
### Apache Cloudberry version

main (89baa48a4f4) — also present in 2.x releases.

### What happened

On a Rocky Linux 10 cluster (Python 3.12), every invocation of `gpsync` /
`gpssh` / `gpssh-exkeys` prints `SyntaxWarning` messages to stderr before it
does any work. It surfaced while installing the Cloudberry backup utilities,
whose `make install` calls `gpsync` to distribute binaries to the segments:

```
$ make install
cp /home/gpadmin/go/bin/gpbackup ... /usr/local/cloudberry-db/bin
/usr/local/cloudberry-db/lib/python/gppylib/util/ssh_utils.py:268: SyntaxWarning: invalid escape sequence '\ '
  '''Escape occurrences of \ and $ as needed and package the line as an "eval" shell command'''
/usr/local/cloudberry-db/bin/lib/pexpect/pxssh.py:105: SyntaxWarning: invalid escape sequence '\['
  self.UNIQUE_PROMPT = "\[PEXPECT\][\$\#] "
/usr/local/cloudberry-db/bin/lib/pexpect/pxssh.py:109: SyntaxWarning: invalid escape sequence '\$'
  self.PROMPT_SET_SH = "PS1='[PEXPECT]\$ '"
/usr/local/cloudberry-db/bin/lib/pexpect/pxssh.py:110: SyntaxWarning: invalid escape sequence '\$'
  self.PROMPT_SET_CSH = "set prompt='[PEXPECT]\$ '"
Successfully copied gpbackup_helper and gpbackup_s3_plugin to /usr/local/cloudberry-db on all segments
```

The operation itself succeeds — this is cosmetic — but it makes utility output
noisy and looks like a failure to anyone reading a deployment log.

Why it appears now: string literals such as `"\["` or `"\$"` contain escape
sequences Python does not recognise. Until 3.11 that was a `DeprecationWarning`,
hidden by default; **Python 3.12 promoted it to a `SyntaxWarning`, which is
printed** ([What's New in 3.12](https://docs.python.org/3/whatsnew/3.12.html#other-language-changes)).
So the same code is quiet on Rocky 8/9 (Python 3.6/3.9) and noisy on Rocky 10,
Fedora 40+, and Ubuntu 24.10+. The Python docs also state these sequences
[...]

PR #1969, opened 2026-09-07
Fix invalid escape sequences in gpMgmt Python sources

tuhaihe:
Python 3.12 promoted "invalid escape sequence" from a DeprecationWarning (hidden by default) to a SyntaxWarning that is printed whenever the module is compiled. On distributions shipping Python 3.12 or newer -- Rocky Linux 10, Fedora 40+, Ubuntu 24.10+ -- the management utilities therefore print warnings before doing any work:

  /usr/local/cloudberry-db/lib/python/gppylib/util/ssh_utils.py:268:
    SyntaxWarning: invalid escape sequence '\ '
  /usr/local/cloudberry-db/bin/lib/pexpect/pxssh.py:105:
    SyntaxWarning: invalid escape sequence '\['
  /usr/local/cloudberry-db/bin/lib/pexpect/pxssh.py:109:
    SyntaxWarning: invalid escape sequence '\$'
  /usr/local/cloudberry-db/bin/lib/pexpect/pxssh.py:110:
    SyntaxWarning: invalid escape sequence '\$'

The four above come from gpsync/gpssh/gpssh-exkeys and are the ones users see; compiling everything under gpMgmt/ turns up 41 such literals in 16 files.  Nothing misbehaves today -- CPython leaves an unrecognised escape in the string as-is, which happens to be what a regex or a shell snippet wants -- but the warnings are noise on stderr, and the Python docs say these sequences will become a SyntaxError in a future release.

Fix the literals rather than silencing the warning:

  * a literal whose every backslash escape is invalid gets an r prefix
  * a literal that mixes valid and invalid escapes has the invalid backslash doubled instead, since r would also change the meaning of the valid ones

Both transformations leave the literal's runtime value byte-for-byte identical, which was verified mechanically: for each touched file the ordered list of str/bytes constants in the AST is unchanged against the parent commit, and compiling the file under -W error::SyntaxWarning is now clean.

Two of the rewrites are worth a reviewer's eye:

  * gpstate_utils.py:79 and replication_slots_utils.py:31 are the mixed-escape case, so they read '\\%' and '\\A' now.
  * the '\A' in replication_slots_utils.py looks like a stray keystroke in a run of shell line continuations -- the shell it is handed sees "&& A ./demo_cluster.sh" and tries to run A.  That is pre-existing behaviour, so this commit preserves it exactly rather than quietly changing what the behave step does; it wants a separate fix.

gpMgmt/bin/lib/pexpect is a vendored copy of pexpect 3.3; upstream pexpect made these same literals raw long ago.

27 more files under src/ and contrib/ (mostly gporca and try_convert developer scripts) have the same problem and are left for a follow-up.

Fixes #ISSUE_Number

### Type of Change
- [ ] Documentation update

### Test Plan

### Impact

**Performance:**

**User-facing changes:**

**Dependencies:**

### Checklist

### CI Skip Instructions

---

### DIFF
--- a/gpMgmt/bin/gpload.py
+++ b/gpMgmt/bin/gpload.py
@@ -717,7 +717,7 @@ def quote_unident(val):
 
 def match_notice_obj(notice):
     # match the formatting errors in notice
-    r = re.compile("^NOTICE:  found (\d+) data formatting errors.*")
+    r = re.compile(r"^NOTICE:  found (\d+) data formatting errors.*")
     m = r.match(notice)
     if m:
         return int(m.group(1))
@@ -2527,7 +2527,7 @@ WHERE relname = 'staging_gpload_reusable_%s';""" % (encoding_conditions)
         # should not explicitly specify the DISTRIBUTED BY clause.
         # Only the DISTRIBUTED BY clause can take effect if all selected fields 
         # exist in the CREATE TABLE statement.
-        dist_column_list = re.match(".*\((.*)\).*", distcols).group(1).split(",")
+        dist_column_list = re.match(r".*\((.*)\).*", distcols).group(1).split(",")
         target_column_set = set(element[0] for element in target_columns)
         if set(dist_column_list) <= target_column_set:
             quoted_dist_column = convertListToDelimited(dist_column_list)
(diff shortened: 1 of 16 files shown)


tuhaihe:
#1968

MisterRaindrop (approved):
LGTM

leborchuk (approved):
LGTM

PR #1977, opened 2026-09-09
Upgrade gpMgmt's bundled Python dependencies

tuhaihe:
PyYAML 5.4.1 no longer builds its C extension on Rocky 10. Its pre-generated _yaml.c declares the libyaml read/write handlers with char * where libyaml wants unsigned char *, and GCC 14 turned that from a warning into an error:

```
  yaml/_yaml.c:3598:52: error: passing argument 2 of 'yaml_parser_set_input' from incompatible pointer type [-Wincompatible-pointer-types]
  /usr/include/yaml.h:1374:30: note: expected 'int (*)(void *, unsigned char *, size_t,  size_t *)' but argument is of type 'int (*)(void *, char *, size_t,  size_t *)'
  Error compiling module, falling back to pure Python
```

The build then completes with the slow pure-Python parser, so gpload still works; this is quality rot rather than a hard failure. Note that the CFLAGS="-w" already passed here cannot suppress it, since -w only silences warnings and GCC 14 makes this an error.

While in here, the other two bundled packages are equally old: PyGreSQL 5.2 is from 2020-06, psutil 5.7.0 from 2020-02.

```
  PyGreSQL  5.2   -> 5.2.5
  psutil    5.7.0 -> 7.2.2
  PyYAML    5.4.1 -> 6.0.1
```

The ceilings are set by Python 3.6, which Rocky 8 ships as its system python3 and which is still a supported build platform:

* PyYAML 6.0.1 is the newest release declaring >=3.6; 6.0.2 moved to >=3.8.

* PyGreSQL stays on the 5.x line on purpose. 6.x requires >=3.7 and its setup.py refuses older interpreters outright ("Sorry, PyGreSQL 6.1.0 does not support this Python version"). 5.2.5 is also a drop-in, which matters because gpMgmt imports the top-level pg and pgdb modules in eleven files.

* psutil is the exception: even 7.2.2 still declares >=3.6, and the only API gpMgmt uses is virtual_memory(), Process(), Process.children(), Process.memory_info(), wait_procs(), pid_exists() and NoSuchProcess, none of which changed in 6.x or 7.x.

Verified on all five build platforms (Rocky 8/9/10, Ubuntu 22.04/24.04) by running the real download-python-deps, pyyaml, psutil and pygresql targets: all succeed with no errors and no pure-Python fallback, and importing yaml, psutil, pg and pgdb out of the resulting gpMgmt/bin/ext exercises the calls listed above. Rocky 8 covers the Python 3.6 end and Rocky 10 the GCC 14 end.

This is not a security fix: OSV reports no known advisories against either the old or the new versions.

Two things worth knowing before the next bump. From PyYAML 6.0.2 and PyGreSQL 6.0.1 onwards the sdist filename is lowercase, which the hardcoded PyYAML-$(PYYAML_VERSION).tar.gz style URLs and directory names here do not handle; since curl runs without -f the 404 body lands in the tarball and the failure surfaces at tar time. And the "cython<3.0.0" pin in download-python-deps is load-bearing: PyYAML re-cythonizes when Cython is importable, and both 5.4.1 and 6.0.1 die with "AttributeError: cython_sources" under Cython 3.

Assisted-by: Claude Code

Fixes #ISSUE_Number

### Type of Change
- [ ] Documentation update

### Test Plan

### Impact

**Performance:**

**User-facing changes:**

**Dependencies:**

### Checklist

### CI Skip Instructions

---


leborchuk (approved):
LGTM, let's upgrade dependencies. 6.0.1 should be good enough

MisterRaindrop (approved):

reshke (approved):

PR #1980, opened 2026-09-10
Make initial value of the gp_create_table_random_default_distribution GUC equal to the default one

andr-sokolov:
The gp_create_table_random_default_distribution variable was initialized with
true, but the default value of the corresponding GUC is false. Set the initial
value to the GUC's default value to make the code easier to understand.


leborchuk (approved):
LGTM, by default it's  false https://github.com/apache/cloudberry/blob/15d75c672ad360a11f1b7204218a35c016479f0c/src/backend/utils/misc/guc_gp.c#L1837

yjhjstz (approved):

PR #1981, opened 2026-09-10
fix(planner): fix range-pair NULL adjustment under outer joins 

### LINKED ISSUE #1950: [Bug] Planner: FailedAssertion "pselec >= 0.0 && pselec <= 1.0" in adjust_selectivity_for_nulltest() for OR clause under LEFT JOIN
### Apache Cloudberry version

main branch (recent build, assert-enabled)

### What happened

On an assert-enabled build, the planner (not ORCA) hits an assertion in join size estimation when an OR clause mixes a predicate on the outer side of a LEFT JOIN with a range predicate on the nullable inner side:

```
FailedAssertion("pselec >= 0.0 && pselec <= 1.0", costsize.c:5506)
```

The assertion is the last line of `adjust_selectivity_for_nulltest()`, which is Cloudberry-specific code called from `calc_joinrel_size_estimate()` under `gp_adjust_selectivity_for_outerjoins` (default on). Upstream PostgreSQL has no equivalent assertion. `pselec` comes from `clauselist_selectivity()` on the pushed-down OR clause and is already outside [0, 1] before the null-test adjustment (which does not fire here since the clause is a BoolExpr). Which sub-term produces the out-of-range value has not been narrowed down yet. On a non-assert build this silently produces a bad row estimate instead of crashing. Line number is from current `main`.

Found by SQLancer.

The selectivity should stay within [0, 1] and the query should return a count without crashing.

### How to reproduce

```sql
CREATE TABLE m1(c0 inet);
CREATE TABLE m2(c0 inet);
INSERT INTO m2 VALUES ('88.147.138.141'), ('76.163.212.11'), ('214.10.65.144');
ANALYZE m1, m2;

SELECT COUNT(*) FROM ONLY m1 LEFT OUTER JOIN m2 ON true
WHERE (m1.c0 IS NOT NULL)
   OR (m2.c0 BETWEEN SYMMETRIC '75.175.243.19' AND '230.9.216.68');
-- FailedAssertion("pselec >= 0.0 && pselec <= 1.0", costsize.c:5506)
```

### Operating System

Linux

### Anything else

Found while running SQLancer against Cloudberry `main` (3-segment demo cluster, `--enable-cassert`). See the proposal to run SQLancer continuously: https://github.com/apache/cloudberry/discussions/1952

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

yjhjstz:
On an assert-enabled build the following query crashed the QD with FailedAssertion("pselec >= 0.0 && pselec <= 1.0", costsize.c:5506):

  CREATE TABLE m1(c0 inet);
  CREATE TABLE m2(c0 inet);
  INSERT INTO m2 VALUES ('88.147.138.141'), ('76.163.212.11'), ('214.10.65.144');
  ANALYZE m1, m2;
  SELECT COUNT(*) FROM ONLY m1 LEFT OUTER JOIN m2 ON true
  WHERE (m1.c0 IS NOT NULL)
     OR (m2.c0 BETWEEN SYMMETRIC '75.175.243.19' AND '230.9.216.68');

The pushed-down WHERE clause is an OR of an IS NOT NULL test on the outer side (selectivity 0.999, no stats on the empty table) and a range pair on the nullable side.  When clauselist_selectivity() merges the range pair (hibound + lobound - 1) it adds back the column's null fraction to undo the double exclusion of NULLs, and it does so by calling nulltestsel(IS_NULL, ..., jointype).  GPDB's nulltestsel() special-cases IS NULL under an outer join and returns 0.5 (a guess for the anti-join "WHERE inner.col IS NULL" pattern, removed upstream in e006a24a), so the range pair comes out as 0.99 + 0.5 = 1.49 instead of 0.99 + 0.0.  Combining that with the IS NOT NULL arm via s1 + s2 - s1*s2 gives 1.00049, which adjust_selectivity_for_nulltest() asserts on.

Two changes:

1. In clauselist_selectivity_ext(), ask nulltestsel() for the column's real null fraction by passing JOIN_INNER.  The range-pair correction is a statistical adjustment, not an IS NULL predicate evaluated at the join level, so the outer-join guess never belonged here.  Before this, every range pair evaluated under an outer join was inflated by an absolute 0.5; after it, a LEFT JOIN ON clause with a range condition gets the same estimate as the equivalent inner join.  This matches upstream behaviour, where nulltestsel() ignores jointype.

2. Selectivities are probabilities, so also clamp jselec and pselec to [0, 1] in calc_joinrel_size_estimate() before handing them to adjust_selectivity_for_nulltest(), so round-off in the OR combination can never trip the assertion again.

Add the SQLancer query to bfv_planner as a regression test.

Fixes #1950

### Type of Change
- [ ] Documentation update

### Test Plan

### Impact

**Performance:**

**User-facing changes:**

**Dependencies:**

### Checklist

### CI Skip Instructions

---


leborchuk (approved):
LGTM

It's amazing.

e006a24ad152b3faec748afe8c1ff0829699b2e6 (Tom Lane, 2008-08-14, PG 8.4, "Implement SEMI and ANTI joins") does contain exactly this fix in src/backend/utils/adt/selfuncs.c:

-     if (IS_OUTER_JOIN(jointype) && nulltesttype == IS_NULL)
-             return (Selectivity) 0.5;


 Tom added it in 0ee5a398627 (2007-08-31, PG 8.3) as an explicit band-aid for t1 LEFT JOIN t2 USING(key) WHERE t2.key IS NULL, saying "I went with 0.5 for lack of a better idea." He removed it in e006a24a because real ANTI-join detection made it unnecessary.

Cloudberry's behaviour was not changed 16 years ago. GPDB kept the hack when it cherry-picked that same commit in fe2eb2c9ed3 (2017-05-09), adding the note that survives verbatim in our tree at src/backend/utils/adt/selfuncs.c:1719:

▎ GPDB_84_MERGE_NOTE: Following hack is removed in the upstream commit e006a24a. However, removing this causes cost differences for some ICG queries. Hence, keeping the hack in GPDB

Time is come. The bug lived with us more then 18 years. Now he's an adult, and we can finally get rid of him.

chenjinbao1989 (approved):

reshke (approved):

PR #1984, opened 2026-09-11
Make initial value of the gp_create_table_random_default_distribution GUC equal to the default one

andr-sokolov:
The gp_create_table_random_default_distribution variable was initialized with
true, but the default value of the corresponding GUC is false. Set the initial
value to the GUC's default value to make the code easier to understand.

cherry picked from 01c6e69d807c62fda97779378b014f2f7906a695


leborchuk (approved):
LGTM

yjhjstz (approved):

PR #1987, opened 2026-09-11
Fix licence info about boost

leborchuk:
LICENSE (root) — added the missing coverage:

1. Boost Software License 1.0 section under the tabulate/MIT area — names the three bundled *-lite components that are not MIT, with exact paths:
   - optional_lite.hpp (optional-lite, Martin Moene)
   - variant_lite.hpp (variant-lite, Martin Moene)
   - string_view_lite.hpp (string-view-lite, Martin Moene) — this one has no in-tree LICENSE.* file, so it was the least visible of the three
   - single_include/tabulate/tabulate.hpp — the amalgamated header, which embeds copies of all three plus termcolor

   The existing tabulate MIT entry now reads "(excluding the bundled third-party components listed below)" so the MIT-only claim is no longer overbroad.
2. termcolor (BSD-3-Clause) added to the BSD 3 Clause section — also bundled by tabulate, also not MIT, and also not previously identified.
3. Ryu (Boost-1.0) added under the PostgreSQL section — src/common/{d2s.c,d2s_full_table.h,d2s_intrinsics.h,f2s.c,ryu_common.h} and src/include/common/shortest_dec.h. Upstream Ryu is Apache-2.0-or-Boost dual-licensed; PostgreSQL explicitly took it under Boost, and this was a second undeclared Boost-1.0 component in the archive.
4. Fixed dependencies/yyjson → dependency/yyjson (the declared path didn't exist).

New files (the root LICENSE now points into licenses/ for every entry, per existing convention):
- licenses/LICENSE-boost.txt — canonical BSL-1.0 text
- licenses/LICENSE-termcolor.txt — termcolor BSD-3 text

pom.xml — excluded those two new verbatim license texts from the Apache RAT scan, mirroring the existing tabulate/LICENSE.termcolor exclusion (that one is already excluded, which tells us RAT doesn't match that BSD variant).


tuhaihe:
I verified the claims against the pinned `tabulate` commit (`3a58301`) — Ryu file list, paths,
license texts and the RAT excludes all check out. Two things to fix:

**1. The MIT carve-out is too broad for the amalgamated header**

`single_include/tabulate/tabulate.hpp` is named under "the bundled third-party components listed
below", so it's now excluded from MIT entirely. But it's a composite file: lines 1–5624 are the
three `*-lite` libs (BSL-1.0), ~5625–6071 termcolor (BSD-3), and ~6072–9385 are tabulate's own MIT
code (`SPDX-License-Identifier: MIT`). That last ~3,300 lines is now covered by nothing.

Suggest making the carve-out apply to the embedded *portions*, not to whole files:

      contrib/pax_storage/src/cpp/contrib/tabulate
      (tabulate's own code only; the third-party code bundled and embedded
       within it is listed below under its own licenses)
      see licenses/LICENSE-tabulate.txt

**2. `utils/amalgamate/` is still undeclared**

    contrib/pax_storage/src/cpp/contrib/tabulate/utils/amalgamate/

`LICENSE.md` there is BSD-3-Clause, "Copyright (c) 2012, Erik Edlund" — same situation as termcolor:
not tabulate's MIT, ships with the submodule, not in the root LICENSE. Worth adding to the BSD 3
Clause section in the same pass.

(The two failing CI jobs are unrelated — no code changes here.)

tuhaihe (approved):

MisterRaindrop (approved):

tuhaihe:
Once this is merged, we can cherry-pick the commit to main.

PR #2000, opened 2026-09-15
Fix the clients path script name in gpAux/Makefile

tuhaihe:
The clients target copies the shell script that sets GPHOME_CLIENTS into the package it assembles:

	cp -f client/scripts/greenplum_$@_path$(SCRIPT) $(CLIENTSINSTLOC)/

With $@ expanding to "clients" and SCRIPT to ".sh" that resolves to client/scripts/greenplum_clients_path.sh, which does not exist.  The file was renamed to cloudberry_clients_path.sh during the rebranding and this reference was not updated with it, so the target would fail here if anyone reached it.

Nobody has: the top-level GNUmakefile deliberately steps over gpAux/Makefile, and the body of the clients target is skipped unless BLD_TARGETS contains "clients", which nothing in the tree sets.  So this is a latent break rather than a live one, and fixing it costs a word.

Verified with `make -C gpAux -n clients BLD_TARGETS=clients` on Rocky 9. Before, the expanded command reads

	cp -f client/scripts/greenplum_clients_path.sh ...

and after

	cp -f client/scripts/cloudberry_clients_path.sh ...

which matches the file in the tree and copies successfully when run from gpAux/, the directory make uses.

Note that gpMgmt/bin/gpload, gpload.py and gpdirtableload still look for greenplum_loaders_path.sh under $GPHOME_LOADERS.  Those are deliberate compatibility lookups against a separately installed Greenplum loaders package, not references to anything in this repository, so they are left alone.

Assisted-by: Claude Code

Fixes #ISSUE_Number

### Type of Change
- [ ] Documentation update

### Test Plan

### Impact

**Performance:**

**User-facing changes:**

**Dependencies:**

### Checklist

### CI Skip Instructions

---


leborchuk (approved):
LGTM, most probably we missed it while renaming, great to fix it

MisterRaindrop (approved):

PR #2002, opened 2026-09-15
Remove duplicated declaration of gp_create_table_random_default_distribution

andr-sokolov:
gp_create_table_random_default_distribution is a GUC variable and it is
declared in guc.h. There is no need to declare it a second time in cdbvars.h.


leborchuk (approved):
LGTM, already declared in https://github.com/apache/cloudberry/blob/320a6d35ca69c076a1ea93e107e38df7132ceb77/src/include/utils/guc.h#L341

yjhjstz (approved):

PR #2004, opened 2026-09-15
Add export control notice for bundled cryptography

tuhaihe:
Justin Mclean noted during the 2.2.0-rc1 IPMC vote that the release has no crypto export notice despite shipping contrib/pgcrypto, and that Cloudberry does not appear on https://www.apache.org/licenses/exports/.

infra.apache.org/crypto.html requires the notice in the README of each distribution, so add it there, listing the components of the release that involve cryptography: OpenSSL for TLS and SCRAM-SHA-256, optional GSSAPI and LDAP over TLS, and the bundled pgcrypto extension.  The build links against the operating system's OpenSSL and does not ship it, which the notice states because it determines who is named as manufacturer in the BIS notification.

This commit only covers the README.  Two steps remain, and neither lives in the source tree:

  - adding Cloudberry to https://www.apache.org/licenses/exports/, which is a pull request against apache/www-site, file data/eccn/eccnmatrix.yaml.  Podlings do not get their own top-level entry there; sixteen of them sit as products under a single

        - href: https://incubator.apache.org/
          name: Apache Incubator Project
          contact: Apache Incubator PMC <private@incubator.apache.org>

    umbrella, so Cloudberry belongs under that one, with the contact inherited from the Incubator PMC rather than one of our own lists.

  - sending the ECCN 5D002 notification to crypt@bis.doc.gov, enc@nsa.gov and web_site@bis.doc.gov.  This is a named legal notification submitted on behalf of the ASF.  The procedure describes it as coming from the PMC chair, and for a project in incubation the PMC is the Incubator PMC; infra.apache.org/crypto.html says nothing about podlings, so who sends it is a question for our mentors.  The site build generates the text of this email from the matrix entry, so it does not have to be written by hand.

Per the same page those are supposed to happen before the code is posted publicly, so they are overdue rather than new; the notification names the product without a version and does not need repeating per release.

Assisted-by: Claude Code

Fixes #ISSUE_Number

### Type of Change
- [ ] Documentation update

### Test Plan

### Impact

**Performance:**

**User-facing changes:**

**Dependencies:**

### Checklist

### CI Skip Instructions

---


tuhaihe:
Now it's ready for review. This change primarily takes this document as a reference: https://infra.apache.org/crypto.html#inform.

cc @leborchuk

README.md
```diff
@@ -132,6 +132,38 @@ Apache Cloudberry has a sound foundation.
...
+
+The following provides more details on the included cryptographic
+software:
+
+- Apache Cloudberry uses OpenSSL for TLS support on client and server
+  connections, and for the SCRAM-SHA-256 authentication method.
+- Apache Cloudberry optionally uses GSSAPI (Kerberos) for authentication and
+  connection encryption, and LDAP over TLS for authentication.
+- The bundled `pgcrypto` extension provides cryptographic functions to SQL,
+  using either OpenSSL or its own built-in implementations.
```
leborchuk, line 162: Also I've found in `src/bin/gpfdist/gpfdist.c` definition:

```
#ifdef USE_SSL
#include <openssl/ssl.h>
#include <openssl/rand.h>
#include <openssl/err.h>
#endif
```

Also gpfdist could use ssl certificates. It means it also use asymmetric cryptohraphy. I do not now if we should include gpfdist too, just pointed out

leborchuk (approved):
LGTM, I've read https://infra.apache.org/crypto.html#inform - all looks fine

tuhaihe:
Good catch, thanks — you are right, and I have added it in b23b9e6.

I checked it against a release-configuration build on Rocky 9 rather than
going by the source alone:

- `pg_config.h` carries `#define USE_OPENSSL 1`, so the `USE_SSL` blocks you
  found are compiled in
- `bin/gpfdist` ships in the packages and links `libssl.so.3` and
  `libcrypto.so.3`, with 14 SSL symbols in the binary
- its help text offers `--ssl dir : start HTTPS server. Use the certificates
  from the specified directory`
- `gpfdist.c:4364` calls `SSL_CTX_use_certificate_chain_file()` to load the
  chain

So it terminates TLS as a server using X.509 certificates, which is squarely
within 5D002.

I listed it on its own line rather than folding it into the first entry, for
two reasons: it is a standalone executable that people often deploy on hosts
away from the database, and it acts as a TLS server loading certificates
rather than just encrypting a client connection.

My first pass covered the server process and pgcrypto and missed the separate
binaries, so this was a real gap.

One related note: the eccnmatrix entry in apache/www-site#749 does not need a
change for this. Its controlled-source link points at the whole cloudberry
repository, so gpfdist is already covered.

README.md
tuhaihe: added it in https://github.com/apache/cloudberry/commit/b23b9e6c7ac6fd87d50e8b43a4df597e10d4b715

MisterRaindrop (approved):

PR #2022, opened 2026-09-16
Treat 8000-9000 oid range as non-pinned by system

reshke:
In PostgreSQL upsteam, OIDs from 8000-9000 are never used as system catalog oids in released branches.
So, Cloudberry extension Yezzey uses it as its poor-man-predefied-oid. 

add check for this case in IsPinnedObject so yezzey-managed objects are removable.


MisterRaindrop:
I don't think treating the whole [8000, 9000) range as unpinned is safe.

The 8000-9999 range is reserved for development OIDs (including in-progress patches and forks), rather than for unpinned extension objects. A core catalog object may legitimately use an 8xxx OID during development and should still be treated as pinned.

More importantly, this introduces an inconsistency with IsCatalogRelationOid(), which still treats every relation with relid < FirstUnpinnedObjectId (12000) as a system catalog. For example, a Yezzey relation with OID 8500 would become:

IsPinnedObject(..., 8500) == false

IsCatalogRelationOid(8500) == true

So the object becomes droppable while still being classified as a system relation.

